PR #2, opened 2015-12-15
Adding .deb installer

borisfom:
Most changes are in debian/ directory. Fixed tests not to depend on nvidia-ml.


borisfom:
Please hold on integrating this: I am rebuilding the installer and will update the PR shortly.

debian/control
```diff
@@ -0,0 +1,26 @@
...
+Maintainer: Boris Fomitchev <bfomitchev@nvidia.com>
+Priority: optional
+Build-depends: debhelper(>=9), nvidia-352-dev, cuda-toolkit-7-0,
+               cuda-curand-dev-7-0, cuda-cublas-dev-7-0, cuda-cudart-dev-7-0, cuda-ld-conf-7-0
+Standards-Version: 3.9.5
+
+Package: nccl
+Section: universe/tools
+Architecture: amd64
+Depends: ${misc:Depends}, ${shlibs:Depends},  nvidia-352,
```
lukeyeager: Is it necessary to depend on a specific driver version? What if they installed with a RUN file?

Kaixhin: Can confirm that I use the RUN file and know plenty of others do. A range of drivers should ideally be supported.

lukeyeager: I'd vote for moving `nvidia-352` to the `Suggests` field - i.e. not installed by default.

borisfom:
Superseded by Luke's request

PR #14, opened 2016-03-01
Added Debian bits, man page

borisfom:
Version bumped. Checked in new .deb packages binaries.


scripts/build-package.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# The envirinment set by create-ubuntu-environmens sscript is trusty, so we should match it.
+# Look at debian/gbp.conf for default optiuons sent to gbp.
+rm -r debian/tmp debian/*.log
+git commit -a -m 'Preparing for pbuild' && git push
```
lukeyeager: That seems like a dangerous/annoying piece of code to sneak in here. Why is it necessary to create a new commit?

lukeyeager:
Can we not check the .deb packages into git directly, but instead attach them to the tagged release?

See .deb packages here for an example:
https://github.com/NVIDIA/nvidia-docker/releases

scripts/build-package.sh
borisfom: In fact, gbp would fail the build if the workspace is not clean. Push is not necessary though.

debian/changelog
```diff
@@ -1,3 +1,10 @@
+nccl (1.1.0) trusty; urgency=medium
```
borisfom: Next time!

borisfom:
Created release

lukeyeager:
> Created release

Yeah just like that, but in this fork obviously.

Packaging scripts look good to me, however I haven't actually tested them.

borisfom:
Will submit a superceding request

PR #18, opened 2016-04-13
Add --no-as-needed to make sure that cudart library gets linked

apaszke:
This fixes ngimel/nccl.torch#5. Apparently without this flag `libcudart.so` wasn't present in the dynamic table of `libnccl.so`, and wasn't loaded when required in torch.


onepiece666:
can you please show me the details.  I already added libnccl.so to my LD_LIBRARY_PATH.i do not know how to modify de nccl's makefile

PR #41, opened 2016-08-19
Some minor fixes for compile/usage

jia-kai:
Two minor modifications are included in this PR:

1. Add `-lrt` to `LDFLAGS`. `shm_open` needs this flag, and it seems that nvcc automatically adds it; however, if one wants to link `libnccl.so` manually with other projects using `ld`, it would fail due to undefined symbol.
2. Change the type of `devlist` param in `ncclCommInitAll` from `int*` to `const int*`, since it would not be modified and the user might actually pass a `const int` pointer.


sjeaugey:
Hi Jia-kai,

Thanks for the fixes !

PR #47, opened 2016-09-15
NCCL Fortran bindings

kylefernandes:
Hi All,

I've added Fortran bindings for ncclBCast, ncclAllGather and ncclReduce. I plan on adding ncclAllReduce and ncclReduceScatter very soon.

I've also added tests/samples for using these with floats using both "Fortran Array" and "Pointer" syntax. The tests/samples for ncclAllGather and ncclReduce are for "out-of-place" operation. I plan on adding "in-place" tests/samples soon too.

Cheers,
Kyle


sjeaugey:
Hi Kyle,

Thanks for making NCCL work in Fortran programs too !

I'm a little concerned however about the increased maintenance : every time we modify the API, we would have to update the Fortran bindings too. I guess if we only add new functions and values it shouldn't break anything, right ? In that case maybe that would be OK. I just want to make sure we have the minimum amount of code and maintenance while keeping the ability to use NCCL inside Fortran programs.

Also, I know Fortran modules are great but they're also tied to the Fortran compiler which can be quite annoying when doing releases (unless you know a way to pre-compile the fortran module for a variety of compilers and automatically load the right one ?). I prefer .h files since they are smaller, simpler and work all the time (though I agree also when it even shouldn't compile).

kylefernandes:
Hi Sylvain,

Thanks for your response.

Interface maintenance is a bit of a nightmare. I've had my fair share of frustration with some of the PGI CUDA Fortran interfaces (e.g. cuBlas), so I completely understand your concerns. Like you say, new additions shouldn't break anything. Unfortunately, if existing functions are changed, I can't think of any solution other than to update the interfaces as required. Hopefully this wouldn't involve too much work and hopefully the interfaces wouldn't change too much/too often?

Again, I understand your concerns with the distribution of Fortran modules. I think it's safe to say that this would only be compiled with a CUDA Fortran enabled compiler (?), of which there are not that many at the moment and of which there probably won't be too many in the future? As horrible and clunky as it is, could it be possible to release a module for each compiler (perhaps with the compilers name or some other unique identifier in the name of the module file) and make it clear that the correct one needs to be included? For people downloading and installing NCCL from source, rules could be included in the NCCL Makefile to automatically create the module file with the correct unique compiler name identifier based on whichever compiler they have chosen? Hope this makes sense...

Cheers,
Kyle

sjeaugey:
Hum, that is interesting. I didn't realize even the internals of NCCL were reflected in the fortran bindings. Is that really necessary ? Couldn't we treat the ncclComm_t as an opaque object ?

If not, then my concerns are even bigger : the interface should not change too much, but the internals of a NCCL communicator will change a lot ...

kylefernandes:
Hi Sylvain,

I’m not sure how, but I somehow convinced myself that ncclComm (and not only ncclComm_t) was being accessed directly from the use/test examples. I assume it was a combination of 1) not reading correctly/reading source rather than tests and 2) me testing that the members of ncclComm have the same values in Fortran as they do in C/C++. I’ve concealed ncclComm by treating ncclComm_t as an opaque object as you suggested.

Regards,
Kyle

kylefernandes:
Hi Sylvain,

When you get a chance, please have a look at my latest updates and let me know what you think.

sjeaugey:
Thanks Kyle for updating the bindings. I would prefer the Fortran code to be in a fortran directory instead of directly in src. Also, there is a typo on cudaMAlloc in cudaFor.f90, but maybe the case doesn't matter as we're in Fortran.

kylefernandes:
I was thinking that `cudafor.f90` shouldn't be in `src` anyway, so it's no problem to move `ncclfor.f90` too.

I was thinking of having a `cudafor` directory at the same level as `src` just for `cudafor.f90` (as it's technically separate from NCCL). What do you think of this?

Would you like a directory within `src` for `ncclfor.f90` or at the same level as `src`? Any particular preference for a directory name? `fortran` or `ncclfor` or `srcfor` or something else?

Yup, case doesn't matter in Fortran except in the naming of the routine that the binding is for, which you'll see is (and needs to be) the same case as the C/C++ routine. I can change the Fortran names to match the C/C++ API exactly if you prefer. I think `cudaMemCpy` and `ncclBCast` are the only other ones that would need to be changed.

kylefernandes:
Hi Sylvain,

I've moved `src/cudafor.f90` to `cudafor/cudafor.f90` and `src/ncclfor.f90` to `forsrc/ncclfor.f90`.

Cheers,
Kyle

sjeaugey:
Hum, why didn't you move cudafor.f90 in forsrc ?
Also, I'm not sure forsrc will be clear.A "fortran" directory is usually used in similar projects and makes no doubt that this is containing the fortran bindings.

In fact, the fortran bindings can be seen as a layer on top of NCCL. I'm OK to integrate it in the NCCL repository, as it would be more convenient for users who want directly Fortran support. However, it would be interesting to have this as a standalone project with its own Makefile, so that users can even compile the bindings for a NCCL library that is already installed on their machine if they want. The "fortran" target in the top Makefile would just call "make -C fortran" using the local NCCL library if needed.

Does that make sense to you ?

kylefernandes:
Hi Sylvain,

Please have a look and let me know what you think.

It doesn't have the ability to be built on top of a pre-installed NCCL yet. I'll try look into that soon.

At the moment, the NCCL Makefile has the following line:
`CUDA_HOME ?= /usr/local/cuda`

what do you think about changing this to:
`CUDA_HOME ?= $(CUDA_INSTALL_PATH)`

Cheers,
Kyle

kylefernandes:
Hi Sylvain,

Are you happy to merge this pull request?

Cheers,
Kyle

kylefernandes:
Just saw that you didn't answer about potentially changing `CUDA_HOME ?= /usr/local/cuda` to `CUDA_HOME ?= $(CUDA_INSTALL_PATH)` in `$(NCCLROOT)/Makefile`. What do you think?

sjeaugey:
I don't see the goal of changing CUDA_HOME. CUDA_HOME is set by default to /usr/local/cuda which is the default install directory ; you can set CUDA_HOME=$CUDA_INSTALL_PATH in you environment and that should work ... what am I missing ?

About the merge, I'd like to squash the PR to keep the history simple, but I'm wondering which author to use. I could make you the author, or only mention you in the commit message ... let me know what you prefer.

kylefernandes:
I think I may be confused about `$CUDA_INSTALL_PATH`. Please ignore the question.

I would like to come up as one of the contributors with as close as possible to a true reflection of the "size" of the contribution. I could close the pull request and open a new one with a single "Add NCCL Fortran bindings" commit, but that would hide your changes that I merged. Or I could create a single commit with all the changes before your changes that I merged and you can make a single "Consolidate NCCL Fortran bindings" (or something like that) commit with those changes and then merge those 2 commits into NCCL/master? Let me know what you think.

sjeaugey:
No need to reopen a PR. I was ready to squash everything and push it with your name on it. My contribution here was just to make things consistent with the upper Makefile.

About copyright, it reminds me you didn't include one in the fortran files. Could you please open add a patch with only headers to all .f90 files that would look like :
``` 
!*************************************************************************
!* Copyright (c) 2016, RCS Cambridge. All rights reserved.
!*
!* See LICENSE.txt for license information
!************************************************************************
```
Thanks !

kylefernandes:
Hi Sylvain,

Thanks for accepting the pull request. I've added the copyright as requested, but I didn't see that you had merged the pull request before I pushed the copyright changes to the request. Would you prefer for me to open a new pull request with the copyright additions from the latest master?

sjeaugey:
No need to open a PR, I'll cherry pick the commit.

kylefernandes:
Great. Thanks. Should I specifically close this PR?

kylefernandes:
I just realised that I was printing something in the ncclReduce tests that is irrelevant and doesn't make sense to print so I've removed it. Apologies.

sjeaugey:
I cherry-picked the two commits and pushed them. Closing this PR.

Thanks !

PR #54, opened 2016-11-24
Add a static library target "staticlib" to the Makefile.

peterhj:
This commit adds a Makefile target named "staticlib", which as its name suggests builds a static library "libnccl.a" for NCCL.


cliffwoolley:
A static build could result in multiple NCCLs in the same process space.
Do you anticipate any problems with this, @sjeaujey?  It sounds risky to me
given the way NCCL synchronizes itself.

peterhj:
One option is to rename "libnccl.a" -> "libnccl_static.a" following the convention used by the rest of the CUDA libraries. I'll submit a rebased PR shortly.

sjeaugey:
Given the current implementation, using two different static libraries would indeed cause problems. But using a static library is needed in some cases, so this is something we should provide for those who know what they are doing.

PR #77, opened 2017-03-15
Fix compilation error when compiling with 'clang -x cuda'.

ilya-biryukov:
This PR fixes compilation failure when compiling with clang (using -x cuda option). 

The fix simply moves up vFetch and vStore declarations('common_kernel.h'), so they are declared before their usages in ReduceCopy. This does not break compilation with nvcc.

It worked with nvcc before that commit because of the differences in argument-dependent-lookup implementation.


ilya-biryukov:
Sorry, used the wrong email in the commit's author, will resend it a bit later

PR #95, opened 2017-06-08
Windows compatibility maintained while also ensuring linux compatibility.

pseudotensor:


sjeaugey:
Hi Jonathan,

Thanks for your interest in NCCL and your efforts to make it work on all platforms !

We are not ready (yet) to maintain windows code, so we won't be able to merge this PR until we do. It is however a great way to make your windows version visible so that users can pick it and use it in the meantime.

eabase:
Are you ready now?

PR #119, opened 2017-11-12
fix tests on maxwell

sclarkson:
Tests currently fail on Maxwell cards with driver 384.

`Cuda failure test/include/test_utilities.h:313 'invalid argument'`

cudaHostUnregister is being called on the device pointer instead of the host pointer. This works on Pascal because the two are the same, and evidently worked on Maxwell with older driver versions.


SCP-173-cool:
The same error on M40.

PR #156, opened 2018-11-10
Generate host-hash for P2P and SHM based on $(readlink /proc/self/ns/uts) + $(readlink /proc/self/ns/mnt)

### LINKED ISSUE #155: Mesos containerizer does not isolate UTS namespace
Hi folks,

Mesos containerizer [does not have](https://issues.apache.org/jira/browse/MESOS-7605) native support for isolating UTS namespace.

Because of that, NCCL always assumes that different containers running on the same host are able to do P2P and SHM to each other: [code1](https://github.com/NVIDIA/nccl/blob/b56650c7f59b8cd40d18809784a6d6be38ef8acb/src/misc/utils.cu#L38), [code2](https://github.com/NVIDIA/nccl/blob/f93fe9bfd94884cec2ba711897222e0df5569a53/src/transport/shm.cu#L72).

Do you think it'd be reasonable to extend getHostHash() to include other isolation namespaces, e.g. `/proc/self/ns/mnt`?

alsrgv:
Fixes #155

### DIFF
--- a/src/misc/utils.cu
+++ b/src/misc/utils.cu
@@ -8,8 +8,6 @@
 #include "debug.h"
 #include <unistd.h>
 #include <string.h>
-#include <sys/types.h>
-#include <dirent.h>
 
 ncclResult_t getHostName(char* hostname, int maxlen) {
   if (gethostname(hostname, maxlen) != 0) {
@@ -35,30 +33,23 @@ uint64_t getHash(const char* string) {
  * that will be unique for both bare-metal and container instances
  * Equivalent of a hash of;
  *
- * $(hostname) $(readlink /proc/self/ns/\*)
+ * $(hostname) $(readlink /proc/self/ns/uts) $(readlink /proc/self/ns/mnt)
  */
 uint64_t getHostHash(void) {
   char uname[1024];
   // Start off with the hostname
   (void) getHostName(uname, sizeof(uname));
   int offset = strlen(uname);
-
-  DIR *dp = opendir("/proc/self/ns");
-  if (dp != NULL) {
-    struct dirent *ep;
-    while (ep = readdir(dp)) {
-      if (offset >= sizeof(uname) - 1) {
-        break;
-      }
-      char ns_path[1024];
-      sprintf(ns_path, "/proc/self/ns/%s", ep->d_name);
-      int len = readlink(ns_path, uname+offset, sizeof(uname)-1-offset);
-      if (len < 0) len = 0;
-      offset += len;
-    }
-    (void) closedir(dp);
-  }
-
+  int len;
+  // $(readlink /proc/self/ns/uts)
+  len = readlink("/proc/self/ns/uts", uname+offset, sizeof(uname)-1-offset);
+  if (len < 0) len = 0;
+  offset += len;
+  // $(readlink /proc/self/ns/mnt)
+  len = readlink("/proc/self/ns/mnt", uname+offset, sizeof(uname)-1-offset);
+  if (len < 0) len = 0;
+  offset += len;
+  // Trailing '\0'
   uname[offset]='\0';
   TRACE(INIT,"unique hostname '%s'", uname);
 
--- a/src/misc/utils.cu
+++ b/src/misc/utils.cu
@@ -8,6 +8,8 @@
 #include "debug.h"
 #include <unistd.h>
 #include <string.h>
+#include <sys/types.h>
+#include <dirent.h>
 
 ncclResult_t getHostName(char* hostname, int maxlen) {
   if (gethostname(hostname, maxlen) != 0) {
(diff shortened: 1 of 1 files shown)


alsrgv:
@AddyLaddy, @sjeaugey, could you guys take a look?

sjeaugey:
Sorry -- forgot to do it yesterday.

I'm concerned by two things. The first one is basic length size : will 1024 bytes be enough for hostname + all namespaces ?

The second one is more fundamental : I'm still trying to convince myself that we're doing the right thing. I can definitely see the sense of deciding that two processes that are not in the same mnt namespace cannot communicate through SHM (since it's a file). But I'm not sure why, for example, processes in different pid namespaces would not want to communicate using P2P or SHM. I'd need to do more research on namespaces, why they are used and how this affects NCCL.

I guess your approach will work anyway but I'd like to avoid creating situations where performance is very bad and it ends up being caused by some tool or library which uses namespaces for some reason.

AddyLaddy:
A typical output shows

`hostname:[4026531969]uts:[4026531838]ipc:[4026531839]pid:[4026531836]user:[4026531837]mnt:[4026531840]cgroup:[4026531835]`

That gives around 128 chars for a reasonable length hostname, so I think 1024 is sufficient.

AddyLaddy:
But as Ke found, Kubernetes have multi-container PODs; https://www.mirantis.com/blog/multi-container-pods-and-container-communication-in-kubernetes/

kwen2501:
In the case that 1024 is not long enough, we'd need to make sure the important ones come first, e.g. hostname >= uts >= ...

alsrgv:
I think hostname length is usually limited to 255 characters (https://stackoverflow.com/questions/8724954/what-is-the-maximum-number-of-characters-for-a-host-name-in-unix), and there are only 9 namespaces at the moment - so we're well within 1024 character limit for a number of years to come.

Docker uses `uts` namespace, and Kubernetes does not support sharing `uts` namespace between containers anyway - https://docs.docker.com/engine/reference/run/#uts-settings---uts, https://github.com/kubernetes/kubernetes/issues/1615.

In summary, with Docker in the vast majority of cases, you cannot do P2P and SHM because of UTS namespace, even if P2P and SHM are physically possible.  I want to bring Mesos to parity with this by expanding the scope.

If we do want to allow SHM or P2P between containers, I propose we add an explicit opt-in option for that separately (if somebody asks for it).

alsrgv:
Bump :-)

alsrgv:
@sjeaugey, sure - I'm happy to modify the PR to only hash on UTS and MNT namespaces.

alsrgv:
Updated.

alsrgv:
Thanks!

AddyLaddy:
@alsrgv Alex, some internal and external users have asked to enable NCCL P2P/IPC/SHM between containers running on a node. We have tested this and it works fine. So we are proposing making a change to allow this by default, but with a new environment variable that will let the user to specify what files are used to build the unique host hash. e.g. `NCCL_HOSTHASH=uts,mnt` would restore the current behavior. The default will be `NCCL_HOSTHASH=uts`

alsrgv:
@AddyLaddy, we can set this variable in our environment. I'm curious, what tricks do they use for sharing to work? Are there simple examples using nvidia-docker? If they need to set some special flags, it may be able to keep `NCCL_HOSTHASH=uts,mnt` by default for safety, and have them specify `NCCL_HOSTHASH=uts` as an override.

AddyLaddy:
@alsrgv Thanks for your feedback. We are now pursuing an alternative approach where we replace the current <hostname>+<uts>+<mnt> string with `/proc/sys/kernel/random/boot_id`
There will then be a new env var; `NCCL_HOSTID` which can be used to replace that string if necessary. So a user could recreate the current behaviour by setting that environment variable using a wrapper script.
We have also added checks for the device MAJOR:MINOR of /dev/shm so that NCCL will only attempt to make P2P/IPC or SHM connections if the device is common between the containers.

sjeaugey:
To give a bit more background as to why we changed direction : I have the feeling the approach of using namespaces is very dependent on the way users were configuring their namespaces, hence on the container system and also the use case.

Which is why we found several cases where things would not work the way we wanted to. The goal of the new approach is to be universal and get the best performance regardless of which container is used and how namespaces are configured, and what is shared between containers or not.

We consider that all containers systems mount /proc so we have a reliable way of knowing if we are on the same host. If we are on the same host, then we can check for `/dev/shm` inode major/minor which tells us whether the user mounted the same `/dev/shm` on the containers or created separate ones. It also tells us whether we can try using p2p between devices.

Let us know if we missed anything important or if you see cases where this approach would not work.

alsrgv:
@sjeaugey, @AddyLaddy, if you guys have a PR open I’d love to try it out. Not having to worry about P2P/SHM visibility would be magical.

AddyLaddy:
@alsrgv See PR #248

PR #168, opened 2018-12-13
Use CUDA_VERSION instead of __CUDACC_VER_*__

chsigg:
This is the first of four CLs to improve source compatibility with clang.


sjeaugey:
Had to revert this change ... it breaks the compilation.

@chsigg would it work to replace CUDA_VERSION by CUDART_VERSION ?

chsigg:
Yeah, I noticed exactly that too in the meantime. Sorry about that.

PR #169, opened 2018-12-13
Two temporary workarounds for cuda-clang issues.

chsigg:
This is the second of four CLs to improve source compatibility with clang.

### DIFF
--- a/src/collectives/device/functions.cu
+++ b/src/collectives/device/functions.cu
@@ -56,9 +56,17 @@
 
 // Must be consistent with the ncclFuncSet enum
 __device__ ncclKern_t ncclFuncs[ncclCollCount*ncclNumOps*ncclNumTypes*2] = {
+// Don't try to initialize the host shadow copy of this device-side global
+// variable. There is no host pointer to a device-side function, which
+// confuses clang. This will be fixed in the next clang release.
+#if __CUDA_ARCH__
   NCCL_FUNCS2B(ncclBroadcast),
   NCCL_FUNCS2A(ncclReduce),
   NCCL_FUNCS2B(ncclAllGather),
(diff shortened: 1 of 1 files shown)


src/collectives/device/functions.cu
```diff
@@ -56,9 +56,17 @@
...
   NCCL_FUNCS2B(ncclBroadcast),
   NCCL_FUNCS2A(ncclReduce),
   NCCL_FUNCS2B(ncclAllGather),
   NCCL_FUNCS2A(ncclReduceScatter),
   NCCL_FUNCS2A(ncclAllReduce)
+#endif
 };
+
+// Workaround for https://reviews.llvm.org/D55580
+__device__ void ncclWorkaroundClangD55580() {}
```
sjeaugey: Curious to understand what this is doing.

chsigg: It prevents a clang from tripping on a corner case which is described in the summary of the given link. The linker strips this function because it's unused, so there is no negative side effect to this workaround.

sjeaugey: I was actually curious as to how it works. Does clang detect the "WorkaroundClang" keyword and do something special ?

chsigg: Ah, sorry for the misunderstanding. No special detection. Clang just requires to have at least one function defined in each compilation unit that has function references. Without RDC, this is trivially the case. Adding any function definition triggers clang to initialize a field that is used when processing the function references. The bug has already been fixed upstream, but it will take a while for this to land on the user's machines.

chsigg: In other words, the function could also be called foobar().
;-)

sjeaugey: Oh.. OK, makes sense. Thanks !

PR #171, opened 2018-12-13
Replacing non-standard variable length arrays with unique_ptr.

chsigg:
This is the fourth of four CLs to improve source compatibility with clang.

### DIFF
--- a/src/init.cu
+++ b/src/init.cu
@@ -607,6 +607,7 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, ncclUniqueId* comm
   if (intraRank == -1 || intraRank0 == -1 || rankInfos[intraRank0].comm == NULL) {
     WARN("Failed to determine intra ranks hostHash[%d] %lx intraRank %d intraRanks %d intraRank0 %d",
         rank, rankInfos[rank].hostHash, intraRank, intraRanks, intraRank0);
+    free(rankInfos);
     return ncclInternalError;
   }
   NCCLCHECK(ncclCommSetIntra(comm, intraRank, intraRanks, rankInfos[intraRank0].comm));
--- a/src/init.cu
+++ b/src/init.cu
@@ -29,7 +29,6 @@
 #include <errno.h>
 #include <assert.h>
 #include <dlfcn.h>
-#include <memory>
 
 #define STR2(v) #v
 #define STR(v) STR2(v)
@@ -332,8 +331,10 @@ static ncclResult_t fillConnect(struct ncclInfo* allInfo, int nranks, int rank, 
 }
 
 static void swap(void* mem1, void* mem2, int size) {
-  std::unique_ptr<char[]> tmp(new char[size]);
-  memcpy(tmp.get(), mem1, size); memcpy(mem1, mem2, size); memcpy(mem2, tmp.get(), size);
+  char* tmp;
+  NCCLCHECK(ncclCalloc(&tmp, size));
+  memcpy(tmp, mem1, size); memcpy(mem1, mem2, size); memcpy(mem2, tmp, size);
+  free(tmp);
 }
 
 #define MAXWIDTH 20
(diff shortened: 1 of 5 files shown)


chsigg:
Thanks Sylvain, I was aware that this change was sticking out, but couldn't think of a better way. I will switch to ncclCalloc.

src/init.cu
```diff
@@ -604,6 +610,7 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, ncclUniqueId* comm
     return ncclInternalError;
```
sjeaugey: We should add `free(rankInfos);` here as well (before return ncclInternalError).

chsigg: Done, thanks for spotting that!

src/transport/net.cu
```diff
@@ -152,12 +152,15 @@ static inline int groupBestEnd(int nranks, int* groups, int group, int* subgroup
...
   for (int c=0; c<NET_MAX_IFS*nGroups; c++) cardUsed[c] = 0;
 
   for (int ring = 0; ring<*nringsRet; ring++) {
-    int starts[nGroups];
-    int ends[nGroups];
+    int* starts;
+    NCCLCHECK(ncclCalloc(&starts, nGroups));
+    int* ends;
+    NCCLCHECK(ncclCalloc(&ends, nGroups));
     for (int group = 0; group<nGroups; group++) {
```
sjeaugey: Maybe we should get those two allocations out of the loop to avoid multiple malloc/free.

src/transport/net.cu
```diff
@@ -183,6 +186,9 @@ ncclResult_t netGetRings(int nranks, int* groups, int* subgroups, ncclTvalue_t*
       }
       if (starts[group] == -1 || ends[group] == -1) {
         *nringsRet = ring;
+        free(starts);
```
sjeaugey: Once all alloc/free are outside the loop, we can call `goto final;` and have a `final:` label before the 3 calls to `free();` and `return ncclSuccess;`

sjeaugey:
@chsigg while ncclCalloc can easily be used in some places, in some others, there are lots of calls to `NCCLCHECK`/`CUDACHECK` which will return if there is an error. It is therefore hard to add calls to `free()` everywhere. One solution would be to use `NCCLCHECKGOTO`/`CUDACHECKGOTO` to make sure we always go to the end and free data.

Another solution might be the use of `alloca()`. Do you think that would be a viable option to minimize the impact on the code ?

borisfom:
I thought clang has the same variable length array extension as gcc ?

Anyway, if we have to get rid of them, I would recommend simpler and safer solutions:
1) in places where there is  a known maximum length - just change variable length buffer to a fixed max size;
2) In places where fixed maximum buffer does not work, use "vector<T> buf(N)" instead of "T buf[N]" for minimally invasive change, no manual allocation/deallocation and exception safety.

I would not recommend using alloca() directly as it is even less portable than variable length arrays.

borisfom:
And yes those buffers should be moved out of the loops - what happens now under the scenes is allocating/deallocating stack on each iteration. Not a lot of work, but waste anyway - and would become more substantial if changed to vector<>.

src/transport/net.cu
chsigg: Done, here and in p2p.cu.

src/transport/net.cu
chsigg: Done.

sjeaugey:
@chsigg I merged the 3 other PRs. On this one, it seems we need to think a bit more about it.

How urgent is it for you ?

chsigg:
Thanks Sylvain and Boris. It's not urgent. Clang does support VLAs, we just need to disable the warning.
I'm happy to take another stab when you decided what approach you prefer (or leave as is). Just let me know.

chsigg:
Hi Sylvain and Boris. Did you come to a conclusion how to best avoid VLAs? It's still not urgent, but it would be nice to get this off the table. Thanks!

sjeaugey:
Not yet. I'm fearing this work might need to be redone for 2.4 (since it is a major code rewrite), so I'd defer it to after 2.4 is merged into master.

sjeaugey:
This should have been fixed in 2.4. Closing.

PR #172, opened 2018-12-15
NCCL 2.4 : tree algorithms

sjeaugey:
This is a preview branch for 2.4. It is not meant to be used in production yet.

To add tree algorithms, a significant part of the code has been rewritten. Performance on 200+ GPUs should be significantly improved.

There is still some tuning to do on the small number of nodes so you might see corner cases where performance is under 2.3.7.

Main changelog :
Add tree algorithms for allreduce to improve performance at scale.
Add abort mechanism, function to get asynchronous errors.


sjeaugey:
Thanks for the report @kuenishi ! Fixed.

rashikakheria:
Thanks @sjeaugey for the mention. As far as I can tell, addition of regMr / deregMr APIs are relevant to plugin supporting host address pointers so will not require any addition to OFI plugin other than function definitions.

I will give this branch a try with the latest plugin and update if I find any issue.

kuenishi:
@sjeaugey Thanks for the heads up. That would be fine to me as the change would be as small just as my `ncclCommDestroy` rewritten to `ncclCommAbort` .

rashikakheria:
Too late to the party, but this patch works with the plugin without any changes required.

kuenishi:
@sjeaugey Thanks for updating the branch. Do you have any plan to add error code like `CUDNN_STATUS_NOT_SUPPORTED` for NCCL, say, like `NCCL_STATUS_NOT_SUPPORTED` ? I know adding fresh new functions in new versions doesn't need error code like this for C library itself, (only in case of adding new features in existing functions), but for a wrapper library like our [cupy](https://cupy.chainer.org/) supports multiple versions of NCCL (and CUDA, CuDNN and such), adding new function stubs for old versions.

PR #174, opened 2019-01-07
Fix memory leak in bootstrapRoot()

chsigg:


sjeaugey:
Thanks. Note this would have been fixed in 2.4. https://github.com/NVIDIA/nccl/blob/v2.4/src/bootstrap.cu#L129-L130

PR #180, opened 2019-02-13
Fix memory leaks.

chsigg:


chsigg:
May I also suggest that you test NCCL with a leak detector internally?

sjeaugey:
These should have been fixed in 2.4.

Issue #186, opened 2019-03-01
include .pc (pkgconfig) file in the repo


sl1pkn07:
Hi

actually, the project lack the pkgconfig (.pc) file, usefull for do other project detect the library/includes for use in other projects

is possible add it?

greetings

AddyLaddy:
Sure we could add one. Do you have an example for us to look at ?

sl1pkn07:
@AddyLaddy  please test PR #188 

greetings

sl1pkn07:
merged by #190 

greetings

PR #188, opened 2019-03-05
Add pkgconfig file

sl1pkn07:
generate and install pkgconfig file (https://www.freedesktop.org/wiki/Software/pkg-config/)

related: https://github.com/NVIDIA/nccl/issues/186

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -17,16 +17,20 @@ LIBSRCFILES := init.cu channel.cu bootstrap.cu transport.cu enqueue.cu \
 ##### lib files
 LIBNAME     := libnccl.so
 STATICLIBNAME := libnccl_static.a
+##### pkgconfig files
+PKGCONFIGFILE := nccl.pc
 ##### dirs
 BUILDDIR ?= $(abspath ../build)
 INCDIR := $(BUILDDIR)/include
 LIBDIR := $(BUILDDIR)/lib
 OBJDIR := $(BUILDDIR)/obj
+PKGDIR := $(BUILDDIR)/lib/pkgconfig
 ##### target files
 INCTARGETS := $(INCEXPORTS:%=$(INCDIR)/%)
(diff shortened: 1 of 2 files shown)


src/Makefile
```diff
@@ -77,6 +81,16 @@ $(LIBDIR)/$(STATICLIBTARGET): $(LIBOBJ) $(DEVICELIB)
 	ar cr $@ $(LIBOBJ) $(TMP)/*.o
 	rm -Rf $(TMP)
 
+$(PKGDIR)/nccl.pc : nccl.pc.in
+	@mkdir -p $(PKGDIR)
+	printf "Generating %-35s > %s\n" $< $@
+	printf $(PREFIX)
```
sjeaugey: Not sure if that printf was supposed to stay

src/Makefile
```diff
@@ -77,6 +81,16 @@ $(LIBDIR)/$(STATICLIBTARGET): $(LIBOBJ) $(DEVICELIB)
...
+	sed -e 's|$${nccl:Prefix}|\$(PREFIX)|g' \
```
sjeaugey: Is there a reason for this sed expr being different from the others below ? Note I don't have a preference, just looking for consistency.

sl1pkn07: because i test with other form and sed fails, i think because the '$(PREFIX)', the start character is '/'

sl1pkn07: the generated pc file is:

~~~
prefix=/usr/local
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: nccl
Description: Optimized primitives for collective multi-GPU communication
Version: 2.4.2
Libs: -L${libdir} -lnccl
Cflags: -I${includedir}
~~~
(makefile launched with: make CUDA_HOME='/opt/cuda' -j40 V=1)

src/Makefile
sl1pkn07: oh, yes, is a typo

fixing

PR #189, opened 2019-03-05
Add pkgconfig file (v2)

sl1pkn07:
generate and install pkgconfig file (https://www.freedesktop.org/wiki/Software/pkg-config/)

related: #186

(ammend from the PR #188)

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -17,16 +17,20 @@ LIBSRCFILES := init.cu channel.cu bootstrap.cu transport.cu enqueue.cu \
 ##### lib files
 LIBNAME     := libnccl.so
 STATICLIBNAME := libnccl_static.a
+##### pkgconfig files
+PKGCONFIGFILE := nccl.pc
 ##### dirs
 BUILDDIR ?= $(abspath ../build)
 INCDIR := $(BUILDDIR)/include
 LIBDIR := $(BUILDDIR)/lib
 OBJDIR := $(BUILDDIR)/obj
+PKGDIR := $(BUILDDIR)/lib/pkgconfig
 ##### target files
 INCTARGETS := $(INCEXPORTS:%=$(INCDIR)/%)
 LIBSONAME  := $(LIBNAME:%=%.$(NCCL_MAJOR))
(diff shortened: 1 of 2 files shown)


sl1pkn07:
fua ><

sorry guys

sjeaugey:
What's the reason for that second PR ? I don't see any difference with #188.

sjeaugey:
Sorry -- didn't see you closed it.

sl1pkn07:
yep. 

ammend the previous pr, but i forgot do git  add file before amend, :/

sorry:


BTW, is possible change the commits in a created PR? how?

sjeaugey:
If you commit and push to your branch, I think the changes will be reflected in the PR. Then when we merge we can squash all commits into a single one (for a clean history).

sl1pkn07:
seems not work like expected if the pr is from the first commit of the new branch. because the 188 and 189 did not update with the ammend commit

if you can, plesae remove the pr 188 and 189 for clear the history

very sorry and greetings

sjeaugey:
sl1pkn07 force-pushed the sl1pkn07:pkgconfig branch from b2a6c1e to 08c84e4 an hour ago

sl1pkn07:
yep, i force it, but in the 188 and 189 pr (web) still show the unammend commit

sjeaugey:
> yep, i force it, but in the 188 and 189 pr (web) still show the unammend commit

PR #190 (which is still open) shows at the end :
« Add more commits by pushing to the pkgconfig branch on sl1pkn07/nccl. »
While 188/189 don't because (I believe) they are closed. 188 got the commit from 189 before being closed.

sl1pkn07:
yep, you are rigth, seems in the pr 188 the review comments keep in the web the old code instead of the new code, and in the 189 not reflect this because you say

all "ok" then

if you can, please remove the pr 188 and 189 from the history, and keep/reviewed the pr 190. i be more carefull in the next time

greetings

sjeaugey:
I don't think I can delete anything (not even hide it). But that doesn't really matter. And don't worry too much about it, we all struggle on our first Github PR :-)

Issue #219, opened 2019-05-07
Include gencodes for CUDA 10


jakirkham:
Currently [CUDA 8 and CUDA 9 gencodes]( https://github.com/NVIDIA/nccl/blob/9db4b1d801624a00591b7aafd426d6dd23547443/makefiles/common.mk#L26-L36 ) are listed. However CUDA 10 gencodes are not listed. Would it be possible to add the CUDA 10 gencodes as well?

sjeaugey:
CUDA 10 did not introduce new SM architectures (the latest, `sm_70`, is already supported by CUDA 9), so we didn't need to add a new line for CUDA 10.

jakirkham:
Thanks for the clarification.

jakirkham:
Submitted PR ( https://github.com/NVIDIA/nccl/pull/221 ) adding a comment to this effect. Please let me know what you think. 🙂

PR #221, opened 2019-05-07
Comment on CUDA 10 gencodes in `common.mk`

jakirkham:
Adds a clarifying comment on CUDA 10 gencodes to point out they are the same as CUDA 9 gencodes.

Fixes https://github.com/NVIDIA/nccl/issues/219

### DIFF
--- a/makefiles/common.mk
+++ b/makefiles/common.mk
@@ -25,6 +25,7 @@ CUDA_MINOR = $(shell echo $(CUDA_VERSION) | cut -d "." -f 2)
 
 # Better define NVCC_GENCODE in your environment to the minimal set
 # of archs to reduce compile time.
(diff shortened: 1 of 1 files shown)


jakirkham:
Any thoughts on this, @sjeaugey?

sjeaugey:
I think this later comment already explains that :
`# Include Volta support if we're using CUDA9 or above`

Although we might want to update it to :
`# Include Volta/Turing support if we're using CUDA9 or above`

jakirkham:
From my perspective, this comment is a bit clearer. That said, I don't have strong feelings (it is just a comment :). Please feel free to merge or close as you see fit. Thanks for your help!

jakirkham:
How should we handle this @sjeaugey? Is this worth merging or would you prefer it be closed?

PR #222, opened 2019-05-08
allow arm64 for debian package

slug:


sjeaugey:
Can you elaborate on why we'd want to add ARM64 to the deb build ? Was NCCL tested on ARM64 ?

sjeaugey:
Fixed in 2.5.6

PR #232, opened 2019-06-13
V2.3.7 1 enable profile

sj6077:


sjeaugey:
I guess that PR was pushed by mistake.

Anyway, for reference, NCCL can be built with `-DPROFAPI` to let libraries intercept the NCCL calls (for statistics/profiling/anything), like with MPI.

The main symbols will then be called `pncclFunction` ; `ncclFunction` being a weak symbol pointing to `pncclFunction` (hence used unless intercepted by the profiling library).

So you can build a library that redefines `ncclFunction`, calling the real `pncclFunction` :
```
ncclResult_t ncclFunction(...args...) {
  update_stats(...args...);
  return pncclFunction(...args...);
}
```
And link with that library before nccl in your program.

PR #233, opened 2019-06-21
Fix out-of-bounds read in ncclStrToCpuset

flx42:
The affinityStr string was not null-terminated but was passed to strlen(3).

Signed-off-by: Felix Abecassis <fabecassis@nvidia.com>


AddyLaddy:
Thanks Felix we'll look at merging that in. Looks pretty straight forward.

sjeaugey:
Indeed, thanks for the fix. Merging.

PR #236, opened 2019-07-02
Add the exact matching modifier support "=" to the NCCL_IB_HCA variable

drpnd:
This PR is to add the exact matching support with the modifier “=“ to the NCCL_IB_HCA variable in order to solve a problem with the current prefix-based matching.

Background:
We have encountered the problem that we cannot specify the exact HCAs in the NCCL_IB_HCA variable when we have more than 10 HCAs with the same prefix (e.g., `mlx5_`).

For example, assuming we have `mlx5_1` and `mlx5_11`, and try to use only `mlx5_1`, we may set the variable as `NCCL_IB_HCA=“mlx5_1:1”`.  However, `mlx5_1:1` breaks down into interface `mlx5_1` and port `1` by `parseStringList()`.  Then,  `mlx5_1` is matched as a prefix, and consequently, it matches `mlx5_11` as well as `mlx5_1`.

I understand that this behavior is as-is documented, but we would like to specify the exact name of the interfaces in the NCCL_IB_HCA variable.

Summary:
This PR adds a new modifier “=“ to the NCCL_IB_HCA and NCCL_SOCKET_IFNAME variables, and passes the information to the `matchIfList()` function (through the `exactMach` bool argument) .  When this modifier is specified, exact matching `matchIf()` is performed rather than the original prefix matching `matchPrefix()` in the `matchIfList()` function.

The modifier is allowed to be specified at the beginning of the variables with the same way as the “^” modifier.  Therefore, this modifier cannot be used with the “^” modifier.
Note that we haven’t made any modifications to the rest of the variables for backward compatibility.

### DIFF
--- a/src/include/socket.h
+++ b/src/include/socket.h
@@ -66,6 +66,7 @@ static int findInterfaces(const char* prefixList, char* names, union socketAddre
 #endif
   struct netIf userIfs[MAX_IFS];
   bool searchNot = prefixList && prefixList[0] == '^';
+  bool searchExact = prefixList && prefixList[0] == '=';
   int nUserIfs = parseStringList(prefixList, userIfs, MAX_IFS);
 
   int found = 0;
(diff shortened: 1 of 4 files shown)


sjeaugey:
Thanks for the patch. I proposed a small code simplification as a PR on your branch. Other than that it looks good to me.

drpnd:
Thanks! I've merged your PR to the branch.

PR #242, opened 2019-07-31
Refine RPM package building spec file.

lowintelligence:
Add /sbin/ldconfig into RPM package install operations.

NCCL RPM packages built with current spec file would not update the ldconfig cache.

When building TensorFlow from source, its 'configure.py' script would try to find NCCL library installation path automatically by searching ldconfig cache. This search would fail if we only installed the NCCL RPMs, thus NCCL installation path needs to be explicitly set.

This commit try to make a improvement by adding the post-installation ldconfig run into the RPM build spec.


sjeaugey:
Thanks for the PR !

PR #271, opened 2019-12-06
Fix clang build

chsigg:
Clang doesn't understand `optimize("O0")`. It has `noopt` instead, which GCC doesn't understand. Wrap the difference in a macro.

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -124,9 +124,17 @@ ncclResult_t ncclGetUniqueId(ncclUniqueId* out) {
 }
 
 // Prevent compiler from optimizing out these operations
-void __attribute__((optimize("O0"))) commPoison(ncclComm_t comm) {
+#ifdef __clang__
+#define NCCL_NO_OPTIMIZE __attribute__((noopt)))
+#else
+#define NCCL_NO_OPTIMIZE __attribute__((optimize("O0")))
+#endif
+
+void NCCL_NO_OPTIMIZE commPoison(ncclComm_t comm) {
   comm->rank = comm->cudaDev = comm->busId = comm->nRanks = -1;
 }
+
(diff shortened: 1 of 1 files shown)


sjeaugey:
@chsigg actually that change now makes compilation error out with clang :
```
init.cc:133:6: error: expected unqualified-id
void NCCL_NO_OPTIMIZE commPoison(ncclComm_t comm) {
     ^
init.cc:128:48: note: expanded from macro 'NCCL_NO_OPTIMIZE'
#define NCCL_NO_OPTIMIZE __attribute__((noopt)))
                                               ^
init.cc:176:3: error: use of undeclared identifier 'commPoison'
  commPoison(comm);
  ^
2 errors generated.
```
Do you also see this ? (I'm using clang 6.0.0)

sjeaugey:
I fixed it (there was an extra ")", but I still get this warning :
```
init.cc:133:6: warning: unknown attribute 'noopt' ignored [-Wunknown-attributes]
void NCCL_NO_OPTIMIZE commPoison(ncclComm_t comm) {
     ^
init.cc:128:41: note: expanded from macro 'NCCL_NO_OPTIMIZE'
#define NCCL_NO_OPTIMIZE __attribute__((noopt))
                                        ^
1 warning generated.
```

chsigg:
Hi Sylvain, I'm sorry about my sloppy PR. I typed it up in the GitHub web UI without testing, which was obviously a bad idea. I made another PR to fix the fix. Thanks!

sjeaugey:
Yeah you probably should not write code in the GitHub web UI. It's easier for us if we can trust you blindly on the clang part :-)

chsigg:
Yup, lesson learned. I did test the latest PR.

PR #272, opened 2019-12-06
Fix clang warning

chsigg:
Add extra brazes to initialize multi-dimensional array.


sjeaugey:
Oops, I inadvertently pushed a fix to this issue together with commit aa15dfb. I think we can close this now.

Issue #287, opened 2020-01-26
Variable name typo? "slice" must be "chunk" in AllReduceRingKernel


aokomoriuta:
https://github.com/NVIDIA/nccl/blob/3701130b3c1bcdb01c14b3cb70fe52498c1e82b7/src/collectives/device/all_reduce.h#L40

This variable name seems to be "chunk", not "slice".

According to later lines, e.g., https://github.com/NVIDIA/nccl/blob/3701130b3c1bcdb01c14b3cb70fe52498c1e82b7/src/collectives/device/all_reduce.h#L43-L47, the "slice" variable is set to the rank number.
And the "offset", which is used to calculate the head pointer of sending data whose size is chunkSize, is calculated as `chunkOffset + slice * chunkSize`.

Those must say that the "slice" variable represents a **chunk** number, not a slice number.
Generally speaking, a expression `offsetX + numberX*sizeX` naturally must represent the head of X.

Additionally, the term "slice" means the number of send/recv on the `primitives::GenericOp` https://github.com/NVIDIA/nccl/blob/299c554dccf923230321ad7495946543f3e9b457/src/collectives/device/primitives.h#L189 .
The meaning is different from the code on the `AllReduceRingKernel`.

Am I wrong? or may I send PR to fix this?

sjeaugey:
Indeed, `chunk` might be a better name for it. I don't remember the history for that, but in any case, I won't complain if the code is easier to understand.

Feel free to submit a PR; just make sure to check the other operations as well (reduce_scatter and all_gather in particular).

sjeaugey:
Merged. Sorry for the delay.

Issue #275, opened 2019-12-10
NCCL hangs in P2P


AHEADer:
In init.cc, NCCL will do the [connect](https://github.com/NVIDIA/nccl/blob/master/src/init.cc#L422) operation with NCCLCHECK. However, it will only report the error in logs but will still return ncclSuccess to the caller. In p2p.cc, if this [failure](https://github.com/NVIDIA/nccl/blob/master/src/transport/p2p.cc#L168)  occurs, its return value cannot be propagated to the user application, which will make the application hang for a long time(maybe forever?)

sjeaugey:
This is not normal. The error should be reported to the original ncclCommInitRank call or the ncclGroupEnd call, if used. Can you share an example and log where there is an error but you get a success ?

AHEADer:
I ran tensorflow with horovod and nccl, but GPUs(1080Ti) in my cluster already had too many P2P connections. Then my job hanged there, only printing "failed to open CUDA IPC handle".

AHEADer:
Actually you can look at which function invoke send/connect functions. The function which invokes send/connect functions just does NCCLCHECK for this ncclInternalError but it will not report to functions that invoke it, on the contrary, it will only return ncclSuccess even one of its NCCLCHECK fails.

sjeaugey:
Sorry for responding late. I'm not sure where we return ncclSuccess instead of an error.

If `cudaIpcOpenMemHandle` fails, it should return `ncclUnhandledCudaError`. See `p2pSendConnect` here : https://github.com/NVIDIA/nccl/blob/master/src/transport/p2p.cc#L244 and `p2pRecvConnect` here : https://github.com/NVIDIA/nccl/blob/master/src/transport/p2p.cc#L275.

NCCLCHECK should then propagate the error all the way up to the `ncclCommInitRank`/`ncclCommInitAll` call. Setting `NCCL_DEBUG=INFO`, you should see the whole backtrace until the `ncclCommInit*` call.

stephenmsachs:
### Issue Cleanup: Helping Us Focus on Current Challenges

We're reviewing older issues to ensure we prioritize the most relevant and active ones. Since this issue hasn't seen updates in over 6 months, we'll be closing it for now.

*This change helps us focus our efforts on addressing any current issues our users are facing.* If this issue still affects you, please don't hesitate to reopen it with a quick update (e.g., "Still relevant on [version=X]").
Thanks for your understanding and for contributing to NCCL.

PR #288, opened 2020-01-29
Fix wrong variable name "slice" to "chunk"

aokomoriuta:
As #287

There's no variable named "slice" in other operations (all_gather, broadcast, reduce, and reduce_scatter)


aokomoriuta:
@sjeaugey Thank you for your response on #275 . I'd like to review my commit to merge.

PR #297, opened 2020-03-03
Update init.cc: NCCL_LAUNCH_MODE support set to GROUP

tanguofu:
In current init comm->launchMode, the NCCL_LAUNCH_MODE take no effect  when set to GROUP， so there is no way to force comm->launchMode is ncclComm::GROUP when  comm->intraRanks == 1

this PR want to  make  NCCL_LAUNCH_MODE  envs  as the first just principle to set comm->launchMode.

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -369,9 +369,14 @@ ncclResult_t ncclCommSetIntra(struct ncclComm* comm, int rank, int ranks, struct
   // Set CG Mode
   comm->launchMode = ncclComm::GROUP;
   char* str = getenv("NCCL_LAUNCH_MODE");
-  if (comm->intraRanks == 1 || (str && strcmp(str, "PARALLEL") == 0)) {
-    comm->launchMode = ncclComm::PARALLEL;
+  if (str != NULL){
+     comm->launchMode = strcmp(str, "PARALLEL") == 0 ? ncclComm::PARALLEL : ncclComm::GROUP;
+     TRACE(NCCL_INIT,"set launchMode to %s by envionment NCCL_LAUNCH_MODE=%s",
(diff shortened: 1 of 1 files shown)


sjeaugey:
Is there a reason to use group launch when launching on a single GPU ? Parallel mode is supposed to be faster and if there is a single GPU per process, it is just as safe as group mode.

tanguofu:
we found there is hangup when run with horovod that each process only run with one gpu.
maybe there is some mistake in our train code,  but after change the NCCL_LAUNCH_MODE to group, there is no  hang any more.

so it is better to give user to change NCCL_LAUNCH_MODE to group will be helpful them to debug i think

sjeaugey:
Well, it's likely that GROUP launch is causing less parallelism compared to PARALLEL launch, which means scheduling on the GPU is different and the hang doesn't happen. But it could be that the hang still happens, just less frequently.

I don't really see how GROUP launch would be a "safer" mode for debug... but unless we understand the real reason for the hang we don't really know.

Are you using multiple NCCL communicators (either modifying Horovod or having another part of the code creating NCCL communicators as well) ?

tanguofu:
we do not using multiple NCCL communicators,  just use horovod 0.19.0  with NCCL 2.5.7.1, cuda10.0   
and the GPU is V100*8 with NVLink. 

the hang look like  a little suddenly， maybe cause by mellonx rdma， may be by pci-e driver in motherboard， it take us many days to digger， but we did not know the real reason.. 

if  nccl can give some hang check  that will be very helpful

sjeaugey:
If there is a communication problem, you should see a `WARN` message (provided you set `NCCL_DEBUG=WARN` or `INFO`), so that is quite easy to figure out, and definitely the first thing to do when debugging, if not done already.

That would also print warnings in case of size mismatch caused by mis-ordering of operations between different ranks.

Issue #310, opened 2020-03-23
Can we remove CUDA dependencies from nccl_net.h?


capcah:
Currently `nccl_net.h` pulls `nccl.h` only for using ncclResult_t, as far I understand. However, nccl.h pulls in <cuda_runtime.h> and <cuda_fp16.h>, which means that plugins can only be built in machines with the CUDA headers available.

We could remove this dependency by splitting `nccl.h.in` into two:  `types.h.in` and `nccl.h`. The first would contain the general type definitions (everything above the [collective comms ops](https://github.com/NVIDIA/nccl/blob/master/src/nccl.h.in#L119)), and the second would include the first and contain the rest of the file. With this split, only `nccl.h` needs the CUDA headers.

This way we can make the .in file smaller, separate API from types, and allow `nccl_net.h` to include only `types.h` without pulling cuda headers in. What do y'all think?

sjeaugey:
Indeed we could probably split them into two parts.

Alternatively, since the API is versioned, you can copy the definitions from nccl_net.h and nccl.h in your project, and remove what you don't need. The definitions won't change in the version you choose to implement. When you want to support a new version, you can update your includes to the new definitions (or even add a new version of the struct along the other if you want to support multiple versions).

capcah:
I sent a pull request implementing a version of this, let me know if this works for you. Copying the file into our projects works, but at Google we prefer to track upstream repositories as much as possible, so if you think that this PR (or something similar, open to suggestions!) works, we'd prefer to get that upstreamed.

sjeaugey:
Thanks. The patch looks good to me. We're pretty busy on other things right now so it might take us a week or two before we can look into this; let me know if you need this merged quicker than that.

Also don't hesitate to ping us again if nothing happened in two weeks from now.

capcah:
That's alright, I can keep the work on my side. Should I also send PRs for the 2.6 branch?

sjeaugey:
It would be indeed better to base your patches on 2.6. But hopefully, given those parts didn't change much in 2.6, the rebase should be easy.

stephenmsachs:
### Issue Cleanup: Helping Us Focus on Current Challenges

  We're reviewing older issues to ensure we prioritize the most relevant and active ones. Since this issue hasn't seen updates in over 6 months, we'll be closing it for now.

[...]

PR #312, opened 2020-03-25
Move the NCCL types to its own file.

capcah:
This allows us to remove CUDA dependencies from nccl_net.h as described in
https://github.com/NVIDIA/nccl/issues/310.


capcah:
Hey @sjeaugey, could you take a look at this PR?

sjeaugey:
Sorry for not responding earlier. I looked at it and wanted to solve that problem but as usual got caught by something else.

If the file is named nccl_types.h then I'd expect to only see types in there. It seems you put everything that didn't have CUDA streams in it, including some of the functions of the API (ncclCommInitRank, ncclGetUniqueId, ...).

Also there is the question of the NCCL_VERSION definitions and where they should reside. I could see why you'd want to have it in this file.

capcah:
> If the file is named nccl_types.h then I'd expect to only see types in there. It seems you put everything that didn't have CUDA streams in it, including some of the functions of the API (ncclCommInitRank, ncclGetUniqueId, ...).

That's correct, my bad. Fixed it.

> Also there is the question of the NCCL_VERSION definitions and where they should reside. I could see why you'd want to have it in this file.

I think it makes sense to move them to the types file so that the plugin can also access it without pulling the CUDA dependencies. Another minor improvement here is that the API on nccl.h is actually treated as a C file for purposes of syntax highlighting, indexing, etc.

PR #313, opened 2020-03-26
Allow specifying a different network plugin.

capcah:
For this we introduce a new mechanism for specifying string NCCL
parameters and change the name of the library where dlopen happens.

We also introduce a special behavior when passed an empty plugin name,
namely loading the plugin from the same executable that is running NCCL.
This is useful for the cases where the plugin and NCCL are statically
linked in the same binary. This is achieved by passing NCCL to dlopen,
which will cause it to return a handle to the main program.

### DIFF
--- a/src/include/param.h
+++ b/src/include/param.h
@@ -77,4 +77,24 @@ int64_t ncclParam##name() { \
   return value; \
 }
 
+
+// Returns a pointer to NULL-terminated static string.
+#define NCCL_STR_PARAM(name, env, default_value) \
+pthread_mutex_t ncclParamMutex##name = PTHREAD_MUTEX_INITIALIZER; \
+char* ncclParam##name() { \
+  static_assert(default_value != NULL, "default value cannot be NULL"); \
+  static char* value = NULL; \
+  pthread_mutex_lock(&ncclParamMutex##name); \
+  if (value == NULL) { \
+    value = default_value; \
+    char* str = secure_getenv("NCCL_" env); \
+    if (str != NULL) { \
+      value = str; \
+      INFO(NCCL_ALL,"%s set by environment to %lu.", "NCCL_" env, value);  \
+    } \
+  } \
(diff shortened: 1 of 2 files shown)


src/include/param.h
```diff
@@ -77,4 +77,24 @@ int64_t ncclParam##name() { \
...
+char* ncclParam##name() { \
+  static_assert(default_value != NULL, "default value cannot be NULL"); \
+  static char* value = NULL; \
+  pthread_mutex_lock(&ncclParamMutex##name); \
+  if (value == NULL) { \
+    value = default_value; \
+    char* str = secure_getenv("NCCL_" env); \
+    if (str != NULL) { \
+      value = str; \
+      INFO(NCCL_ALL,"%s set by environment to %lu.", "NCCL_" env, value);  \
```
sjeaugey, line 93: Shouldn't that be "set by environment to %s" ?

capcah:
Fixed and rebased the change to head.

sjeaugey:
Recent NCCL versions now have a `NCCL_NET_PLUGIN` environment variable which can be set to specify a suffix to the NCCL plugin name.

That means users can now install multiple plugins, and use a symlink to libnccl-net.so as the default one, e.g.:
```
ls -l /path/to/nccl-plugins/
total 852
-rwxr-xr-x 1 user users 289520 Jan 21 03:08 libnccl-net-ib.so
lrwxrwxrwx 1 user users     17 Jan 21 03:09 libnccl-net.so -> libnccl-net-ib.so
-rwxr-xr-x 1 user users 289520 Jan 21 03:09 libnccl-net-aws.so
-rwxr-xr-x 1 user users 289520 Jan 21 03:09 libnccl-net-fastsocket.so
```
Then users can set `NCCL_NET_PLUGIN=aws` or `NCCL_NET_PLUGIN=fastsocket` to select a different one.

Issue #309, opened 2020-03-23
Improvement in the documentation of the plugin interface.


capcah:
Hello all,

I'm implementing a plugin using the [nccl_net.h API](https://github.com/NVIDIA/nccl/blob/master/src/include/nccl_net.h) and I found the documentation of the functions pretty hard to understand. If you folks don't mind, I'll start sending PRs to improve the documentation there, based on my current understanding of the interface.

Additionally, there are some things that are not clear to me even after reading the code, such as: _In [`test()`](https://github.com/NVIDIA/nccl/blob/master/src/include/nccl_net.h#L95), can the plugin disregard `size` if the request is not complete or should it contain the size of partially downloaded data?_

Let me know if you want me to work on this.

sjeaugey:
The size needs only to be set when the request is complete.

Indeed the documentation is not as good as it could be; it's pretty much only the comments in nccl_net.h, the rest being in the implementation. The discovery / topology detection functions are used by src/graph/topo.cc, while the connection/communication part only happens in  src/transport/net.cc (communication is all inside send/recv proxy functions).

We never put a lot of effort in documenting this, since it was not intended to be used by users, more by vendors/partners which we usually work closer with. Better documentation is always better though, so clarifying things won't hurt.

Also please note the API is evolving a bit version after version. Make sure you check the v3 structure here for NCCL 2.6 : https://github.com/NVIDIA/nccl/blob/v2.6/src/include/nccl_net.h.

Also we're interested in knowing who is developing and maintaining plugins to that we can give heads up and pig you when we make the API evolve (like in 2.6).

capcah:
Hey Sylvain,

Thanks for the clarification! I'll reach out over email with some contact points in our side.

Regarding the API, there are a couple more things that are not clear:

1. How does exactly the `handle` argument in connect() and listen() work? As it understand, it is initialized during listen() and accessed during connect(), but that seems odd: Hows is the code calling connect() aware of what the handle is supposed to be?
2.  For RMA memory registration, AFAIK, NCCL calls regMr() with a block of memory that to be registered for remote access. Is there any way of giving NCCL a pointer to a memory region to use for its buffers instead or some way of controlling those addresses from the plugin?

Thanks!

sjeaugey:
1. Between listen() and connect(), NCCL exchanges the handle through a side channel (sockets).
[...]

PR #327, opened 2020-04-17
Extend network plugin to support allocations.

capcah:
Implement Alloc/Free functions to give more control over the memory
being allocated to the memory plugins, as discussed on https://github.com/NVIDIA/nccl/issues/309

These new functions are unsed instead of the memory registration
functions, so I decided to make these changes backwards-compatible by
requiring that plugins using the alloc/free calls set the reg/unreg
field to NULL.

The memory returned by alloc/free have some requirements, which are
documented in nccl_net.h.

### DIFF
--- a/src/include/alloc.h
+++ b/src/include/alloc.h
@@ -10,12 +10,21 @@
 #include "nccl.h"
 #include "checks.h"
 #include "align.h"
+#include "net.h"
 #include <sys/mman.h>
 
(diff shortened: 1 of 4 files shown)


capcah:
Gave a shot at a new alloc/free interface and tested it by replacing the regMr and deregMr calls in the net.cc API with it. What do you think?

PR #335, opened 2020-05-18
add warning to ncclGroupEnd with no ncclGroupStart

zhuzilin:


sjeaugey:
Actually this is already in the next version (see the preview of NCCL 2.7 on the p2p branch) :
https://github.com/NVIDIA/nccl/blob/p2p/src/group.cc#L147

We return ncclInvalidUsage. Granted, we could print a WARN to explain why and ease debug.

PR #346, opened 2020-06-24
Adding sm_75 support to the makefile for Turing

sandeepkumar-skb:
Noticed that there was no sm_75 support added in the makefile which is for Turing. Technically it should be fine with sm_70 but adding sm_75 support for Turing optimized SASS. 
Was there a specific reason for not adding sm_75?


sjeaugey:
The main reason is to limit the library size and compilation time. We only add more architectures when it's either needed or improves performance significantly. We didn't see a significant performance difference between sm_70 and sm_75 on Turing for the type of operations that NCCL does (basically load/stores and a few sums), so we didn't add it to the compilation list.

sandeepkumar-skb:
Thank you Sylvain. Closing PR

PR #350, opened 2020-07-02
Fix build action order

Riatre:
Add $(INCDIR)/nccl.h to build dependencies of $(LIBOBJ) and
$(DEVICELIB).

As there were no .d dep files during the first build, Make may kick off source
compilation before nccl.h got generated, which leads to occasional build
failures on systems with high core count. The build failure could be
reproduced reliably with a `sleep 5` in $(INCDIR)/nccl.h rule.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -46,7 +46,7 @@ lib : $(INCTARGETS) $(LIBDIR)/$(LIBTARGET) $(PKGDIR)/$(PKGTARGET)
 
 staticlib : $(LIBDIR)/$(STATICLIBTARGET)
 
-$(DEVICELIB): ALWAYS_REBUILD
+$(DEVICELIB): ALWAYS_REBUILD $(INCTARGETS)
 	$(MAKE) -C collectives/device
 
 # Empty target to force rebuild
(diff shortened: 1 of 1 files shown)


sjeaugey:
Thanks much for the report ! That was a tricky race condition.

Wouldn't it be easier to just add include dependencies on the CC/CU compilation directives (see below patch) and also use INCTARGETS ? Kind of makes sense to me as those files include the files in build/include, and even if include files were not generated, they'd still need to be copied there.

```
diff --git a/src/Makefile b/src/Makefile
index d065888a..487f7909 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -46,7 +46,7 @@ lib : $(INCTARGETS) $(LIBDIR)/$(LIBTARGET) $(PKGDIR)/$(PKGTARGET)
 
 staticlib : $(LIBDIR)/$(STATICLIBTARGET)
 
-$(DEVICELIB): ALWAYS_REBUILD
+$(DEVICELIB): ALWAYS_REBUILD $(INCTARGETS)
        $(MAKE) -C collectives/device
 
 # Empty target to force rebuild
@@ -107,7 +107,7 @@ $(PKGDIR)/%.pc : %.pc
        mkdir -p $(PKGDIR)
        install -m 644 $< $@
 
-$(OBJDIR)/%.o : %.cc
+$(OBJDIR)/%.o : %.cc $(INCTARGETS)
        @printf "Compiling  %-35s > %s\n" $< $@
        mkdir -p `dirname $@`
        $(CXX) -I. -I$(INCDIR) $(CXXFLAGS) -Iinclude -c $< -o $@
```

Riatre:
Thanks! I've updated the PR. That's much simpler. I was worrying about excessive dependency to nccl.h and nccl_net.h but turns out this is a non-issue as nearly all .cc files do include these two header files.

PR #364, opened 2020-08-03
Don't require NIC devices to have specific PCI class

badgerious:
If a PCI node is the parent of a NIC, treat it as such, regardless of
the PCI class code for the device. This allows non-traditional devices
to act as NICs via the net plugin mechanism.

### DIFF
--- a/src/graph/topo.cc
+++ b/src/graph/topo.cc
@@ -317,21 +317,19 @@ ncclResult_t ncclTopoAddPci(struct ncclXmlNode* xmlPci, struct ncclTopoSystem* s
   NCCLCHECK(busIdToInt64(str, &busId));
 
   struct ncclTopoNode* node = NULL;
-  if (type == GPU) {
-    struct ncclXmlNode* xmlGpu;
-    NCCLCHECK(xmlGetSub(xmlPci, "gpu", &xmlGpu));
-    if (xmlGpu == NULL) return ncclSuccess;
+  struct ncclXmlNode* xmlGpu = NULL;
+  NCCLCHECK(xmlGetSub(xmlPci, "gpu", &xmlGpu));
+  if (xmlGpu != NULL) {
     int index;
     NCCLCHECK(xmlGetAttrIndex(xmlGpu, "rank", &index));
     if (index == -1) return ncclSuccess;
     NCCLCHECK(ncclTopoCreateNode(system, &node, type, busId));
     NCCLCHECK(ncclTopoAddGpu(xmlGpu, system, node));
   }
-  if (type == NIC) {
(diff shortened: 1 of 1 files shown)


badgerious:
If the net device provided by the net plugin doesn't match the expected PCI class code, NCCL fails like this:

```
mantaray:5351:5359 [0] NCCL INFO Setting affinity for GPU 0 to 0fffff
mantaray:5351:5359 [0] NCCL INFO graph/paths.cc:297 -> 3
mantaray:5351:5359 [0] NCCL INFO transport/net.cc:111 -> 3
mantaray:5351:5359 [0] NCCL INFO transport.cc:30 -> 3
mantaray:5351:5359 [0] NCCL INFO transport.cc:49 -> 3
mantaray:5351:5359 [0] NCCL INFO init.cc:766 -> 3
mantaray:5351:5359 [0] NCCL INFO init.cc:840 -> 3
mantaray:5351:5359 [0] NCCL INFO group.cc:73 -> 3 [Async thread]
```
Since `ncclTopoAddNic()` was never called for the NIC, `ncclTopoIdToIndex()` will find `system->nodes[NIC].count` is 0.

I'll admit I've had trouble understanding why the class code filtering in `ncclTopoAddPci()` is needed, but if I've missed something, hopefully someone can provide some guidance to improve the patch.

sjeaugey:
Thanks for the PR. Makes sense.

snyjm-18:
@badgerious  can you please apply the attached patch to the PR. This just does the same thing for GPUs as well as NICs so we keep it consistent.
`diff --git a/src/graph/topo.cc b/src/graph/topo.cc
index b9dd43e..a917340 100644
--- a/src/graph/topo.cc
+++ b/src/graph/topo.cc
@@ -317,10 +317,9 @@ ncclResult_t ncclTopoAddPci(struct ncclXmlNode* xmlPci, struct ncclTopoSystem* s
   NCCLCHECK(busIdToInt64(str, &busId));
 
   struct ncclTopoNode* node = NULL;
-  if (type == GPU) {
-    struct ncclXmlNode* xmlGpu;
-    NCCLCHECK(xmlGetSub(xmlPci, "gpu", &xmlGpu));
-    if (xmlGpu == NULL) return ncclSuccess;
+  struct ncclXmlNode* xmlGpu = NULL;
+  NCCLCHECK(xmlGetSub(xmlPci, "gpu", &xmlGpu));
+  if (xmlGpu != NULL) {
     int index;
     NCCLCHECK(xmlGetAttrIndex(xmlGpu, "rank", &index));
     if (index == -1) return ncclSuccess;
`

badgerious:
Patch updated to treat GPUs similarly.

PR #435, opened 2020-12-07
x86: Add CPU detection for Zhaoxin processors

zzjianhui:
Add ZX’s vendor id, including CentaurHauls and Shanghai.
Set p2pLevel to PATH_PXB.
Set the bandwidth between the two sockets in the numa node to 6.

Signed-off-by: Jonas Zhou <JonasZhou@zhaoxin.com>

### DIFF
--- a/src/graph/paths.cc
+++ b/src/graph/paths.cc
@@ -283,6 +283,9 @@ ncclResult_t ncclTopoCheckP2p(struct ncclTopoSystem* system, int64_t id1, int64_
     if (model == NCCL_TOPO_CPU_TYPE_BDW) p2pLevel = PATH_PXB;
     else p2pLevel = PATH_PHB;
   }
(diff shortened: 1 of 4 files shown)


sjeaugey:
Thanks, the patch looks great.

One minor comment: could it be a better strategy to apply that code to all CPUs from Zhaoxin, not only to the Lujiazui model ? In general, we prefer doing so, since the next CPU model will more probably prefer the Zhaoxin settings rather than the default ones. Plus, it would make the patch smaller. We can add a distinction later on, if we need different code for the next CPU model.

zzjianhui:
Hi @sjeaugey , set a default value `ZIP_WIDTH`. Add micro-architecture YONGFENG.

zzjianhui:
Hi @sjeaugey , do you have any suggestions, thank you very much

zzjianhui:
Hi @sjeaugey , we modified the code so that it can be used for all Zhaoxin CPUs. Added the next Yongfeng models code and set the `width` to 9. Are there any other questions?

sjeaugey:
Sorry to have kept you waiting. Merged.

PR #437, opened 2020-12-08
Add early break in setupLaunch

zhuzilin:
This PR changed:
```c++
  for (int c=0; c<comm->p2pnChannels; c++) {
    if (comm->channels[c].workCount) params->gridDim.x = c+1;
  }
```
to 
```c++
  for (int c=comm->p2pnChannels-1; c>=0; c--) {
    if (comm->channels[c].workCount) {
      params->gridDim.x = c+1;
      break;
    }
  }
```
to shorten the loop.

Thank you for your time on reviewing this PR.

### DIFF
--- a/src/enqueue.cc
+++ b/src/enqueue.cc
@@ -110,8 +110,11 @@ static ncclResult_t getNextOp(struct ncclChannel* channel, struct ncclWork** wor
 
 static ncclResult_t setupLaunch(struct ncclComm* comm, struct cudaLaunchParams* params) {
   // Only launch blocks where we have work to do.
-  for (int c=0; c<comm->p2pnChannels; c++) {
-    if (comm->channels[c].workCount) params->gridDim.x = c+1;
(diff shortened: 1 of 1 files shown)


sjeaugey:
Can you explain what is the goal of this PR ? Is it fixing a bug ?

Besides, the only reason for this loop is to find the max channel index which has work to do. Stopping at the first one would break that logic.

zhuzilin:
@sjeaugey The goal is to reduce some unnecessary iteration. I changed the loop direction to check from the largest channel instead of from channel 0, so it won't break the logic.

sjeaugey:
Well given it adds code, it would need to be counterbalanced by a visible performance improvement. Do you see a difference in performance ?

zhuzilin:
@sjeaugey I just think this is a better way to do this loop... And as for code increasement, this would only add one line (the break), so I wonder why we need to see a visible performance improvement for a one line change...

PR #468, opened 2021-02-18
[Cleanup] Remove unused computation on basePath

void-main:
In paths.cc, there's a `getPath` from `baseNode`, but the computed `basePath` is never used.

So I removed these lines of code to avoid useless computations.


sjeaugey:
Actually reverted that patch. We need that code to initialize the base node in the topology graph.

Issue #549, opened 2021-08-10
will ncclCommAbort fail to abort communications in some cases?


wangraying:
Hi, I started a group of processes to perform allreduce operations. Each process started another thread to call `ncclCommAbort` at certain timepoint. 

It is expected that all processes will eventually exit, but I found that sometimes, the job will accidentally hang. (Rarely happens but it does)

Does it make sense? 

Thanks for your help.

sjeaugey:
It should not hang. Can you see where the process is hung using `gdb`?

wangraying:
@sjeaugey 

Yes, I checked the hanged process using `gdb`.

It seems two threads are waiting on  `clock_gettime`.

Thread 1:
```
(gdb) where 
#0  0x00007ffee4abb6c2 in clock_gettime ()
#1  0x00007f979399ed36 in __GI___clock_gettime (clock_id=4, tp=0x7ffee4a8abb0) at ../sysdeps/unix/clock_gettime.c:115
#2  0x00007f96fee2b70e in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#3  0x00007f96fef04837 in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#4  0x00007f96fedccb6c in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#5  0x00007f96fed02f4f in ?? () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#6  0x00007f96fee87259 in cuMemFree_v2 () from /usr/lib/x86_64-linux-gnu/libcuda.so.1
#7  0x00007f97160362c0 in cudart::cudaApiFree(void*) () from /home/wangrui08/bagua-sys/bagua-core/python/bagua_core/.data/lib/libnccl.so
#8  0x00007f97160683e1 in cudaFree () from /home/wangrui08/bagua-sys/bagua-core/python/bagua_core/.data/lib/libnccl.so
#9  0x00007f9715fb2b06 in commFree (comm=0x55d5d2790840) at init.cc:204
#10 commDestroy (comm=comm@entry=0x55d5d2790840) at init.cc:990
#11 0x00007f9715fb9a15 in ncclCommAbort (comm=0x55d5d2790840) at init.cc:1026
```

Thread 7
```
(gdb) bt  
#0  0x00007ffee4abb6c2 in clock_gettime ()
[...]

Issue #279, opened 2019-12-27
ncclCommGetAsyncError doesn't report errors for failures within a host.


pritamdamania87:
For reference I'm running two instances of the following script:

```
from __future__ import absolute_import, division, print_function, unicode_literals

import torch.distributed as c10d
import torch
import argparse
import os
import logging
logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s', level=logging.INFO)

if __name__ == "__main__":
    parser = argparse.ArgumentParser(
        description='Simple script to simulate NCCL errors. The script is '
        'supposed to be run on multiple different nodes simultaneously with '
        'appropriate rank and world_size. The script run an allreduce() on '
        'the rank 0 node and aborts all the other nodes to simulate an error '
        'in NCCL')
    parser.add_argument('addr', help='address of the master node to connect to.')
    parser.add_argument('port', help='port of the master node to connect to.')
    parser.add_argument('rank', help='rank of this node')
    parser.add_argument('world_size', help='number of nodes in process group')
    args = parser.parse_args()
    rank = int(args.rank)
    world_size = int(args.world_size)
    port = int(args.port)

    store = c10d.TCPStore(args.addr, port, world_size, rank == 0)
    process_group = c10d.ProcessGroupNCCL(store, rank, world_size)
    logging.info('Running first allreduce')
    process_group.allreduce(torch.rand(10).cuda(rank)).wait()
    if rank == 0:
        logging.info('Running second allreduce only on rank 0')
        work = process_group.allreduce(torch.rand(10).cuda(rank))
        logging.info('Waiting for allreduce to complete...')
        work.wait()
        logging.info('Second allreduce successful: {}'.format(work.is_success()))
    else:
        logging.info('Aborting all other ranks.')
        os.abort()
```

I run two instances of the script as follows:

```
NCCL_BLOCKING_WAIT=1 python test/simulate_nccl_errors.py <addr> <port> 0 2
NCCL_BLOCKING_WAIT=1 python test/simulate_nccl_errors.py <addr> <port> 1 2
```

The problem I run into is that when I run both instances on the same host, `ncclCommGetAsyncError` doesn't report an error and the operation eventually times out. Although, If I run them on separate hosts `ncclCommGetAsyncError` does report an error and we see a separate exception being thrown. For additional context, this is the PyTorch PR that uses `ncclCommGetAsyncError`: https://github.com/pytorch/pytorch/pull/25012.

[...]

PR #564, opened 2021-09-08
add ncclCommMarkAbort api in convienence of upper ai system controller

woodlgz:
as [issue 279](https://github.com/NVIDIA/nccl/issues/279)  and [issue 549](https://github.com/NVIDIA/nccl/issues/549) discussed, it's up to upper ai system controller to handle error when some rank losts.

it 's common for upper ai system controller to call ncclCommAbort to  abort communicator related operations when rank lost situation is detected.
however, in system  where multiple nccl communicators get involved, say horovod with HOROVOD_NUM_NCCL_STREAMS set to above 1, when lost rank is detected,  trying to abort or destroy all these communicators will cause system hang. It's due to cuda implicit synchronization. 
Image a scenario with 2 communicator (say A,B), system tries to abort all nccl communicators in sequence of A,B. If cuda kernel related to communicator B keeps running and won't exit until abort is detected while controller try to free communicator A(cudaFree involved), deadlock is met. 

mark all communicator as aborted and give up all operations first, and release related resources later can avoid above problem.

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -1026,6 +1026,15 @@ ncclResult_t ncclCommAbort(ncclComm_t comm) {
   return commDestroy(comm);
 }
 
+NCCL_API(ncclResult_t, ncclCommMarkAbort, ncclComm_t comm);
+ncclResult_t ncclCommMarkAbort(ncclComm_t comm) {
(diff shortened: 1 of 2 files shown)


woodlgz:
@alsrgv take a look at this please

Issue #555, opened 2021-08-17
[Rem Allocator] Allocation failed 


xutianming:
![image](https://user-images.githubusercontent.com/4970790/129704114-2cc94f1d-a069-443e-9322-99495104708b.png)

When training with A100 GPUs , I encountered OOM error of `Rem Allocator`. As far as I know, the remote allocator is only used to do send/receive between GPUs which do not have a direct NVLink connection but can communicate through an intermediate GPU. 

But in my case, the GPUs are fully connected through NVLink. Why did I still meet this error?
cc @sjeaugey 

![image](https://user-images.githubusercontent.com/4970790/129704879-6d8c1b14-cbdd-4bdb-b45b-95e12f515d8d.png)

sjeaugey:
Is that issue reproducible or did it only happen once? It seems the remote allocation thread is being sent a message that was not for it, hence experiences random errors: socket closing too soon, then first 64 bits sent -- the size -- being random data hence beyond the GPU memory size.

Would you have more of the log, in particular what happens for the main threads after they mistakenly communicate with the remote memory allocation thread? (main threads are the ones printing "NCCL INFO ... via P2P/IPC/read")

xutianming:
@sjeaugey The issue rarely occurred, and I only met once. The main threads continued working without more logs. So the error can be ignored safely, right?

I am trying to reproduce it with NCCL_DEBUG_SUBSYS=INIT,ALLOC for more logs.

sjeaugey:
No need to add `ALLOC` in the DEBUG_SUBSYS list, that would probably make it very large yet not more useful. Please keep NCCL_DEBUG_SUBSYS unset (default).

I'd like to see what the main thread was doing when it connected to the remote mem alloc thread, and see if there was any error after that that would indicate where it is. The only thin I have right now is what happened just before.

xutianming:
![image](https://user-images.githubusercontent.com/4970790/130195961-aefd3f64-370b-46ca-8e91-e3cad95e6435.png)
These are all the NCCL logs I got and the job continued without other errors. Since I got `Connection closed by remote peer` here, I guess there might be an unexpected failed socket/queue pair ?

sjeaugey:
Okay, it is surprising that the job managed to continue without errors ... Now if that's not NCCL connecting to the remote allocator (like, some random other service connecting to the wrong port), the remote allocator will simply ignore the request and indeed everything else will work just fine...

chr1sj0nes:
[...]

PR #599, opened 2021-11-15
Add logging and basic verification to remote allocator.

chr1sj0nes:
Provides basic protection against spurious connections to the remote allocator service, as seen in #555.

### DIFF
--- a/src/bootstrap.cc
+++ b/src/bootstrap.cc
@@ -224,9 +224,22 @@ struct extState {
 
 #define MAX_SEGMENTS 128
 
+// Magic number prefixed to remote allocation requests.
+static const int remAllocMagicNumber = 0xDEADC0DE;
+
 static ncclResult_t remoteAlloc(void** ptr, int fd, union socketAddress *addr) {
+  char socketName[SOCKET_NAME_MAXLEN+1];
+  int prefix;
+  NCCLCHECK(socketRecv(fd, addr, &prefix, sizeof(int)));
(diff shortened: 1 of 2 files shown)


src/bootstrap.cc
```diff
@@ -224,9 +224,22 @@ struct extState {
...
+static const int remAllocMagicNumber = 0xDEADC0DE;
+
 static ncclResult_t remoteAlloc(void** ptr, int fd, union socketAddress *addr) {
+  char socketName[SOCKET_NAME_MAXLEN+1];
+  int prefix;
+  NCCLCHECK(socketRecv(fd, addr, &prefix, sizeof(int)));
+  if (prefix != remAllocMagicNumber) {
+    INFO(NCCL_INIT, "[Rem Allocator] invalid request from %s",
+         socketToString(addr, socketName));
+    return ncclInternalError;
```
byronyi, line 237: Is it better to simply ignore when the magic string doesn't match?

eric-haibin-lin:
Is there an ETA for merging this PR?

sjeaugey:
The remote allocator has been completely replaced in NCCL 2.12, so unfortunately this PR can't be merged. The idea, however, remains valid, so this is something we'll consider.

sjeaugey:
Closing this PR as 2.12 has been pushed for some time now. If the issue persists, please update #555 and we'll try to fix it in 2.12.

PR #600, opened 2021-11-18
Perform `busIdToInt64` on the stack.

chr1sj0nes:
I noticed when I enabled `NCCL_DEBUG_SUBSYS=ALLOC` that this function is
called thousands of times, making the log output unintelligible.
Fortunately, this function can be implemented without heap allocations.


sjeaugey:
Thanks Chris for the PR, indeed, it's better to avoid lots of unnecessary allocs.

PR #602, opened 2021-11-24
Add env NCCL_NET_DISABLE_INTRA

kwen2501:
Disable NET transport for intra-node communication by setting the env to 1
It provides an option to error out instead of falling back to NET when superior intra-node transports (P2P and SHM) are unavailable

### DIFF
--- a/src/transport.cc
+++ b/src/transport.cc
@@ -36,8 +36,8 @@ static ncclResult_t selectTransport(struct ncclComm* comm, struct ncclTopoGraph*
       return ncclSuccess;
     }
   }
-  WARN("No transport found !");
-  return ncclInternalError;
+  WARN("No transport found for rank %d[%lx] -> rank %d[%lx]", myInfo->rank, myInfo->busId, peerInfo->rank, peerInfo->busId);
+  return ncclSystemError;
 }
 
 ncclResult_t ncclTransportP2pConnect(struct ncclComm* comm, struct ncclChannel* channel, int nrecv, int* peerRecv, int nsend, int* peerSend, int connIndex) {
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -56,8 +56,18 @@ struct netRecvResources {
   uint64_t llLastCleaning;
 };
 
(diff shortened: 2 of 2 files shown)


sjeaugey:
Just to confirm my understanding is correct...

I'm guessing on a DGX system (and most GPU systems relying on PCI switches or NVLink), admins would want to set `NCCL_SHM_DISABLE=1` to avoid silent fallback from NVLink to SHM in case there is a problem with P2P on a node. And in that precise case, we'd want to also avoid a silent fallback to NET for intra-node communication, hence this new env var `NCCL_NET_DISABLE_INTRA` which would be set to 1. With all that set, if there is an issue with P2P we would error out with `ncclSystemError` which would denote a issue with the system configuration.

Is that the rationale? Is there more to it?

kwen2501:
Hi @sjeaugey thanks for the quick response. It would be nice to support the scenario you described.

As another example, it would make it easier to detect sub-optimal subgroup mapping on a DGX1-like system, for example, GPUs 0,1,6,7.

sjeaugey:
So in the case you mention  (0,1,6,7), we would fail. But in another case (0,4,6,7) we would not fail, while actually getting the same performance. I'm not sure why you call 0,1,6,7 suboptimal... using NET in that case might actually be not that bad.

kwen2501:
Sorry for the confusion, I called the (0,1,6,7) mapping suboptimal as I was comparing it with (0,1,2,3) or (4,5,6,7) which are fully on NVLinks. And yes, it is the intent to fail in the case of (0,1,6,7).

As to (0,4,6,7), I guess it would not fail due to the help of indirect NVLink, right? If the system admin considers that acceptable, it would be okay to stay (after all, it has the benefit of not exposing traffic to the network, which may have complications in the self routing needed). If the system admin does not consider (0,4,6,7) acceptable, they would still be able to make it fail by setting `NCCL_NVB_DISABLE=1`, right?

The mapping detection is just one aspect of this env, the link failure detection you mentioned in comment 2 is another important aspect.

xw285cornell:
@sjeaugey any followup on this? We need this PR to fix some of the issues in prod :)

sjeaugey:
Sorry for the delay -- done.

PR #605, opened 2021-12-03
Improve warning message about truncated messages

kwen2501:
The "truncated message" error is most often triggered by a mismatch in collective size or env settings between ranks.

For better interpretability, the patch displays hints of cause so that it would be easier for user to debug.

It also changes the error type from InternalError to InvalidUsage to reflect the above most-likely causes.


sjeaugey:
Thanks for the PR. I can agree with the intent but shouldn't we do the same for IB, and split the tests [here](https://github.com/NVIDIA/nccl/blob/v2.12/src/transport/net_ib.cc#L989) in two parts, and return invalid usage for the first check (`size > slots[r].size`), keeping internal error for the others?

kwen2501:
Sylvain, that's a great idea, thanks! I added a commit to implement what you refer to.

PR #616, opened 2021-12-28
Add NCCL_NET_TRUST_GDR_CAP

dmaryin:
Prevent deadlock for net plugins with synchronus .connect


dmaryin:
Going to provide better solution

PR #617, opened 2021-12-30
Remove check for GDR cap by registering and deregistering memory region

dmaryin:
Issue: <tbd>


jbachan:
A more robust approach would be to change `ncclGpuGdrSupport` to spawn a new temporary thread, and perform `connect()` and `accept()` from distinct threads. No new env knob needed.

PR #623, opened 2022-01-14
Fix deadlock caused by race condition during initialization.

chr1sj0nes:
I'm not certain, but I believe the issue is caused by some threads running ahead and enqueuing lots of kernels before another thread has finished initializing. This can lead to the `cudaMalloc` call within `devCommSetup` becoming deadlocked on a mutex within the CUDA driver.

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -940,6 +940,14 @@ ncclResult_t ncclCommInitRankSync(ncclComm_t* newcomm, int nranks, ncclUniqueId 
   NCCLCHECKGOTO(commAlloc(newcomm, nranks, myrank), res, cleanup);
   NCCLCHECKGOTO(initTransportsRank(*newcomm, &commId), res, cleanup);
   NCCLCHECKGOTO(devCommSetup(*newcomm), res, cleanup);
+  
+  // Prevent deadlock in CUDA driver by ensuring that all ranks have finished
+  // allocating before allowing any to proceed and enqueue kernels.
+  int isLast;
+  NCCLCHECKGOTO(ncclCpuBarrierIn(*newcomm, &isLast), res, cleanup);
+  if (isLast) NCCLCHECKGOTO(ncclCpuBarrierLast(*newcomm), res, cleanup);
+  INFO(NCCL_INIT,"comm %p rank %d nranks %d cudaDev %d busId %lx - Init BARRIER", *newcomm, myrank, nranks, (*newcomm)->cudaDev, (*newcomm)->busId);
+  NCCLCHECKGOTO(ncclCpuBarrierOut(*newcomm), res, cleanup);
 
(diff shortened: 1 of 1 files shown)


sjeaugey:
Thanks, so this is an intra-process deadlock I guess. Did you see a hang with a stack trace showing one thread in devCommSetup while others were past ncclCommInitRank?

An alternative could be to move the intra-node barrier at the end of initTransportsRank after devCommSetup. That should work as well.

Do you have a reproducer?

chr1sj0nes:
I saw a stacktrace with one or two threads stuck at a `cudaMalloc` call, while all other devices had finished `ncclCommInitRank` and gone on to enqueue kernels. `devCommSetup` didn't appear in the stacktrace (I guess it was inlined) but I tracked it down to that malloc using trace logging.

I'm afraid I don't have a simple reproducer for this. The error first appeared recently in a fairly standard ML model running via JAX / XLA on a single host with 8 x V100s. I believe that this error wasn't seen previously as we used to call `ncclCommInitRank` from a single thread, but we now call it from parallel threads (https://github.com/tensorflow/tensorflow/commit/4098210298e4f8b1ec81171a02e2c35023f64c08).

sjeaugey:
We have added a global intra-node barrier at the [end of the init function](https://github.com/NVIDIA/nccl/blob/master/src/init.cc#L1002) since NCCL 2.10, which should cover the intra-process barrier (and more). If you're still experiencing hangs it would be good to report the hang backtrace so that we can understand where we're stuck.

hawkinsp:
I don't know whether this is related, but I added a barrier to JAX (XLA) code to work around a similar-sounding deadlock a few weeks ago: https://cs.opensource.google/tensorflow/tensorflow/+/f9df01236193494a18f05fdc0b12391552db0b00

(This won't fix the recent issues reported by @chr1sj0nes et al. because they must have that fix already.)

sjeaugey:
@chr1sj0nes I'm going to push the change below soon. This is a bit different than your patch but it does apply well to recent NCCL versions as well. Please let me know if you have any concern. Sorry for the time it took to get this through.

```
diff --git a/src/init.cc b/src/init.cc
index 25c8d5d..42c1090 100644
--- a/src/init.cc
+++ b/src/init.cc
@@ -1037,6 +1037,8 @@ collnet_cleanup:
     }
   }
 
+  NCCLCHECKGOTO(devCommSetup(comm), ret, affinity_restore);
+
   /* Local intra-node barrier */
   NCCLCHECK(bootstrapBarrier(comm->bootstrap, comm->localRankToRank, comm->localRank, comm->localRanks, comm->localRankToRank[0]));
 
@@ -1087,7 +1089,6 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
   }
   NCCLCHECKGOTO(commAlloc(newcomm, nranks, myrank), res, cleanup);
   NCCLCHECKGOTO(initTransportsRank(*newcomm, &commId), res, cleanup);
-  NCCLCHECKGOTO(devCommSetup(*newcomm), res, cleanup);
 
   // update communicator state
   comm->initState = ncclSuccess;
```

PR #625, opened 2022-01-18
Allow network plugins to provide inter node latency

rashikakheria:
This PR includes the following major changes:

1.  Fix number of inter-node hops when calculating all reduce collective time for ring algorithm.
2. Allow network plugins to provide inter-node latency for NCCL's tuning logic. This helps picking the right algorithm for different networks.

This PR has been successfully tested on AWS EC2 infrastructure with EFA networks.

### DIFF
--- a/ext-net/google-fastsocket/Makefile
+++ b/ext-net/google-fastsocket/Makefile
@@ -1,10 +1,10 @@
 CUDA_HOME?=/usr/local/cuda
 INC:=-I$(CUDA_HOME)/include
-PLUGIN_SO:=../../build/libnccl-net.so
+PLUGIN_SO:=libnccl-net.so
 
 default: $(PLUGIN_SO)
 
-$(PLUGIN_SO): nccl-fastsocket/net_fastsocket.cc nccl-fastsocket/compat.cc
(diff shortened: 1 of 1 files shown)


sjeaugey:
Thanks for the PR. I tested it a bit and I had to make a few modifications:
  - Fix the internode steps fomula so that on a single node we would still have 0 internode steps, i.e. replace `2*nNodes` by `(nNodes > 1 ? 2*nNodes :0)`. This was breaking single node tuning.
 - Use 0 as default latency instead of -1. This is simply more practical and a latency of 0 also makes no sense so it shouldn't cause confusion.
  - memset net properties to 0 so that plugins not setting the latency value won't have a random latency.
  - Make `latencyinter` optional in graph XML loading to be able to use old graph XMLs.

Please let me know if you see any issue with those changes. Thanks!

rashikakheria:
Thanks for the feedback! I will upload the new PR with the suggested changes.

sjeaugey:
No need for updates, I ported it to 2.12 and made the changes already. Just wanted to confirm the changes were ok with you.

PR #642, opened 2022-02-22
fix cuda10.0 can't build std::memcpy

CeryZhai:
fix cuda 10.0 compile error


jbachan:
What host compiler are you using? `std::memcpy` should have been brought in by the `#include <cstring>` from the top of `enqueue.cc`. Can you reproduce this:

PR #645, opened 2022-03-03
SM-efficient Tree algorithm

kwen2501:
Reduce NCCL channels (hence SM usage) by merging two complementary binary trees into the same channel.
On systems with per-NIC speed <= 100 Gbit/s, this reduces SM usage by half without lowering performance.
On systems with per-NIC speed of 200 Gbit/s, the original number of channels is used.
Also avoids tree overlapping in multi-communicator cases.

### DIFF
--- a/src/graph/tuning.cc
+++ b/src/graph/tuning.cc
@@ -234,7 +234,6 @@ ncclResult_t ncclTopoTuneModel(struct ncclComm* comm, int minCompCap, int maxCom
     comm->threadThresholds[a][NCCL_PROTO_SIMPLE] = NCCL_SIMPLE_THREAD_THRESHOLD;
   }
   comm->threadThresholds[NCCL_ALGO_RING][NCCL_PROTO_LL] *= nRanks;
-  comm->threadThresholds[NCCL_ALGO_COLLNET][NCCL_PROTO_SIMPLE] = 512;
 
   // Override defaults with user env
   char* str = getenv("NCCL_THREAD_THRESHOLDS");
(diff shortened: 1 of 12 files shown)


src/graph/tuning.cc
```diff
@@ -249,6 +248,17 @@ ncclResult_t ncclTopoTuneModel(struct ncclComm* comm, int minCompCap, int maxCom
     }
   }
 
+  // Tune channel factor based on intra-channel parallelism and graph speed
+  comm->channelFactors[NCCL_ALGO_RING] = 1;
+  // For Tree: when NIC bw is smaller than 100G (12GB/s), we can use one SM (two trees)
+  // to drive one NIC; otherwise, two SM's are needed (as if one SM only contributes
+  // one tree)
+  const float treeSpeed = nNodes <= 2 ? treeGraph->speedIntra : treeGraph->speedInter;
+  comm->channelFactors[NCCL_ALGO_TREE] = treeSpeed <= 12.0 ? NTREES : 1;
```
kumpera, line 257: Isn't this conditional flipped?

kwen2501: Hi @kumpera, thanks for your interest and review!

In line 256, we are getting the per-channel speed of tree. If there is only one node, we look at intra-node speed, hence `speedIntra`; if there are two nodes, due to tree's double-bandwidth effect in this case, mostly likely `2*speedInter` is higher than `speedIntra`, so we still look at `speedIntra`; if there are more than two nodes, we look at `speedInter`.
I should note that this line corresponds to [line 122](https://github.com/NVIDIA/nccl/blob/9c50caf95dffa43ae0d1e8d47d13459a594f6042/src/graph/tuning.cc#L122).

Line 257 means that if the tree speed is not so high (i.e. <= 12 GB/s or 100 Gb/s), we can use one SM to drive two trees, hence the channel factor is set to 2; otherwise, channel factor equals is set to 1, meaning one SM for one tree.

PR #664, opened 2022-04-05
yet another cmake port

cyyever:
I just copy the content of makefiles in CMake files. Compared with #95 this PR uses newer features of CMake and should be more easy to understand. No Windows support is planed, thus the changes are as small as possible. Of course, adding Windows supported based on this PR is more easy(All bash substitution code is implemented in CMake).
I have linked this CMake generated lib in my pytorch build and it works well. If the community don't want to merge, I think it can be left to help those guys that need Windows builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,24 @@
+cmake_minimum_required(VERSION 4.0)
+
+project(nccl LANGUAGES CUDA CXX VERSION 2.27.7)
+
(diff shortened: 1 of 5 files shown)


cyyever:
I have integrated the cmake version into my pytorch build and it works.

teojgo:
In release [v2.28.3-1](https://github.com/NVIDIA/nccl/releases/tag/v2.28.3-1) CMake support was added.

PR #678, opened 2022-05-10
Make NCCL_STEPS tunable parameter

koyongse:


koyongse:
> Making `NCCL_STEPS` vary operation to operation sounds to me rife with bugs. In the fifo's, the steps are the addressable "slots". The address of a slot given by `fifoBase + slotId*(fifoSize/NCCL_STEPS)`. Fifo's are not reset between consecutive operations. So if operation A last wrote to slot `x` of a fifo, then operation B will begin writing to slot `(x+1)%NCCL_STEPS`. If operation B were to have an increased value for `NCCL_STEPS`, then that addressing formula could easily result in aliasing the memory from operation A. Please prove to me this isn't possible.

Thanks a lot for your comment. Yes, you are totally right. I couldn't catch the FIFO is not reset between operations. Even if I can make it a valid scenario, yes, it should still be bug-prone.
 
> If I understand the intent of this PR correctly, which is to dynamically tune the granularity of fifo access, I think a better approach would be to make either the step-per-slice or slice-per-chunk quantities variable, but holding `NCCL_STEPS = MAX_STEP_PER_SLICE * MAX_SLICE_PER_CHUNK` as constant.

Yes, this is better idea for sure. I will update the PR with new code.

Thanks!

PR #691, opened 2022-06-07
Dynamic buffer depth

koyongse:
This replaces https://github.com/NVIDIA/nccl/pull/678.

This is to make NCCL buffer depth dynamic by adjusting sliceSteps and chunkSteps.


rashikakheria:
The current patch series scale up `NCCL_STEPS` for all NCCL workloads (using any network). This will affect the sizes of each chunks and slices. However, the equivalent changes to network buffer size happens only for EFA networks. Wouldn't this discrepancy performance for other networks?

roywei:
Hi, @jbachan @sjeaugey could you help take a look at this PR again? We found this will improve PyTorch FSDP model training by ~10%. It would be great if NCCL main branch can have such change. Thanks!

rahul003:
NCCL team, any update on this PR merge? Seems to have a significant performance impact.

src/transport/net.cc
```diff
@@ -793,6 +792,7 @@ static ncclResult_t sendProxyProgress(struct ncclComm* comm, struct ncclProxyArg
       int stepSize = resources->buffSizes[p] / NCCL_STEPS;
       char* localBuff = NCCL_NET_MAP_GET_POINTER(&resources->map, cpu, buffs[p]);
       int buffSize = stepSize*args->sliceSteps;
+      int maxDepth = resources->shared ? NCCL_SHARED_STEPS/args->nsubs : NCCL_STEPS;
```
sjeaugey, line 795: Is that used somewhere?

src/include/devcomm.h
```diff
@@ -56,7 +56,7 @@ union ncclLLFifoLine {
 #define NCCL_LL_FLAG_MAX   0x100
 #define NCCL_LL_FLAG(a) ((uint32_t)((a) % NCCL_LL_FLAG_MAX))
 #else
-#define NCCL_LL_CLEAN_MASK 0x7ffffff8
+#define NCCL_LL_CLEAN_MASK 0x7fffff20
```
sjeaugey, line 59: I think that should be `0x7ffffe0`

sjeaugey:
> NCCL team, any update on this PR merge? Seems to have a significant performance impact.

Sorry I had a few comments pending but did not submit them.

This is a significant change and it is not easy to convince ourselves this won't cause issues down the line. That means we need to spend a lot of time on this patch which we're struggling to find.

sjeaugey:
Sorry this is taking a long time, but as I expected, there is a lot more work than this patch.
Changing NCCL_STEPS means making LL/LL128/Tree use more than one step per slice by default and that's ingrained everywhere. Also, point-to-point communication needs to be able to use larger sizes than 128K.

To try to list all the things I'm working on:
 - Reduced the LL/LL128 buffsize which was now 4x larger (especially a problem for the LL128 which is bigger than the Simple buffer already!)
 - Fixed the LL cleanup in 2 places
 - Made the number of steps configurable with env vars
 - Allowed for send/recv sliceSteps > 1 (not complete yet -- this has lots of ramifications)
 - Fixed tuning to use chunkSize instead of stepSize in some places
 - Fixing some LL128 data corruption (not done)
 
And there will probably be more other issues down the line with tuning, buffer management, ...

bwbarrett:
We've been working internally to rebase the patch onto master, as you noted, not a trivial job.  We'd still like something similar to this patch available for our customers, and we have some developer resources available now to work on it.  What's the best path forward to avoid massive duplication of work?

koyongse:
I'm closing this PR as nvidia thankfully rebased it - https://github.com/NVIDIA/nccl/tree/inc_nsteps

PR #682, opened 2022-05-24
NCCL 2.13 Preview

sjeaugey:
Pushing this branch as a preview of NCCL 2.13. Feel free to give it a try and report issues on this PR.

@rashikakheria @changlan this is introducing the v6 NET plugin API which includes support for dmabuf. There is still a v4->v5->v6 compat layer.

PR #711, opened 2022-08-04
fix NCCL_DEBUG_FILE behavior

kingchc:
`NCCL_DEBUG_FILE` does not work properly since the recent v2.13.4 updates (https://github.com/NVIDIA/nccl/pull/682) because it now sets `ncclDebugLevel` after parsing `NCCL_DEBUG_FILE`. Hence, `NCCL_DEBUG_FILE` will never be parsed.

This patch moves parsing `NCCL_DEBUG` and setting`tempNcclDebugLevel` before parsing `NCCL_DEBUG_FILE`, and ensure `NCCL_DEBUG_FILE` is parsed only when `NCCL_DEBUG > NCCL_LOG_VERSION` (same as previous behavior, but now check `tempNcclDebugLevel` instead of `ncclDebugLevel`).

### DIFF
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -26,6 +26,20 @@ void ncclDebugInit() {
   pthread_mutex_lock(&ncclDebugLock);
   if (ncclDebugLevel != -1) { pthread_mutex_unlock(&ncclDebugLock); return; }
   const char* nccl_debug = getenv("NCCL_DEBUG");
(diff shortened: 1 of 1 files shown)


wenbilliams:
Thanks, @kingchc. Your MR was just integrated into master and will be generally available on the next release.

PR #720, opened 2022-09-13
Fix intermittent 11.6 builds: generate unique .cu file per object file

janeyx99:
This applies a patch that @peterbell10 proposed that fixed PyTorch 11.6 builds from failing intermittently: see https://github.com/pytorch/pytorch/pull/84245

### DIFF
--- a/src/collectives/device/gen_rules.sh
+++ b/src/collectives/device/gen_rules.sh
@@ -13,6 +13,9 @@ then
     datatypes+=" bf16"
 fi
 
+echo "CURDIR := \$(dir \$(realpath \$(word \$(words \$(math\$(MAKEFILE_LIST))-1), \$(MAKEFILE_LIST))))"
++echo ""
+
 targets="GENOBJS := \\\\\n"
 
 for base in sendrecv all_reduce all_gather broadcast reduce reduce_scatter; do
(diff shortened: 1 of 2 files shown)


src/collectives/device/gen_rules.sh
```diff
@@ -13,6 +13,9 @@ then
     datatypes+=" bf16"
 fi
 
+echo "CURDIR := \$(dir \$(realpath \$(word \$(words \$(math\$(MAKEFILE_LIST))-1), \$(MAKEFILE_LIST))))"
++echo ""
```
sjeaugey, line 17: This doesn't seem to work; the + is probably unwanted here.

src/collectives/device/gen_rules.sh
```diff
@@ -21,10 +24,17 @@ for base in sendrecv all_reduce all_gather broadcast reduce reduce_scatter; do
     dtn=0
     # Order must match that of the ncclDataType_t enum
     for dt in ${datatypes}; do
-      echo "${dir}/${base}_${op}_${dt}.o : ${base}.cu ${dir}/${base}.dep"
+      # Generate a unique filename for each compilation unit,
+      # otherwise the __nv_module_id may conflict at link time
+      echo "${dir}/${base}_${opn}_${dtn}.cu :"
```
sjeaugey, line 29: I get the point of using #include here, but all those .cu files are almost empty, all the code is in the .h file. Therefore, it's much simpler to just copy them: that avoids the CURDIR definition above.

src/collectives/device/gen_rules.sh
```diff
@@ -21,10 +24,17 @@ for base in sendrecv all_reduce all_gather broadcast reduce reduce_scatter; do
...
+      echo "	echo \"#include \\\"\$(CURDIR)${base}.cu\\\"\" > \$@"
+      echo ""
+      # Compile the file
+      echo "${dir}/${base}_${op}_${dt}.o : ${dir}/${base}_${opn}_${dtn}.cu ${base}.cu ${dir}/${base}.dep"
```
sjeaugey, line 33: I'm not sure why we're using opn/dtn here. For readability purposes I would use op/dt to be consistent with the rest of the file naming.

sjeaugey:
Would this patch be okay?

https://github.com/NVIDIA/nccl/commit/3348a8387975549a6aa513fde21dce85766c6795

janeyx99:
Thanks @sjeaugey for the thoughtful review. I've reflected my PR to basically be your commit as it should do the job/makes sense to me.

PR #724, opened 2022-09-22
Return result outside CHECK macro in ibvwrap.cc

minsii:
Summary: This patch refactors all CHECK macros in ibvwrap.cc to NO_RETURN version which only sets ncclResult_t without return inside the macro. It allows potential profiling points to be inserted after the ibv calls.


src/misc/ibvwrap.cc
```diff
@@ -219,47 +228,69 @@ const char *wrap_ibv_get_device_name(struct ibv_device *device) {
...
-  IBV_PTR_CHECK_ERRNO(ibv_internal_alloc_pd, ibv_internal_alloc_pd(context), *ret, NULL, "ibv_alloc_pd");
+  ncclResult_t res;
+  IBV_PTR_CHECK_ERRNO_NO_RETURN(ibv_internal_alloc_pd, ibv_internal_alloc_pd(context), *ret, NULL, "ibv_alloc_pd", res);
+  return res;
 }
 
 ncclResult_t wrap_ibv_dealloc_pd(struct ibv_pd *pd) { /*returns 0 on success, or the value of errno on failure (which indicates the failure reason)*/
-  IBV_INT_CHECK_RET_ERRNO(ibv_internal_dealloc_pd, ibv_internal_dealloc_pd(pd), 0, "ibv_dealloc_pd");
+  ncclResult_t res;
+  IBV_INT_CHECK_RET_ERRNO_NO_RETURN(ibv_internal_dealloc_pd, ibv_internal_dealloc_pd(pd), 0, "ibv_dealloc_pd", res);
```
sjeaugey, line 286: Not sure we need that extra line?

src/misc/ibvwrap.cc
```diff
@@ -140,64 +140,71 @@ ncclResult_t wrap_ibv_symbols(void) {
   return initResult;
 }
 
-#define IBV_PTR_CHECK_ERRNO(name_internal, call, retval, error_retval, name) \
+#define IBV_PTR_CHECK_ERRNO_NO_RETURN(name_internal, call, retval, error_retval, name, res) \
```
sjeaugey, line 143: Actually if we move ALL calls to those "no return" macros then maybe we don't need to rename them, just change their behavior?

So I'd even remove the _NO_RETURN everywhere.

minsii: Makes sense. I modified the patch to remove _NO_RETURN. Can you please take a look? Thank you.

PR #747, opened 2022-11-09
Fix `suggest braces around initialization of subobject` from clang

r-barnes:
Fixes a clang warning:
```
include/nvtx3.hpp:1669:9: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        0,                              // payload value (union)
        ^
        {}
include/nvtx3.hpp:1671:9: warning: suggest braces around initialization of subobject [-Wmissing-braces]
        0                               // message value (union)
        ^
        {}
```


kiskra-nvidia:
Thank you for your submission, and sorry for the long wait!

This issue has been fixed upstream in the NVTX header files; the updated version should be included in the next NCCL release.

PR #758, opened 2022-12-14
unlock mutex on failure

cyyever:
Some failure cases forget to unlock the mutex.


kiskra-nvidia:
The code has changed since, but on closer examination we still found a couple of cases that were missing. This should be fixed in the next NCCL release. Thank you!

PR #772, opened 2023-01-18
Fix maximum handle size for NCCL Net v4 API

rashikakheria:
NCCL Net v4 supports a maximum handle size of 64 bytes whereas the ext-net example header files set it for NCCL Net v3. Since, `aws-ofi-nccl` plugin plans to follow the example header files, fix it here.

Signed-off-by: Rashika Kheria <rashika@amazon.com>


sjeaugey:
Indeed, my mistake, the handle size was increased on v5.

Issue #750, opened 2022-11-15
NCCL Hang with CUDA_LAUNCH_BLOCKING=1


xw285cornell:
It seems to be a regression in v2.13 (2.10 should work) that when I run a simple MLP model (with all reduce) will hang when CUDA_LAUNCH_BLOCKING is present. The hang happens on all ranks with 100% GPU spinning on something. The backstack from gdb is this. It looks this ncclStrongStreamLaunchKernel is new in v2.13, so wondering if this is a new regression

#0  0x00007fec9b630c50 in ?? () from lib/libcuda.so.1
#1  0x00007fec9b49dcf7 in ?? () from lib/libcuda.so.1
#2  0x00007fec9b3f8b9f in ?? () from lib/libcuda.so.1
#3  0x00007fec9b3fa1b8 in ?? () from lib/libcuda.so.1
#4  0x00007fec9b3e301f in ?? () from /lib/libcuda.so.1
#5  0x00007fec9b636830 in ?? () from lib/libcuda.so.1
#6  0x00007fec9b370bd6 in ?? () from lib/libcuda.so.1
#7  0x00007fec9b372bc2 in ?? () from libcuda.so.1
#8  0x00007fec9b40d4d5 in ?? () from lib/libcuda.so.1
#9  0x00007fec9ae1405c in ?? () from /lib/libcudart.so.11.4.108
#10 0x00007fec9ae67f06 in cudaLaunchKernel () from /lib/libcudart.so.11.4.108
#11 0x00007fecdf7d6cf9 in ncclStrongStreamLaunchKernel(ncclCudaGraph, ncclStrongStream*, void*, dim3, dim3, void**, unsigned long) ()
#12 0x00007fecdf7a56c8 in ncclLaunchKernel(ncclComm*, ncclKernelPlan*) () 
#13 0x00007fecdf7c494d in ncclGroupEndInternal() () 
#14 0x00007fecdf7c5301 in ncclGroupEnd () from 
#15 0x00007fecdf794f4c in torch::cuda::nccl::AutoNcclGroup::~AutoNcclGroup() () 
#16 0x00007fecdf725655 in c10d::ProcessGroupNCCL::allgather(std::vector<std::vector<at::Tensor, std::allocator<at::Tensor> >, std::allocator<std::vector<at::Tensor, std::allocator<at::Tensor> > > >&, std::vector<at::Tensor, std::allocator<at::Tensor> >&, c10d::AllgatherOptions const&) ()
[...]

PR #774, opened 2023-01-20
Partial fix CUDA_LAUNCH_BLOCKING=1

jbachan:
Resolve some but not all deadlock when running with CUDA_LAUNCH_BLOCKING=1.

The two known sources of deadlock:

1) When the proxy thread is used (e.g. network peers): NCCL launches the kernel before enqueueing work to the proxy thread since kernel launches are higher latency. But, CUDA_LAUNCH_BLOCKING=1 causes that kernel launch to block until the kernel completes, which the kernel can't do because the proxy thread is blocked waiting for the work.

2) When managing multiple devices per OS thread: As NCCL has moved away from the deprecated cudaLaunchCooperativeKernelMultiDevice, we now loop through the devices owned by a thread and launch the kernels in turn. Since the first launch will block on its kernel which is blocked waiting for all the launches which communicate together, we deadlock.

This commit addresses only case 1. We simply make sure proxy work is enqueued first when CUDA_LAUNCH_BLOCKING=1.

### DIFF
--- a/src/enqueue.cc
+++ b/src/enqueue.cc
@@ -15,6 +15,19 @@
 #include <cstring> // std::memcpy
 #include <cinttypes> // PRIx64
 
+static pthread_once_t initOnceControl = PTHREAD_ONCE_INIT;
+static bool isCudaLaunchBlocking = false;
+
+static void initOnceFunc() {
+  char* val = getenv("CUDA_LAUNCH_BLOCKING");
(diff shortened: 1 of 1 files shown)


jbachan:
Addresses issue #750

teojgo:
Closing since this has already been merged.

PR #779, opened 2023-02-02
Print remote and local GID in error trace

YulunW:
The remote and local GID information would be very useful in finding the root cause of a network issue. It could help us quickly pinpoint the two faulty GIDs without inspecting all the possible paths. Therefore, we would like to include this information in the error trace to speed up error tracing.

### DIFF
--- a/src/transport/net_ib.cc
+++ b/src/transport/net_ib.cc
@@ -380,6 +380,10 @@ struct ncclIbRequest {
   int type;
   int events;
   struct ncclSocket* sock;
+  // Only one of sendComm and recvComm can be not NULL.
+  // Both can be NULL for gpuFlush.
+  struct ncclIbSendComm* sendComm;
+  struct ncclIbRecvComm* recvComm;
   int nreqs;
   union {
     struct {
@@ -431,6 +435,10 @@ struct ncclIbSendComm {
   int nqps;
   struct ibv_mr* fifoMr;
   int ar;
+
+  int devId;  // Index in ncclIbDevs
+  union ibv_gid localGid;
+  union ibv_gid remoteGid;
 };
 // The SendFifo needs to be 32-byte aligned and each element needs
 // to be a 32-byte multiple, so that an entry does not get split and
@@ -464,6 +472,8 @@ struct ncclIbRecvComm {
   struct ibv_qp* qps[NCCL_IB_MAX_QPS];
   int nqps;
   struct ncclIbGpuFlush gpuFlush;
+  union ibv_gid localGid;
+  union ibv_gid remoteGid;
 };
 static_assert((offsetof(struct ncclIbRecvComm, remFifo) % 32) == 0, "ncclIbSendComm fifo must be 32-byte aligned");
 
@@ -609,6 +619,7 @@ ib_connect_check:
   if (!ready) return ncclSuccess;
 
   // IB Setup
+  comm->devId = dev;
   struct ibv_context* ctx;
   ctx = ncclIbDevs[dev].context;
   NCCLCHECK(ncclIbInitVerbs(dev, ctx, &comm->verbs));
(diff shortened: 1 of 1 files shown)


sjeaugey:
Thanks for the request. It makes sense. I'm not sure I'd have implemented it that way, but this would indeed help debugging. We're pretty busy at the moment so not sure we'll have time to work on that soon, but please ping us again if we don't make any progress.

src/transport/net_ib.cc
```diff
@@ -1329,4 +1362,3 @@ ncclNet_t ncclNetIb = {
   ncclIbCloseRecv,
   ncclIbCloseListen
 };
-
```
vtlam, line 1332: remove this deletion (irrelevant and unnecessary)

src/transport/net_ib.cc
```diff
@@ -380,6 +380,10 @@ struct ncclIbRequest {
   int type;
   int events;
   struct ncclSocket* sock;
+  // Only one of sendComm and recvComm can be not NULL.
+  // Both can be NULL for gpuFlush.
+  struct ncclIbSendComm* sendComm;
+  struct ncclIbRecvComm* recvComm;
```
vtlam, line 386: Add brief comments to elaborate the meaning and purpose of these variables

src/transport/net_ib.cc
```diff
@@ -431,6 +435,10 @@ struct ncclIbSendComm {
   int nqps;
   struct ibv_mr* fifoMr;
   int ar;
+
+  int devId;  // Index in ncclIbDevs
+  union ibv_gid localGid;
+  union ibv_gid remoteGid;
```
vtlam, line 441: Add brief comments on the variables

src/transport/net_ib.cc
```diff
@@ -464,6 +472,8 @@ struct ncclIbRecvComm {
   struct ibv_qp* qps[NCCL_IB_MAX_QPS];
   int nqps;
   struct ncclIbGpuFlush gpuFlush;
+  union ibv_gid localGid;
+  union ibv_gid remoteGid;
```
vtlam, line 476: Add brief comments on the variables

src/transport/net_ib.cc
```diff
@@ -1237,11 +1258,23 @@ ncclResult_t ncclIbTest(void* request, int* done, int* sizes) {
...
         char line[SOCKET_NAME_MAXLEN+1];
+        char localGid[INET6_ADDRSTRLEN];
+        char remoteGid[INET6_ADDRSTRLEN];
         union ncclSocketAddress addr;
         ncclSocketGetAddr(r->sock, &addr);
-        WARN("NET/IB : Got completion from peer %s with error %d, opcode %d, len %d, vendor err %d",
-             ncclSocketToString(&addr, line), wc->status, wc->opcode, wc->byte_len, wc->vendor_err);
+        if (r->sendComm != NULL) {
+          inet_ntop(AF_INET6, &(r->sendComm->localGid), localGid, sizeof(localGid));
+          inet_ntop(AF_INET6, &(r->sendComm->remoteGid), remoteGid, sizeof(remoteGid));
```
vtlam, line 1269: These inet_ntop() calls may fail by returning NULL (see https://man7.org/linux/man-pages/man3/inet_ntop.3.html). For completeness, I think we should check inet_ntop() failures, in which case we can either fall back to the old WARN message or just print "unknown" for whichever GID that failed?

Also I think you can refactor this error handle out to a helper method  if you feel it is getting complex.

src/transport/net_ib.cc
YulunW: This seems to be a styling preference for our linter... Thanks for pointing this out! Reverted the change.

src/transport/net_ib.cc
YulunW: Changed the logic to handle inet_ntop failures.

AddyLaddy:
There are compilation warnings from your patch. I believe the WARN messages are not being passed the correct string parameters:
```
transport/net_ib.cc:1289:16: warning: format ‘%s’ expects argument of type ‘char*’, but argument 12 has type ‘ibv_gid*’ [-Wformat=]
 1289 |           WARN("NET/IB : Got completion from peer %s with error %d, opcode %d, len %d, vendor err %d (%s) localGid %s remoteGid %s",
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1290 |               ncclSocketToString(&addr, line), wc->status, wc->opcode, wc->byte_len, wc->vendor_err,commType, localGid, remoteGid);
      |                                                                                                               ~~~~~~~~
      |                                                                                                               |
      |                                                                                                               ibv_gid*
```

YulunW:
> 

Ah thanks for the heads up! Yes I am supposed to use the string version. Just changed it.

PR #789, opened 2023-02-22
Fix char8_t typedef

nzmsv:
Prior to this change the `#if` blocks were inside another `#if` branch that limited them to `#ifdef __cplusplus`, rendering the C part of the feature test non-functional.  There was also a stray line that looked like it came from the definition of `nvtxExtPayloadTypeInfo`.  In short, it appears there has been a mis-merge.

Second, the standard versions used in the test were not correct.
- C: `char8_t` will be introduced in C23 
  Source: https://gustedt.gitlabpages.inria.fr/c23-library/#char8_t-added-mandatory-change
  `char16_t` and `char32_t` were introduced in C11 (existing test was correct)
- C++: `char8_t `was introduced in C++20
  `char16_t` and `char32_t` were introduced in C++11
  Source: https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros

### DIFF
--- a/src/include/nvtx3/nvtxExtDetail/nvtxExtPayloadTypeInfo.h
+++ b/src/include/nvtx3/nvtxExtDetail/nvtxExtPayloadTypeInfo.h
@@ -54,11 +54,14 @@ MKTYPEDEF(size_t);
 MKTYPEDEF(pointer_type);
 
 MKTYPEDEF(wchar_t);
-#if (__STDC_VERSION__ > 201710L) || (defined(__cplusplus) && __cplusplus > 201703L)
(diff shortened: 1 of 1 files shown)


nzmsv:
This still looks wrong in the new NCCL release with the NVTX update.

sjeaugey:
Indeed this particular issue was not resolved in 2.17 which was frozen before you posted this issue. It will be fixed in 2.18.

nouiz:
Is this fixed? 2.18.1 is out: https://docs.nvidia.com/deeplearning/nccl/release-notes/rel_2-18-1.html#rel_2-18-1

nzmsv:
Looks like this is fixed.  Thank you!

PR #791, opened 2023-02-23
Fix memory leak.

chsigg:


KaimingOuyang:
Hi chsigg,
We fix the issue slightly differently. Here is the patch, please see if it works for you. Thanks!
[0001-Free-proxyState.listenSock-before-proxy-thread-exits.patch](https://github.com/NVIDIA/nccl/files/10827491/0001-Free-proxyState.listenSock-before-proxy-thread-exits.patch)

chsigg:
That works, thanks a lot. Can I recommend that you check for memory leaks before releasing a new version? This is not the first time that memory is leaked even for basic operations.

KaimingOuyang:
Thank you Chris!
Sorry for the inconvenience.
Yes, we will check the gpu and cpu memory leak carefully next time.

PR #800, opened 2023-03-13
Support building NCCL with RDMA-core [part 1 of 2]: refactoring ibvwrap.{h, cc}

vtlam:
Currently NCCL dynamically loads IB verbs at runtime by dlopen() on libibverbs.so shared objects in ibvwraps.cc.
However, there are use cases for explicit linkage (note: this can be either static or shared) with rdma-core at build time:

- Streamlining rdma-core version in job build config and bundling rdma-core in job binary package. This can help enforcing all relevant RDMA applications (e.g., trainers and monitoring daemons) to use the same rdma-core version at build time (e.g., regardless of multi-tenancy and host platforms).
- Avoiding hard-coding shared object names, which can cause issues in case certain names are changed between rdma-core versions (e.g., libmlx5-rdmav25.so in v35 vs libmlx5-rdmav34.so in v32).
- Simplifying rdma-core testing before production rollout: avoiding extra out-of-band rolling back and forth system rdma-core rpm packages on test machines. Supporting testing multiple rdma-core versions concurrently on the same set of machines. 

For that purpose, this change refactors ibvwrap.{h, cc} into 3 modules:
- ibvcore.h: verbatim first part of the old ibvwrap.h that contains duplicate IB verbs structs (e.g., ibv_qp, ibv_cq, ...) being copied from verbs.h. This header should be mutually exclusive of <infinband/verbs.h>.
- ibvsymbols.{h,cc}: constructs IB verbs symbols, i.e., either dynamically loading with dlopen() or just pass-through pointers to native IB verbs included from <infinband/verbs.h>
- ibvwrap.{h, cc}: focusing only on NCCL's wrapper methods wrap_ibv_*()

This is split into two stacked pull requests:
- Part 1 (this PR): no-op refactoring of ibvwrap.{h, cc} to facilitate native rdma-core linkage.
- Part 2 (PR [[802](https://github.com/NVIDIA/nccl/pull/802/commits/18fabf7e57217b4ef821f79b53bc096eb72b3250)]): 
defines the macro NCCL_BUILD_RDMA_CORE (i.e., specifying with -DNCCL_BUILD_RDMA_CORE in Makefile) that will chose the explicit linkage mode. Otherwise if this macro is not defined, the NCCL code structure should be exactly the same as before this change.


sjeaugey:
For context, the reason we never did it ourselves is because:
 - This code is part of the NCCL core, which is already tied to a particular CUDA version so we didn't want it to be bound to a particular OFED version as well.
 - This code is part of the NCCL core, so we wanted to make it easy for anyone to build NCCL (including with IB support) without having to find the right includes and go through a `configure` step.
 - Symbols in libibverbs are versioned and we can therefore reliably use the version which we support, as well as detect the IB verbs version and whether a symbol is present or not.
 - When we want to build against a particular version of OFED, we use the [external plugin](https://github.com/Mellanox/nccl-rdma-sharp-plugins) which is designed to be built against a specific network stack to test it.

So overall, while I have no objection in principle (the code was designed originally to permit this), I doubt we'll be using it ourselves, meaning that we might break it if we make any change to IB. So my concern here is only that we would need to test this somehow if we don't want to risk breaking it, which has a cost for us (maintenance and testing) -- when you might be able to achieve the same effect using the external plugin.

What do you think?

vtlam:
hi Sylvain, thanks for the quick reply.

So there is an inherent dependency of NCCL on rdma-core and it is just the matter of where we resolve that dependency. Deferring the resolution until runtime can simplify build time and as you mentioned NCCL can adapt based on detected rdma-core version (something not under your control in open-source setting).

However from our perspective as production operators, production reliability is our important priority and we would like to proactively detect issues the sooner the better in the software release workflow. We also have full ownership of end-to-end software stack, so ideally we would like to be deterministic about dependencies and resolve conflicts at compile time as much as we can. For reliability, we'd like to ship to production only binaries that have been fully tested with their known dependencies.

Waiting until runtime can be too late at scale, since any runtime issue will typically guarantee to cause the jobs to degrade or crash, and undermine our SLA at scale.

I have refactored this PR into two parts: (1) the first part is only refactoring of ibvwrap.{h,cc} into smaller modules and (2) the second part is the plumbing of native <infiniband/verbs.h>. See the bottom of updated description in this PR.

I think (1) should be immediately ready for merging since it's essentially no-op and should not have any impact.
Looking at NCCL network plugin, my impression is that net_ib.cc is a big chunk and maintaining it as a plugin seems too much with code duplication if the purpose is only pointing to native ib verbs (e.g., comparing to Part (2)).

We already have to maintain our own internal testing for NCCL correctness & performance anyway (i.e., regardless of rdma-core linkage mode). So with this change on your side, I expect you can just add an extra test that NCCL can compile with NCCL_BUILD_RDMA_CORE.

Let me know if that makes sense. Thanks!

sjeaugey:
As I was checking the patch, I hit a problem. The code will not compile if the network stack is too old and does not have an IB verb used by NCCL. In my case, I tried to build on a somewhat old environment which did not have `ibv_reg_dmabuf_mr` and it failed.

It sounds to me that's not something you would want in a production environment, as it would prevent you from upgrading to the latest NCCL unless you upgrade your network stack to the latest, which could be tied to the distribution and therefore require an OS upgrade. That's not something you'd want to do just to fix a bug in NCCL, and might even not be possible at all if NCCL uses new IB verbs not yet released in standard distros.

One solution could be to add more defines to define the IB verbs version, but that's just more complexity for everyone.

Regarding the rationale, I'm still not fully convinced. If the libibverbs on the runtime is not the same as the libibverbs you used to compile NCCL, it will still run even when compiled directly against libibverbs (if the library used is more recent) so, making sure the network stack used is the same that the one you tested with should be solved at another level, like the container.

sjeaugey:
This has been added with 2.19.3. Closing.

PR #801, opened 2023-03-14
NCCL performance tuner plugin: tuner interface and default implementation

vtlam:
Currently NCCL performance tuner is implemented as a built-in NCCL component in graph/tuning.cc.
We'd like to explore possible custom tuners separately for various scenarios (e.g., network topologies, hardware platforms) that may exist simultaneously in production.

This PR proposes an interface so that NCCL can support external performance tuner plugin, fully configurable at run time. Note that the existing tuner implementation is still the default tuner when there is no external plugin being specified.
- NCCL_TUNER_PLUGIN: the environment variable to specify plugin shared object
- nccl_performance_tuner.h: defines the tuner interface 
- init.cc: initializes and destroys the tuner plugin
- enqueue.cc: calls the plugin's method to decide the collective algorithm. Falls back the the default topo-based tuning algorithm if necessary.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -12,7 +12,7 @@ INCEXPORTS  := nccl.h nccl_net.h
 LIBSRCFILES := init.cc init_nvtx.cc channel.cc bootstrap.cc transport.cc enqueue.cc group.cc debug.cc proxy.cc net.cc \
 		misc/cudawrap.cc misc/nvmlwrap.cc misc/ibvwrap.cc misc/gdrwrap.cc \
 		misc/utils.cc misc/argcheck.cc misc/socket.cc misc/shmutils.cc misc/profiler.cc misc/param.cc misc/strongstream.cc \
-		misc/ipcsocket.cc \
+		misc/ipcsocket.cc misc/performance_tuner.cc \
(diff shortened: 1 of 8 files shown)


sjeaugey:
I'm not sure I would be okay with such a plugin. Plugins should be external objects, and therefore they should be decoupled from the NCCL core. Here the code is loading an external object which is completely coupled with the NCCL includes (`info`, but also `info->comm` and everything it pulls).

So if we want to make this a plugin, we need to first define what the API will be, then version it, including all relevant structs in the include. That might be complicated. The `ncclInfo` struct is the main interface between the plugin and NCCL core. We'd need to extract everything we might want to use in that struct (including info->comm->nRanks for example) into two separate, versioned structs:
  - One for input data (what we provide to the plugin to make its decision): nranks, nnodes, ngpus, size, ... maybe chunk size as well, even though it's not there today.
  - One for output data: protocol, algorithm, number of CTAs and number of threads (although for those last two, I'm not sure).

vtlam:
hi Sylvain,

Thanks for the feedback. I went through a couple of iterations internally and made changes accordingly. The API is now a lot simpler and also with minimal changes to NCCL-core. Please take another look.

sjeaugey:
The tuner plugin is part of NCCL 2.19.3. Closing.

PR #806, opened 2023-03-17
Flip order of ncclSetThreadContext and cudaSetDevice to use the correct context

### LINKED ISSUE #803: NCCL_SHM_USE_CUDA_MEMCPY=1 causes hang in PyTorch
Using cuda memcpy is much faster, so I'm trying to make it work. On a machine with two non nvlinked 3090s.

```
# boilerplate NCCL init code

nbytes = 65536*2
buf0 = torch.randn(nbytes // 4).cuda(rank)
dist.all_reduce(buf0)
#torch.cuda.synchronize(rank)
buf0.cpu()
```

If I comment out the torch.cuda.synchronize, it works, otherwise it hangs. I suspect it has to do with the stream created in `shmSendProxyConnect` (also, does that stream work with cuda graphs?)

If this is something you'd like to look into, I can help with a more full reproduction.

geohot:
Fix #803

### DIFF
--- a/src/proxy.cc
+++ b/src/proxy.cc
@@ -817,10 +817,10 @@ NCCL_PARAM(ProgressAppendOpFreq, "PROGRESS_APPENDOP_FREQ", 8);
 
 void* ncclProxyProgress(void *comm_) {
   struct ncclComm* comm = (struct ncclComm*)comm_;
-  if (ncclSetThreadContext(comm) != ncclSuccess) {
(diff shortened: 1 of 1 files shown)


stephenmsachs:
It seem this issue has been addressed in https://github.com/NVIDIA/nccl/commit/d97a32fac84f69cbd022ffdc57d5687c760742c7. Thanks for your help. I will go ahead and close this PR. Please open a new one or add a comment if you disagree.

PR #825, opened 2023-04-17
ncclDebugLog: fix for log payload larger than buf

danielahlin:
The logging function ncclDebugLog does not properly check the return value of the vsnprintf call (or, less importantly, the snprintf calls). Consequenly, when the length of the logging header and the logging message payload is larger than buf, then : 

1. Log messages are incorrect.
2. buffer[len++] = '\n' writes out of bounds.
3. fwrite(buffer, 1, len, ncclDebugFile) reads out of bounds.

### DIFF
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -180,13 +180,25 @@ void ncclDebugLog(ncclDebugLogLevel level, unsigned long flags, const char *file
                    hostname, pid, tid, cudaDev, timestamp, filefunc, line);
   }
 
-  if (len) {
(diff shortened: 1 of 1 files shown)


danielahlin:
I found it when developing https://github.com/NVIDIA/nccl/pull/826.

I don't remember right now what I printed to the log, but probably the value of some environment variable.

nzmsv:
I also ran into this and wrote a patch before noticing this ticket (#1215)

xiaofanl-nvidia:
This has been fixed (along with #1215) and will be released in the next GA.

PR #826, opened 2023-04-17
Add NCCL_NET_PLUGIN_PATH to specify net plugin

danielahlin:
Add NCCL_NET_PLUGIN_PATH to allow directly specifying the so-file providing the plugin. The immediate purpose is to avoid having to interfere with LD_LIBRARY_PATH in order to determine which  NCCL plugin to use. This is useful in e.g. spack (see e.g. https://github.com/spack/spack/pull/36557#discussion_r1154339775 for background).

### DIFF
--- a/src/net.cc
+++ b/src/net.cc
@@ -166,14 +166,31 @@ enum ncclNetState ncclNetStates[3] = { ncclNetStateInit, ncclNetStateInit, ncclN
 enum ncclNetState ncclCollNetStates[3] = { ncclNetStateInit, ncclNetStateInit, ncclNetStateInit };
 
 ncclResult_t ncclNetPluginInit() {
-  char ncclNetPluginName[128];
+  char ncclNetPluginName[PATH_MAX];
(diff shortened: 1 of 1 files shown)


sjeaugey:
In theory the plugin should be installed in a path which is managed by ld.so.conf, so LD_LIBRARY_PATH should not need to be set.
Now if you want to use a plugin which is not installed on the system (typical development workflow, or even packaging in some cases), I'm not sure why that would be a problem.

danielahlin:
Indeed, this setting is for situations where you don't want to rely on the system-wide ld.so.conf to comprehensively describe all places where you may have libraries. IMO this is a common scenario, examples include environments like spack and environment modules but probably also python virtual environments, conda, easy-build etc...

One way to adress this is setting an LD_LIBRARY_PATH but that is a global setting which may be troublesome if many different packages are trying to set/append values to it. A more precise one is to specifically point out the library you wish as is suggested in this patch.

PR #848, opened 2023-05-20
fix NCCL_DEBUG initialization

kingchc:
Since `ncclDebugInit` only happens once at the first appearance of `ncclDebugLog` (i.e., when calling TRACE, INFO...), it is possible NCCL attempts to call debug log before setting environment variables from a config file, which happens in `ncclInit` now. This results in discrepancy between rank-0 and other ranks, e.g., rank-0 prints log but other ranks won't, because rank-0 performs `ncclInit` during `ncclGetUniqueId` while other ranks would wait until actual initialization (e.g., `ncclCudaLibraryInit` in https://github.com/NVIDIA/nccl/blob/master/src/init.cc#L1578 may call INFO before reading config file`).

This patch adds extra logic to call `initEnvs` in `ncclDebugLog` and make sure `initEnvs` only performed once. Alternatively, we could also move `ncclInit` before any other calls, but I'm not sure if it has any side effects.

### DIFF
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -6,6 +6,7 @@
 
 #include "core.h"
 #include "nccl_net.h"
+#include "param.h"
 #include <stdlib.h>
 #include <stdarg.h>
 #include <sys/syscall.h>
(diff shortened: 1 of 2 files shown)


sjeaugey:
Shouldn't we just make sure initEnv is called before anything else? The only problem I can see is when using `NCCL_COMM_ID`, as the bootstrap code seems to be executed before `ncclInit` in `ncclCommInitRank`.

Do you see that issue happening under different conditions?

I'm not against the current patch, just trying to make sure I fully understand the problem.

kingchc:
@sjeaugey - any suggestion to fix this properly?

PR #884, opened 2023-06-13
do NOT add job to queue when it encounters errors

yangxingwu:


sjeaugey:
Adding @jbachan to review

PR #916, opened 2023-07-12
Fix undefined behavior in ncclKernel()

Artem-B:
We never set src/dst in the default case and, when compiled with clang, it allows compiler to consider the default branch unreachable (better than launching rockets, I guess) and that leads to a miscompilation with recent clang builds.

### DIFF
--- a/src/collectives/device/common.h
+++ b/src/collectives/device/common.h
@@ -166,7 +166,8 @@ __device__ void ncclKernel(
       bytes = 0;
       break;
     }
-    copyToShmem16(tid%WARP_SIZE, dst, src, bytes);
+    if (bytes)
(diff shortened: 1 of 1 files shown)


sjeaugey:
Thanks, merged for next release. Do you need us to merge it to master now?

Artem-B:
That's up to NCCL owners. I'm just a messenger. :-)

In general, considering that tensorflow had switched to clang, it may make sense to get it into the main branch earlier. They do plan to update clang version soon and the new clang version will expose the bug.

Artem-B:
Clang is about to branch-off v17 and start release candidate builds within the next couple of weeks. TF will likely grab them shortly after that as they need other clang changes in v17 to support Ada/Hopper GPUs.

Issue #950, opened 2023-08-07
Segmentation fault in the proxy profiler


dfyz:
Running any sufficiently large collective with `#define PROFILE_PROXY 1` and `#define ENABLE_TIMER 1` added to [the profiler](https://github.com/NVIDIA/nccl/blob/v2.18.3-1/src/misc/profiler.cc#L9) results in a segmentation fault when the proxy starts trying to track network communications. The fault looks roughly like this:
```
Caught signal 11 (Segmentation fault: address not mapped to object at address 0x8)
...
1 0x000000000008f157 ncclProfilingRecord()  /home/dfyz/nccl/src/misc/profiler.cc:57
...
```

More precisely, it's a null-pointer dereference of `event` [here](https://github.com/NVIDIA/nccl/blob/v2.18.3-1/src/misc/profiler.cc#L56). I think this is what happens:
  * no more than `NCCL_STEPS` steps should be in-flight at any given time, so the profiler code assumes that it is safe to use `profilingEvents[step%NCCL_STEPS]` to access events for step `step`
  * however, in the very beginning of both `sendProxyProgress()` ([here](https://github.com/NVIDIA/nccl/blob/v2.18.3-1/src/transport/net.cc#L917)) and `recvProxyProgress()` ([here](https://github.com/NVIDIA/nccl/blob/v2.18.3-1/src/transport/net.cc#L1067)) we access all steps at once to save the timestamp for `ncclProxyProfileBegin`
  * but this is not safe, because:
    * (for example) for step `8` (` = NCCL_STEPS`), we will overwrite the timestamp for step `0` instead of using a unique slot (which is mostly harmless on its own)
    * the `ncclProxyProfileEnd` event for step `0` will [clear](https://github.com/NVIDIA/nccl/blob/v2.18.3-1/src/misc/profiler.cc#L52) `profilingEvents[0]`, which now contains data for step `8` (by itself it is, again, mostly harmless)
    * however, the `ncclProxyProfileBegin` event for step `8` will never arrive again, so when we try to use `profilingEvents[0]` for step `8`, we do it without proper initialization, and therefore get a null-pointer dereference

I might have completely misunderstood the intent behind the original code, but at least increasing the number of slots in `profilingEvents` (so that no slots are ever shared between the steps), and then changing `profilingEvents[step%NCCL_STEPS]` to `profilingEvents[step%NUM_SLOTS]` appears to work around the problem. The resulting traces look legitimate, although I can't say with certainty that they are correct.

[...]

PR #960, opened 2023-08-14
Fix segmentation fault in the proxy profiler

dfyz:
A draft PR fixing issue #950.

`ncclProfilingRecord(..., ncclProxyProfileBegin)` now just sets a newly introduced `profileBeginTimestamp` for each proxy sub, instead of trying to set the `profilingEvents[ncclProxyProfileBegin]` timestamp for all steps of the sub at once (see the issue for why it is not feasible).

Only later, when either `ncclProxyProfileSendGPUWait` or `ncclProxyProfileRecvWait` event arrives, `ncclProfilingRecord()` actually allocates the event for the current step and copies `profileBeginTimestamp` from the proxy sub to the newly created event.

Events that do not correspond to a step (i.e., anything with `state >= 8`) are treated as before.

Only tested with straightforward collective operations that use a single sub. However, as far as I can see, proxy subs are handled independently of each other, so I think the patch should work for multiple subs, too.


liralon:
Please provide a commit message body for your commit. Explain explicitly what was the issue that was fixed and what is the commit's change that fixes it.

dfyz:
@liralon @sjeaugey I added a commit body with an explanation of the issue and a description of how I tried to fix this. See also [this comment](https://github.com/NVIDIA/nccl/issues/950#issuecomment-1679278393) for some additional details.

dfyz:
I rebased my fix over the latest master. There were no meaningful changes, I just had to resolve a trivial merge conflict, since some new fields were added to `ncclProxySubArgs` recently.

PR #972, opened 2023-08-22
Use bitcast analogue instead of C-style cast due to alignment issue.

PatriosTheGreat:
Hi everyone,
I'm working on a Google OpenXLA project.
During the upgrade to NCCL 2.18.3 I faced a SIGILL exception which is getting thrown during ncclBootstrapHandle->magic field addressing if using the last Clang upstream version.
The problem seems to be related to ncclCommInitRankAsyncJob fields alignment. It looks like the commId field might be not aligned as uint64_t which leads to SIGILL after casting it to ncclBootstrapHandle and addressing the "magic" field.

I see two ways how to fix this issue:
1. Use bit_cast analogue instead of C-style cast (this what I'm trying to do in the current PR).
2. Use explicit uint64 alignment for the commId field at the ncclCommInitRankAsyncJob structure.

### DIFF
--- a/src/graph/xml.cc
+++ b/src/graph/xml.cc
@@ -13,6 +13,9 @@
 #include "core.h"
 #include "nvmlwrap.h"
 #include "xml.h"
+#if defined(__x86_64__)
+#include <cpuid.h>
+#endif
 
 /*******************/
 /* XML File Parser */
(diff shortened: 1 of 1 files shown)


liralon:
I believe this is the issue that I caught with UBSAN (See https://github.com/NVIDIA/nccl/pull/956) and fixed in commit https://github.com/NVIDIA/nccl/pull/956/commits/7bf20ba266c33a0aece679053b5a9d715a09fa4e . The PR was not yet merged. (@sjeaugey FYI)

PR #979, opened 2023-08-31
Allow statically link with network plugin.

PatriosTheGreat:
We want to have ability to link network plugin and NCCL statically on the same binary.
This can be achieved by passing null to dlopen, which will cause it to return a handle to the main program.
For that we added a compile time definition NCCL_NET_PLUGIN_STATIC_LINKING which would identify that network plugin needs to be loaded from the current binary.


gcongiu:
Hi @PatriosTheGreat thank you for this patch. Support for statically linked network plugins will be integrated into NCCL v2.22.

PatriosTheGreat:
Many thanks!

PR #1012, opened 2023-09-27
Fix select net code in case when a node has more GPUs than NICs.

PatriosTheGreat:
I faced with ASAN failure in this part of code on the line:
if (localNetCount > 0 && localNets[localNetCount] == localNets[0]) break;

IIUC the problem here is in addressing localNets array outside of the boundary.
I assume this is happening in cases when nodes have more GPUs than NICs.


sjeaugey:
This has been resolved in 2.19 by replacing:
```
   NCCLCHECK(ncclCalloc(&localNets, system->nodes[NET].count));
```
by:
```
  NCCLCHECK(ncclCalloc(&localNets, MAXCHANNELS));
```

PatriosTheGreat:
Hi Sylvain, many thanks for pointing to the proper fix!
I tried that change and it fixed our failure. Closing this PR.

PR #1064, opened 2023-11-12
fix: do not keep retrying `accept` if handshake fails for "proxy" type sockets

0x804d8000:
In `socketFinalizeAccept`, if the handshake (magic validation) fails, we should return failure for "proxy" sockets.

When `ncclProxyService` tries accepting new connection, with the current behavior, a spurious connection can cause `ncclSocketAccept` endlessly burns CPU cycles. As `ncclProxyService` marks the listening socket non-blocking, once a spurious connection cause `socketFinalizeAccept` to fail, we immediately retry `ncclTryAccept`, and get `EAGAIN`, and retry `ncclTryAccept`...

Unfortunately we can't just change the default behavior though. Several other callsites rely on `ncclSocketAccept` to ignore spurious connections for them, e.g., callsites in `bootstrap.c`. IMHO, those callsites should be refactored and either 1) do retries themselves, or 2) use something new such as `ncclSocketAcceptBlockUntilHandshakeSucceeds`.


0x804d8000:
Hi, @sjeaugey would you mind taking a look at it?

PR #1067, opened 2023-11-13
Enforce Autoformatting

wesbland:
Enforce the auto-formatting scripts that already exist in the NCCL repository.


wesbland:
This repo seems to have GitHub Actions turned off. Obviously, before merging this PR, that should be turned on to enforce that no changes would land without checking the formatting first.

wesbland:
I took a look at creating a GitLab version of the GitHub Action commit, but I think it might be too specific to NVIDIA's internal systems and I'm not sure I'd be able to replicate it.

I hope the GitHub Action example that I've created is obvious enough that it should be easy to migrate internally before merging.

sjeaugey:
Could you comment on the INDENT OFF / INDENT ON exceptions we had to add? Is there some bad formatting which we'd want to keep for visual reasons?

wesbland:
> Could you comment on the INDENT OFF / INDENT ON exceptions we had to add? Is there some bad formatting which we'd want to keep for visual reasons?

There were a few exceptions I made in https://github.com/NVIDIA/nccl/pull/1067/commits/4296f86f09daecece8f1d121f98aed07f5118252 for visual reasons, yes. I think all of them were in headers where there was a nicely formatted struct + comments that ended up looking weird with the rules from the auto-formatter.

For example, this:

```cpp
typedef struct {
  char* name;     // Used mostly for logging.
  char* pciPath;  // Path to the PCI device in /sys.
  uint64_t guid;  // Unique identifier for the NIC chip. Important for
                  // cards with multiple PCI functions (Physical or virtual).
  int ptrSupport; // [NCCL_PTR_HOST|NCCL_PTR_CUDA|NCCL_PTR_DMABUF]
  int speed;      // Port speed in Mbps.
  int port;       // Port number.
  float latency;  // Network latency
  int maxComms;   // Maximum number of comms we can create
  int maxRecvs;   // Maximum number of grouped receives.
}ncclNetProperties_v6_t;
```

becomes this:

```cpp
typedef struct {
  char* name; // Used mostly for logging.
  char* pciPath; // Path to the PCI device in /sys.
  uint64_t guid; // Unique identifier for the NIC chip. Important for
  // cards with multiple PCI functions (Physical or virtual).
  int ptrSupport; // [NCCL_PTR_HOST|NCCL_PTR_CUDA|NCCL_PTR_DMABUF]
  int speed; // Port speed in Mbps.
  int port; // Port number.
  float latency; // Network latency
  int maxComms; // Maximum number of comms we can create
  int maxRecvs; // Maximum number of grouped receives.
}ncclNetProperties_v6_t;
```

(or something close to it).

wesbland:
Also, I obviously didn't run this script on the repo for this PR to make it easier to read. That can either be added here or in another PR if the rest looks good. It's trivial to run `make format`.

sjeaugey:
Thanks a lot, so it's just the auto-formatter doesn't like aligned comments.

I think we should either find a flag to ignore those, or have some exclude list, so that we can skip formatting on those files (I'm thinking about the src/include/nvtx3 directory for example, as well as exported includes).

wesbland:
For anyone reviewing this, I'd highly recommend,

1. Review commit-by-commit instead of looking at the entire PR. It's way too noisy.
2. Turn off whitespace-only changes to see the meaningful changes. Mostly, it's just moving `{`'s on their own lines to the end of the previous line, but that can be changed if that's what people prefer.
Screenshot 2024-01-16 at 11 54 03 AM

PR #1070, opened 2023-11-14
Fix use of CPUID overwriting registers in use

Flamefire:
I debugged a segmentation fault happening with GCC 12 and high optimizations. The cause is that CPUID writes to EAX, EBX, ECX, and EDX so the inline-asm must state that. Otherwise currently in-use register might get overwritten which may cause all kinds of failures like segfaults or wrong results.

Hence `asm volatile("cpuid" : "=a" (cpuid1.val) : "a" (1) : "memory");` is wrong as it doesn't state that and at least needs to be replaced by `asm volatile("cpuid" : "=a" (cpuid1.val) : "a" (1) : "ebx", "ecx", "edx", "memory");`

Similar also for `asm volatile("cpuid" : "=b" (cpuid0.ebx), "=c" (cpuid0.ecx), "=d" (cpuid0.edx) : "a" (0) : "memory");` which doesn't mention the clobbering of the EAX register.

Alternatively `__cpuid` can be used which avoids this and related issues. So do that as suggested in the GCC issue https://gcc.gnu.org/bugzilla/show_bug.cgi?id=112513   
See that also for stacktraces of the observed crash and debug-logs showing a modified parameter leading to this issue

See also https://faydoc.tripod.com/cpu/cpuid.htm for documentation of `cpuid` and the modified registers and https://wiki.osdev.org/CPUID for a wiki entry about `__cpuid` (especially with the "unused" parameter)

### DIFF
--- a/src/graph/xml.cc
+++ b/src/graph/xml.cc
@@ -12,6 +12,9 @@
 #include "core.h"
 #include "nvmlwrap.h"
 #include "xml.h"
+#if defined(__x86_64__)
+#include <cpuid.h>
+#endif
 
 /*******************/
 /* XML File Parser */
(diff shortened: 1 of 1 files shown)


AddyLaddy:
Great thanks for tracking that down and providing the fix. We'll try and get that merged in to github and our master code base very shortly.

PR #1081, opened 2023-11-20
fix protoEnable array when setting LL128

jiangxiaobin96:
```
int protoEnable[NCCL_NUM_PROTOCOLS] = { 1, 2, 1 };
```
When using LL128, `protoEnable[1]` is set by value 2 as initialization. However, if setting `NCCL_PROTO=LL128`, the value `protoEnable[1]` will be 1 after `parseList`, because parseList can only set 0 or 1 which will invalidate the following code in Line 287.
```
if (pEnable == 2 && p == NCCL_PROTO_LL128) {
```

### DIFF
--- a/src/graph/tuning.cc
+++ b/src/graph/tuning.cc
@@ -255,6 +255,7 @@ ncclResult_t ncclTopoTuneModel(struct ncclComm* comm, int minCompCap, int maxCom
   if (protoStr) {
     INFO(NCCL_ENV, "NCCL_PROTO set by environment to %s", protoStr);
     NCCLCHECK(parseList(protoStr, ncclProtoStr, NCCL_NUM_PROTOCOLS, protoEnable));
+    if (protoEnable[1]) protoEnable[1] = 2;
   }
   const char *algoStr = ncclGetEnv("NCCL_ALGO");
   if (algoStr) {


sjeaugey:
This is working as expected. If you set `NCCL_PROTO=LL,LL128,SIMPLE` then you are forcing NCCL to use LL128. By default LL128 is enabled automatically. Changing that may degrade performance or cause data corruption.

Users should not set NCCL_PROTO except for workarounds (e.g. when we want to disable a protocol until a bug is fixed).

jiangxiaobin96:
Sorry, I have not understood.
If set `NCCL_PROTO=LL,LL128,SIMPLE` then will force NCCL to use LL128, and now `protoEnable[LL128] = 1` because parseList can only set 0 or 1. I think this will skip the following `if` because `pEnable!=2`.By default LL128 is enabled automatically and now `protoEnable[LL128] = 2` which will enter the following `if`.
```
int pEnable = protoEnable[p];
if (pEnable == 2 && p == NCCL_PROTO_LL128) {}
```

sjeaugey:
Basically we want `NCCL_PROTO=LL128` to force the use of LL128. If the user says I want to use LL128, we want to honor that, even if that means potential data corruption.

jiangxiaobin96:
Does that mean we don't suggest user force to set `NCCL_PROTO=LL128`? Suggest to set by default or not include LL128?

This PR wants to fix the bug when user force to set `NCCL_PROTO=LL128,LL,...` include LL128.
Logically, `NCCL_PROTO=LL128,LL,Simple`(protoEnable[]={1,1,1}) should be equivalent to default(protoEnable[]={1,2,1})
For example, if user sets `NCCL_PROTO=LL128,LL,Simple`, `protoEnable[]={1,1,1}` which is not equivalent to default and now `pEnable =protoEnable[1]=1` will skip if judgement.

I think this is a bug, but if users should not set NCCL_PROTO, it can be overlook.

PR #1135, opened 2024-01-04
Add support for allocating persistent plans using async alloc

ezhulenev:
Currently when NCCL operations captured in CUDA graphs NCCL allocates an on-device storage for pointer to plan using sync allocations, which adds significant overheads to graph capture.

Add an option to set up a stream for stream ordered allocator to allocate persistent state on device.


ezhulenev:
We do graph capture upfront and reuse them when possible, but we can't guarantee pointer stability so some time we have to re-capture and update graphs. Without this change enabling graph capture for NCCL leads to ~30% regression in large LLM model. With this change it gives ~5-10% speedup.

Issue #1155, opened 2024-01-22
Pack bootstrapAllGather's


kwen2501:
https://github.com/NVIDIA/nccl/blob/b6d7438d3145a619f924dbbca6c96db21fab716e/src/init.cc#L1315-L1316

Any chance of packing these two all-gathers into one?

sjeaugey:
I don't think that would be a problem indeed.

kwen2501:
Closed via NCCL 2.23.

PR #1157, opened 2024-01-22
Pack bootstrapAllGathers in ncclCommSplit

kwen2501:
Implements #1155


stephenmsachs:
This seems to be fixed in [v2.23](https://github.com/NVIDIA/nccl/issues/1155#issuecomment-2383621530).

PR #1183, opened 2024-02-19
Add wc_store_fence support for RISC-V

Xeonacid:


Xeonacid:
> Friendly check-in: this PR hasn't had activity in 486 days. I plan to revisit it next cycle. If it's no longer needed, let me know and I'll close; otherwise happy to incorporate any remaining feedback.

It's still needed. Thanks for the review!

AddyLaddy:
We have some changes in progress that will modify the `__aarch64__` code for `wc_store_fence()` so we'd need a specific `__riscv` elif e.g. 

```
#elif defined(__riscv)
static inline void wc_store_fence(void) { asm volatile("fence ow,ow" : : : "memory"); }
```

Xeonacid:
> We have some changes in progress that will modify the `__aarch64__` code for `wc_store_fence()` so we'd need a specific `__riscv` elif e.g.
> 
> ```
> #elif defined(__riscv)
> static inline void wc_store_fence(void) { asm volatile("fence ow,ow" : : : "memory"); }
> ```

Thanks for taking this. I'd appreciate if you guys could commit it along with the aarch64 changes!

AddyLaddy:
I think we'd a few more changes than just that to support RISC-V.
So I think a specific RISC-V PR should be created.

Xeonacid:
> I think we'd a few more changes than just that to support RISC-V. So I think a specific RISC-V PR should be created.

Agree! I'll be waiting for the aarch64 changes.

PR #1194, opened 2024-02-26
Add nccl parameter to override traffic class for fifo messages

ashshetty-prog:
Summary: This change adds the support to override the traffic class attribute of the qp on which the cts/fifo messages are sent on. We want to prioritize fifo message to be sent on a network queue different from data traffic. Doing so will remove control path messages from causing slow downs in the case of network congestion when loaded latency is higher. This will only be enabled when NCCL_IB_FIFO_TC variable is set otherwise its disabled by default.


sjeaugey:
Merged in 2.22.3.

PR #1217, opened 2024-03-11
[bugfix]save undo plans in some case

crazy-JiangDongHua:
In the original logic, some comm‘s plans will be missed in some cases. 

For example,  suppose there is a comm list here, containing two comms. The first comm contains three plans, and the second comm contains one plan. In the first round of the loop, the first plan of the two comms will start the corresponding kernel. In the second round of the loop, the second plan of the first comm will start the corresponding kernel, and then the second comm has no undo plan, so moreround is set to false, and finally the loop exits. The final result is that the third plan of the first comm has not been executed, the second plan has not yet finish stream sync, and the collective communication is over.

Although this possibility is relatively small, because the work budget of a plan is so large.

### DIFF
--- a/src/group.cc
+++ b/src/group.cc
@@ -142,7 +142,7 @@ static ncclResult_t doLaunches(struct ncclComm* head) {
     }
 
     while (true) { // Iterate rounds of launches for clique.
-      bool moreRounds;
+      bool moreRounds = false;
       comm = cliqueHead;
       do { // Iterate clique members.
         struct ncclComm* next = comm->groupNext;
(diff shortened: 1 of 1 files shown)


crazy-JiangDongHua:
By the way, I also want to ask about clique related issues. I don’t understand under what circumstances the list groupCommHeadMain will contain comms from two processes.

jbachan:
Thanks this looks good. We will get back to you on whether it will be accepted as is or bundled into the next release.

As to your question, a process can call ncclCommInitRank twice with different unique id's making it a member of two different global communicators. Both of those ncclComm_t's could later reside within the same ncclGroupStart/End.

crazy-JiangDongHua:
Thanks for your reply, I understand the problem. I have never written like this style before, I will try it in the future.

PR #1226, opened 2024-03-14
fix  the calculation of comm's maxCompCap & minCompCap

zmxdream:
hi, i found the calculation of comm maxCompCap & minCompCap maybe wrong , can anyone please check it


sjeaugey:
Yes it was wrong; it is fixed already (in next 2.21 release).

zmxdream:
ok

PR #1228, opened 2024-03-14
fix clang errors

kingchc:
Fixes two errors, specifically when compiling with clang-15

1. `-Wmissing-braces`: add missing brackets around object construction in`nctx3.hpp`. Upstream PR of nvtx from https://github.com/NVIDIA/NVTX/pull/88

2. `-Wshift-sign-overflow`
reproducer: `CXX=clang++ CC=clang CXXFLAGS="-Werror -Wshift-sign-overflow" make src.build CUDA_HOME=/usr/local/cuda`
error:
```
net.cc:304:16: error: signed shift result (0x80000000) sets the sign bit of the shift expression's type ('int') and becomes negative [-Werror,-Wshift-sign-overflow]
  if (size >= 1<<31) return ncclInternalError;
```

### DIFF
--- a/src/include/nvtx3/nvtx3.hpp
+++ b/src/include/nvtx3/nvtx3.hpp
@@ -1937,9 +1937,9 @@ class event_attributes {
         0,                              // color value
         NVTX_PAYLOAD_UNKNOWN,           // payload type
         0,                              // reserved 4B
-        0,                              // payload value (union)
+        {0},                              // payload value (union)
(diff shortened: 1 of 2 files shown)


kiskra-nvidia:

Your fixes to `net.cc` should be included in the next NCCL release.

The issue with `nvtx3.hpp` has been fixed upstream in the NVTX header files; the updated version should be included in the next release as well.

PR #1256, opened 2024-04-17
[bugfix] fix multimem reduce result override

Thunderbrook:


sjeaugey:
@jbachan please review and merge. Thanks!

Thunderbrook:
@jbachan plz help to review and merge

Thunderbrook:
@sjeaugey plz help to review

sjeaugey:
This looks like a copy/past issue indeed, but we don't currently have the case where MultimemSrcs is more than one. So we could as well remove the multimem branch in that code path; that code is never used.

So there is no urgency in fixing it.

kiskra-nvidia:
Thank you for your contribution! We expect it to be included in the next NCCL release.

stephenmsachs:
Closing the PR as this change has been merged into v2.23. Thanks again for fixing!

PR #1265, opened 2024-04-25
Bugfix：Avoid deadlock issues caused by logging in a loop

wangfakang:
To prevent deadlock issues caused by logging in a loop, so cache the value before the log operation.
Here are the related function call paths：
```
ncclDebugLog -> ncclLoadParam(lock) -> ncclDebugLog -> ncclLoadParam(deadlock)
```

https://github.com/NVIDIA/nccl/blob/ab2b89c4c339bd7f816fbc114a4b05d386b66290/src/misc/param.cc#L62-L76

The following stack is happend local development.
image

### DIFF
--- a/src/misc/param.cc
+++ b/src/misc/param.cc
@@ -61,10 +61,6 @@ void initEnv() {
 
 void ncclLoadParam(char const* env, int64_t deftVal, int64_t uninitialized, int64_t* cache) {
   static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
-  if (__builtin_expect(__atomic_load_n(cache, __ATOMIC_RELAXED) != uninitialized, true)) {
-    return;
-  }
-
   pthread_mutex_lock(&mutex);
   if (__atomic_load_n(cache, __ATOMIC_RELAXED) == uninitialized) {
     const char* str = ncclGetEnv(env);
--- a/src/misc/param.cc
+++ b/src/misc/param.cc
@@ -61,21 +61,29 @@ void initEnv() {
 
 void ncclLoadParam(char const* env, int64_t deftVal, int64_t uninitialized, int64_t* cache) {
   static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+  if (__builtin_expect(__atomic_load_n(cache, __ATOMIC_RELAXED) != uninitialized, true)) {
+    return;
+  }
+
   pthread_mutex_lock(&mutex);
   if (__atomic_load_n(cache, __ATOMIC_RELAXED) == uninitialized) {
     const char* str = ncclGetEnv(env);
     int64_t value = deftVal;
-    if (str && strlen(str) > 0) {
-      errno = 0;
-      value = strtoll(str, nullptr, 0);
-      if (errno) {
-        value = deftVal;
-        INFO(NCCL_ALL,"Invalid value %s for %s, using default %lld.", str, env, (long long)deftVal);
-      } else {
-        INFO(NCCL_ENV,"%s set by environment to %lld.", env, (long long)value);
-      }
+    if (!str || strlen(str) <= 0) {
+      __atomic_store_n(cache, value, __ATOMIC_RELAXED);
+      pthread_mutex_unlock(&mutex);
+      return;
     }
-    __atomic_store_n(cache, value, __ATOMIC_RELAXED);
+    errno = 0;
+    value = strtoll(str, nullptr, 0);
+    // To prevent deadlock issues caused by logging in a loop,
+    // so cache the value before the log operation.
+    __atomic_store_n(cache, errno ? deftVal : value, __ATOMIC_RELAXED);
+    if (errno) {
+      INFO(NCCL_ALL,"Invalid value %s for %s, using default %lld.", str, env, (long long)deftVal);
+    } else {
+      INFO(NCCL_ENV,"%s set by environment to %lld.", env, (long long)value);
+    }
   }
   pthread_mutex_unlock(&mutex);
 }


wangfakang:
friendly ping @AddyLaddy @sjeaugey .

src/misc/param.cc
```diff
@@ -61,21 +61,29 @@ void initEnv() {
 
 void ncclLoadParam(char const* env, int64_t deftVal, int64_t uninitialized, int64_t* cache) {
   static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+  if (__builtin_expect(__atomic_load_n(cache, __ATOMIC_RELAXED) != uninitialized, true)) {
+    return;
+  }
+
```
kiskra-nvidia, line 67: This shouldn't be necessary. You are basically duplicating what's already done by the caller of this function, which is: https://github.com/NVIDIA/nccl/blob/ab2b89c4c339bd7f816fbc114a4b05d386b66290/src/include/param.h#L24-L26

kiskra-nvidia (commented):

I left one technical comment for you, but overall, I'm skeptical whether this is the best approach to solving your problem, although it's difficult to tell for sure since you didn't include the actual changes that motivate this PR (which I'm guessing is adding timestamps to the log lines?).

The problem is that adding a dependency in the debug code on the auto-generated `ncclParamEnableLogTime` function has created a circular lock dependency, leading to a deadlock. Your PR avoids the deadlock by moving the updating of the cache ahead of the debug output. It works, but it's fragile.

Instead, I recommend that you add the check for the presence of `NCCL_ENABLE_LOG_TIME` to the `ncclDebugInit` function, alongside the existing code over there that checks for `NCCL_DEBUG` and `NCCL_DEBUG_SUBSYS`. That should be more robust and thus more maintainable.

Also, I'm wondering whether `NCCL_DEBUG_TIMESTAMPS` wouldn't be a better name for this new config, but that's just bikeshedding...

wangfakang:
> Thank you for your submission!
> 
> I left one technical comment for you, but overall, I'm skeptical whether this is the best approach to solving your problem, although it's difficult to tell for sure since you didn't include the actual changes that motivate this PR (which I'm guessing is adding timestamps to the log lines?).
> 

Thank you very much for your prompt reply. I added a switch through `xcclParamEnableLogTime` in `ncclDebugLog` to determine whether to enable the log timestamp function. This change triggered the `ncclLoadParam` deadlock issue.

> The problem is that adding a dependency in the debug code on the auto-generated `ncclParamEnableLogTime` function has created a circular lock dependency, leading to a deadlock. Your PR avoids the deadlock by moving the updating of the cache ahead of the debug output. It works, but it's fragile.
> 

Yes, I am in `ncclDebugLog` adds a timestamp of switch, the use of the similar `ncclParamWarnSetDebugInfo`. So the fix mentioned by PR is to avoid this deadlock problem by caching the values before logs.

https://github.com/NVIDIA/nccl/blob/ab2b89c4c339bd7f816fbc114a4b05d386b66290/src/debug.cc#L178-L181

> Also, I'm wondering whether `NCCL_DEBUG_TIMESTAMPS` wouldn't be a better name for this new config, but that's just bikeshedding...

that's a good tips.

src/misc/param.cc
wangfakang: > This shouldn't be necessary. You are basically duplicating what's already done by the caller of this function, which is:
> 
> https://github.com/NVIDIA/nccl/blob/ab2b89c4c339bd7f816fbc114a4b05d386b66290/src/include/param.h#L24-L26

It seems so. However, as an atomic API, wouldn't it be better to have this logic inside the API?

kiskra-nvidia: Maintainability-wise, I agree with you that it would make sense to hide it inside the function. However, NCCL has these calls sprinkled all over the codebase, including in performance-critical places. Looking at the code history, I can see that the code was deliberately structured this way to minimize the runtime impact.

kiskra-nvidia:
> I added a switch through `xcclParamEnableLogTime` in `ncclDebugLog` to determine whether to enable the log timestamp function. This change triggered the `ncclLoadParam` deadlock issue.

That's what I guessed. Are you planning to submit that functionality in a future PR?

> BTW, if we moving the location of `ncclParamWarnSetDebugInfo` to inside the `NCCL_LOG_INFO` branch or printing `WARN` logs inside `ncclLoadParam` would also trigger this deadlock issue.

Yes, I see what you mean. It currently works, but it's fragile. I will propose internally a change to make it more robust, i.e., implementing custom querying code in `ncclDebugInit`. The same approach remains my recommendation for you as well (it should then be included as part of a future PR, should you plan to submit it).

src/misc/param.cc
wangfakang: > Maintainability-wise, I agree with you that it would make sense to hide it inside the function. However, NCCL has these calls sprinkled all over the codebase, including in performance-critical places. Looking at the code history, I can see that the code was deliberately structured this way to minimize the runtime impact.

Yes, I see what you mean, and addressed your comment.

wangfakang:
> > I added a switch through `xcclParamEnableLogTime` in `ncclDebugLog` to determine whether to enable the log timestamp function. This change triggered the `ncclLoadParam` deadlock issue.
> 
> That's what I guessed. Are you planning to submit that functionality in a future PR?
> 

Ok, I can submit this feature separately to another PR.

> > BTW, if we moving the location of `ncclParamWarnSetDebugInfo` to inside the `NCCL_LOG_INFO` branch or printing `WARN` logs inside `ncclLoadParam` would also trigger this deadlock issue.
> 
> Yes, I see what you mean. It currently works, but it's fragile. I will propose internally a change to make it more robust, i.e., implementing custom querying code in `ncclDebugInit`. The same approach remains my recommendation for you as well (it should then be included as part of a future PR, should you plan to submit it).

Yeah. IMO, this lock is to prevent only one thread from reading the environment variable when missing cache, which aims to improve performance rather than concurrent safety. In the scenario where there is already a cache, the overhead of the lock itself is questionable in terms of performance improvement. Therefore, I think the lock at this point can actually be removed.
Of course, I may not have considered the whole, please correct me, thank you.

dmvevents:
Friendly check-in: this PR hasn't had activity in 486 days. I plan to revisit it next cycle. If it's no longer needed, let me know and I'll close; otherwise happy to incorporate any remaining feedback.

PR #1308, opened 2024-05-31
link against rdma-core libs when build with RDMA-CORE

changchengx:


changchengx:
https://github.com/NVIDIA/nccl/pull/802

PR #1325, opened 2024-06-25
fix stack smash

madeleineth:
- In `dumpLine`, in very large comms, the format string " %3d" can be longer than four characters, so it can overflow the `line` and smash the stack. This changes `dumpLine` to not depend on knowing how big formatted strings are by using `snprintf` everywhere.
- Add an explicit error message when `ncclCommInitAll` is passed an invalid device.

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -1735,7 +1735,8 @@ ncclResult_t ncclCommInitAll(ncclComm_t* comms, int ndev, const int* devlist) {
     for (int i = 0; i < ndev; ++i) {
       /* invalid device check. */
       if (devlist[i] < 0 || devlist[i] >= totalnDev) {
-        ret = ncclUnhandledCudaError;
+        WARN("Invalid device %d (totalnDev=%d)", devlist[i], totalnDev);
+        ret = ncclInvalidArgument;
         goto fail;
       }
 


src/init.cc
```diff
@@ -1735,6 +1735,7 @@ ncclResult_t ncclCommInitAll(ncclComm_t* comms, int ndev, const int* devlist) {
     for (int i = 0; i < ndev; ++i) {
       /* invalid device check. */
       if (devlist[i] < 0 || devlist[i] >= totalnDev) {
+        WARN("Invalid device %d (totalnDev=%d)", devlist[i], totalnDev);
```
sjeaugey, line 1738: Not sure how this is related to the commit. 

It's indeed a valid issue (and we should return ncclInvalidArgument, not ncclUnhandledCudaError btw). I guess nobody uses ncclCommInitAll any more which is why nobody noticed so far.

sjeaugey:
Thanks for pointing this out. Indeed, that code was written in 2016 for NCCL 2.0 and wasn't changed since then. :man_facepalming: Crazy that it didn't cause trouble before.

I'd separate ncclCommInitAll change into another commit to avoid mixing things together that don't really relate.

src/graph/rings.cc
```diff
@@ -6,17 +6,13 @@
...
-  char line[STRLENGTH+1];
-  line[STRLENGTH] = '\0';
-  memset(line, ' ', STRLENGTH);
-  strncpy(line, prefix, PREFIXLEN);
-  for (int i=0; i<nranks && i<MAXWIDTH; i++) sprintf(line+prefixlen+4*i, " %3d", values[i]);
-  INFO(NCCL_INIT,"%s", line);
+  char line[128];
+  int n = snprintf(line, sizeof(line), "%s", prefix);
+  for (int i = 0; i < nranks && sizeof(line) > n; i++) {
+    n += snprintf(line + n, sizeof(line) - n, " %d", values[i]);
```
sjeaugey, line 13: Sorry I forgot the most important comment I wanted to make: we used to print with `%3d` so that numbers would be aligned between rings (at least with <1000 ranks) which was quite helpful for visual inspection, to check how rings were created. Was there any issue with `%3d`?

PR #1331, opened 2024-06-27
Set the default for the AMD-x86 using the PATH_PXB p2p communication.

AlextangQ19:
Set the default for the AMD-x86 using the PATH_PXB p2p communication level.
This is to avoid using GPU p2p communication across the AMD-x86 SoC root complex, which could negatively impact the performance of all2all communication.


AddyLaddy:
Thanks that's been on our to do list for multiple releases.

AlextangQ19:
> Thanks that's been on our to do list for multiple releases.

Great. May I ask if there is an approximate release time with this p2p level fix?

stephenmsachs:
It seems this has been fixed in v2.23. I am going to close this issue now. Please let me know if you disagree.

PR #1335, opened 2024-07-01
Include the CUDA call that failed in errors.

tfogal:
To make debugging easier.

### DIFF
--- a/src/include/cudawrap.h
+++ b/src/include/cudawrap.h
@@ -30,7 +30,7 @@ extern CUmemAllocationHandleType ncclCuMemHandleType;
     if( err != CUDA_SUCCESS ) {				      \
       const char *errStr;				      \
       (void) pfn_cuGetErrorString(err, &errStr);	      \
-      WARN("Cuda failure %d '%s'", err, errStr);	      \
+      WARN("Cuda failure %d '%s' from %s", err, errStr, #cmd);	      \
       return ncclUnhandledCudaError;			      \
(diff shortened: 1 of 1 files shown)


kiskra-nvidia:
Thank you for your contribution!

While on some level more information never hurts, I'm not immediately convinced that this actually makes debugging easier?

`WARN` messages already include the source file and line number; mapping that info back to a CUDA should be a no-brainer. One could argue that `INFO` does not do that, but we don't seem to be using `CUCHECKIGNORE` anywhere in our current codebase.

So could you provide an example of when your PR makes a difference?

tfogal:
We talked offline; this is slightly off the mark.

Follow-up on me to check behavior of GetLastError when the env var is not set and potentially rework this towards the goal of making GetLastError behave differently in that case. Draft mode until we're there.

tfogal:
Thanks for checking in, this slipped my mind.
I'm no longer actively working on distributed so you can close. If you wanted to take it up, a couple things I'd love:

1. reconsider the policy on not printing anything: fatal errors should come out to stderr regardless of env vars.
2. include the sub-error (cuda error, here) in `GetLastError`. here the message was `unhandled cuda error` which is not useful; if the error included the actual cuda error (e.g. `cudaErrorInsufficientDriver` or whatever) then it would be useful for the user/admin debugging things.

xiaofanl-nvidia:
Thanks! Closing this for now.

PR #1383, opened 2024-07-31
Fix the issue of printing data overflow.

wangfakang:
Fix the issue of printing data overflow.

uint32_t		byte_len; 
uint64_t		wr_id; 

https://github.com/NVIDIA/nccl/blob/178b6b759074597777ce13438efb0e0ba625e429/src/include/ibvcore.h#L302-L316

For example, in the following example, the len variable becomes a negative number when the size of the variable is out of bounds.
```
xxxx:8036:8538 [0] transport/net_ib.cc:1295 NCCL WARN NET/IB : Got completion from peer x.x.x.x <16275> with error 12, opcode 32749, len -2147483648, vendor err 129 (Send) localGid xxxxx remoteGid xxxxx
```

### DIFF
--- a/src/transport/net_ib.cc
+++ b/src/transport/net_ib.cc
@@ -1981,7 +1981,7 @@ ncclResult_t ncclIbTest(void* request, int* done, int* sizes) {
 
             char line[SOCKET_NAME_MAXLEN+1];
             char *hcaName = r->devBases[i]->pd->context->device->name;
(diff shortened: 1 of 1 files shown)


wangfakang:
friendly ping @AddyLaddy @sjeaugey

stephenmsachs:
This change has been added in v2.27. Thanks a lot for fixing!

Issue #890, opened 2023-06-16
NCCL_IB_HCA suport array, why NCCL_IB_GID_INDEX not support array


firemiles:
we have use nccl in kubernetes + rdma-device-plugin.
there are multiple HCAs in Pod like `NCCL_IB_HCA=mlx5_1:0,mlx5_2:0`。and we also want to use `NCCL_IB_GID_INDEX=3` for mlx5_1:0 and `NCCL_IB_GID_INDEX=6` for mlx5_2:0。what should I do for this usage.

sjeaugey:
That's not something NCCL supports indeed. It's the first time we get such a request. Shouldn't be too hard to implement, but still requires some work.

firemiles:
> That's not something NCCL supports indeed. It's the first time we get such a request. Shouldn't be too hard to implement, but still requires some work.

Thanks for your reply, we are using macvlan to virtualize the rdma network card in the container scenario. When a container is inserted into multiple macvlans, we encounter the problem of inconsistent nccl IB GID, which limits the way our rdma network card is used.

I don't know if there is a plan to support this approach

sjeaugey:
Just to make sure I understand the problem, in your container your NIC has two IP addresses (which map to the same physical NIC) and each interface has its own GID index. So inside the container you have both GID Index 3 and GID Index 6, which both work, but map to different VLANs, and you want to select GID 6 because you need to communicate with another GPU using that second VLAN.

Is that right?

firemiles:
> Just to make sure I understand the problem, in your container your NIC has two IP addresses (which map to the same physical NIC) and each interface has its own GID index. So inside the container you have both GID Index 3 and GID Index 6, which both work, but map to different VLANs, and you want to select GID 6 because you need to communicate with another GPU using that second VLAN.
> 
> Is that right?
Not that.
A node has 4 RDMA network cards and 8 GPUs, and each RDMA network card allows the creation of several macvlan to solve the problem of insufficient RDMA. When Pod1 adds 4 GPUs, we try to add two RDMA macvlan eth1 and eth2 to it. The corresponding masters are mlx5_1 and mlx5_2, and the GIDs are 3 and 6 respectively. I want Pod1 to use eth1 and eth2 for GDR to increase the bandwidth when running NCCL.

sjeaugey:
> creation of several macvlan to solve the problem of insufficient RDMA.

What is the problem of "insufficient RDMA"?

[...]

PR #1427, opened 2024-09-05
net_ib: return ncclSuccess if read roceTypePath failed and errno is E…

limu713:
…INVAL.  #890

Please refer to the problem details：https://github.com/NVIDIA/nccl/issues/890

### DIFF
--- a/src/transport/net_ib.cc
+++ b/src/transport/net_ib.cc
@@ -248,6 +248,10 @@ static ncclResult_t ncclIbRoceGetVersionNum(const char* deviceName, int portNum,
   close(fd);
 
   if (ret == -1) {
+    WARN("NET/IB: read failed: %s, roceTypePath: %s", strerror(errno), roceTypePath);
+    if (errno == EINVAL) {
+      return ncclSuccess;
+    }
     return ncclSystemError;
   }
 


sjeaugey:
I don't see how this is related to issue 890. Did you reference the wrong bug ID?

limu713:
> I don't see how this is related to issue 890. Did you reference the wrong bug ID?

actually, it is the same question.
"we are using macvlan to virtualize the rdma network card in the container scenario. When a container is inserted into multiple macvlans, we encounter the problem of inconsistent nccl IB GID, which limits the way our rdma network card is used."
One rdma network card has multipe sub interfaces and each sub interface has different gid index. 
for example, 
container A has the first sub interface with gid index 7, and container B has the second sub interface with gid index 11,
then you can not run miprun with  specified parameter of NCCL_IB_GID_INDEX.
issue 890 wants NCCL_IB_GID_INDEX to support array, but with this patch, you can ignore NCCL_IB_GID_INDEX. 
It will automatically select the correct gid index.

sjeaugey:
NCCL already has the needed code to automatically select the correct GID index.

Your patch is not adding that functionality, it's somehow changing the behavior in case of an error, but it's not clear why that error case would happen and why you'd want to do that. So, at best it's a workaround for a problem; and we'd want to justify why we need that workaround rather than fixing the error itself.

teojgo:
I think this can be closed since this is handled in https://github.com/NVIDIA/nccl/blob/49839dfde34086ee9aaf9a1c69ae3404f13773e7/src/transport/net_ib/connect.cc#L236

limu713:
please close it. 
thank you.











At 2026-05-15 11:59:28, "Anton A" ***@***.***> wrote:

dmvevents left a comment (NVIDIA/nccl#1427)

Friendly check-in: this PR hasn't had activity in 34 days. I plan to revisit it next cycle. If it's no longer needed, let me know and I'll close; otherwise happy to incorporate any remaining feedback.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #1451, opened 2024-09-18
Fix clang missing braces warning

PatriosTheGreat:
Clang has a warning for nested structure initialization in a single braces.
See: https://godbolt.org/z/bs6dn9Pzx 

Not sure if the named initialization would be better here.


stephenmsachs:
Closing the PR since this has been fixed in v2.25. Thanks for fixing!

Issue #598, opened 2021-11-15
QUESTION❓: What is purpose of NCCL_SPINS_BEFORE_CHECK_ABORT


PeterSH6:
We noticed the constant [`#define SPINS_BEFORE_CHECK_ABORT 1000000`](https://github.com/NVIDIA/nccl/blob/v2.7.6-1/src/collectives/device/primitives.h#L14). 
We found that this constant is used in [`checkAbort`](https://github.com/NVIDIA/nccl/blob/v2.7.6-1/src/collectives/device/primitives.h#L102) function. It seems that it will spin for 1000000 times before getting the `comm->abortFlag`, which I think may influence the time to call `ncclCommAbort()`. 

When we change it to smaller number (e.g. 1000, 100), we find that it works well and the time for `ncclCommAbort()` is smaller. 

❓ So I wonder what is the **purpose** of using `NCCL_SPINS_BEFORE_CHECK_ABORT` and why you set this constant to 1000000 (**which is so large**).

sjeaugey:
The flag address `comm->abortFlag` points to an integer in CPU memory. Polling that flag takes a round-trip to CPU memory, so for performance reasons, we don't want to call that except when something is stuck, which is when we'd expect to see `ncclCommAbort` being called.

How much time do you see it taking before the abort is taken into account? We can reduce it if it's causing abort to take a very long time; we just don't want to poll the abort flag in normal working conditions.

vycezhong:
@sjeaugey  We reduced `NCCL_SPINS_BEFORE_CHECK_ABORT` from 1e6 to 1e3 and observed significant drop in time taken by `ncclCommAbort`. In the meantime, we did not observe any significant performance loss. 

|`NCCL_SPINS_BEFORE_CHECK_ABORT` | `ncclCommAbort` Time (s) | 
|---|---|
|1e6| 303.1185 |
|1e3 | 1.449666667 |


Here is our test code and script https://github.com/vycezhong/cs-notes/tree/NCCL/NCCL/01 . Basically, the main thread keeps launching All-Reduce kernels and a background thread checks the error asynchronously.

sjeaugey:
Thanks for the feedback. 300s is indeed too long; we need to reproduce and adjust the counter to something more reasonable.

sjeaugey:
A quick update: we can reproduce the issue. We're suspecting this is because upon abort we continue to run the full algorithm so we will go through the spinning many, many times and stop every time but only after a million spins.

[...]

PR #1462, opened 2024-09-26
Read abortFlag on the first spin in checkAbort

igozali:
While trying to abort the PyTorch [process groups](https://pytorch.org/docs/stable/distributed.html#torch.distributed.init_process_group) that are involved in a hybrid-sharded data parallel job, we noticed that it takes a long time (>2 mins) to call `ncclCommAbort()`.

Upon further debugging, we realized that when `ncclCommAbort()` was called, there are already a bunch of queued NCCL AllGather kernels, but strangely these NCCL kernels that run after abort are taking a long time to exit (much longer than nominal AllGathers). Concretely, nominal AllGathers would take around 1ms to complete, but after `ncclCommAbort` is called, the post-abort AllGathers would take 200ms to complete, and there are many of these AllGather kernels stuck in the queue. Because there's a stream synchronize call involved in `ncclCommAbort()`, these queued kernels then also cause `ncclCommAbort()` to get stuck until the stream can be cleared.

We traced it down to this https://github-ap.tesla.com/AI/nccl/blob/master/src/device/prims_ll.h#L55-L62 call which seems to only be examining the `abortFlag` after 1,000,000 spins.

This PR proposes to examine the `abortFlag` in the 0th `spin` so that there's an opportunity for the NCCL kernels to exit early if `abortFlag` is already set to 1, therefore also speeding up `ncclCommAbort()`.

I was wondering if this approach is reasonable and whether it has major downsides? cc @sjeaugey @KaimingOuyang

### DIFF
--- a/src/device/prims_ll.h
+++ b/src/device/prims_ll.h
@@ -54,11 +54,12 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p>:
   uint32_t abort = 0;
 
   inline __device__ int checkAbort(int &spins, int send) {
-    spins++;
-    if (abort == 0 && spins == NCCL_SPINS_BEFORE_CHECK_ABORT) {
+    // Check abortFlag on the 0th spin and multiples of NCCL_SPINS_BEFORE_CHECK_ABORT
+    if (abort == 0 && spins % NCCL_SPINS_BEFORE_CHECK_ABORT == 0) {
       abort = *ncclShmem.comm.abortFlag;
       spins = 0;
     }
+    spins++;
     return abort;
   }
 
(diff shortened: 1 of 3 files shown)


KaimingOuyang:
That's indeed a concern if you have collectives back-to-back without checking abort flag right away.
Do you see the perf improvement with your patch? If it is indeed helping the case, we can port the patch to the next release.

AddyLaddy:
My main concern with the patch would be that it introduces a GPU initiated PCI-E read on every invocation of that function.
We added the poll after 1M spins to avoid performance issues from generating a lot of PCI-E traffic and also the latency costs of PCI-E reads.
This issue was also raised back in 2021: https://github.com/NVIDIA/nccl/issues/598

igozali:
@KaimingOuyang yes, it indeed does help improve the duration it takes to flush the queued NCCL kernels after the `abortFlag` is set, and thanks @AddyLaddy for sharing that issue, it's good to know that tests from the linked issue confirm this as well.

We also shared similar concerns about reading the page-locked memory on every NCCL kernel launch. I'll close this PR since it doesn't seem to be the best approach and will ask in the issue what some of the good ideas are 😄

PR #1489, opened 2024-10-19
Fix comm ready check in ncclIbIsend and ncclIbIrecv

WWeiOne:
`ncclIbIsend` & `ncclIbIrecv` contains redundant checks for `comm->base.ready`. These checks were made redundant due to a change in commit 5d3ab08, which removed a call to `ncclSendCheck` that might alter the readiness state of the communicator. The second check for `comm->base.ready` now will never executes.

Commit 5d3ab08 Feb 27 2023
```cpp
  if (comm->ready == 0) NCCLCHECK(ncclSendCheck(comm));
  if (comm->ready == 0) { *request = NULL; return ncclSuccess; }
```

Also, the `*request = NULL;` should be kept, as none request is successfully added into the fifo slot, it's safer to assign NULL and consistent with the check `if (sub->requests[buffSlot] != NULL)` in sendProxyProgress.

### DIFF
--- a/src/transport/net_ib.cc
+++ b/src/transport/net_ib.cc
@@ -1824,8 +1824,11 @@ ncclResult_t ncclIbMultiSend(struct ncclIbSendComm* comm, int slot) {
 
 ncclResult_t ncclIbIsend(void* sendComm, void* data, int size, int tag, void* mhandle, void** request) {
   struct ncclIbSendComm* comm = (struct ncclIbSendComm*)sendComm;
-  if (comm->base.ready == 0) { WARN("NET/IB: ncclIbIsend() called when comm->base.ready == 0"); return ncclInternalError; }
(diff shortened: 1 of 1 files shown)


kiskra-nvidia:
Thank you for your submission! We'll consider it for a future release (it missed the 2.24 window).

kiskra-nvidia:
@WWeiOne This should be fixed in NCCL 2.27.3 -- could you verify?

WWeiOne:
yes, it's fixed.

WWeiOne:
> @WWeiOne This should be fixed in NCCL 2.27.3 -- could you verify?

yes, it's fixed.

mnicely:
Should we close this PR then?

WWeiOne:
> Should we close this PR then?

sure

PR #1507, opened 2024-11-07
align the peer mem access check with API document

changchengx:
In API document, it writes below:
  cudaDeviceCanAccessPeer(int* canAccessPeer, int device, int peerDevice )
  [if device device is capable of directly accessing memory from peerDevice](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__PEER.html#group__CUDART__PEER_1g4db0d04e44995d5c1c34be4ecc863f22)

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -2397,7 +2397,7 @@ ncclResult_t  ncclMemAlloc(void **ptr, size_t size) {
     /* Now allow RW access to the newly mapped memory */
     for (int i = 0; i < dcnt; ++i) {
       int p2p = 0;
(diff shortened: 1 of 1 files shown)


AddyLaddy:
I can't see the issue with that code? We're checking for access from the local device `cudaDev` for each peer `i`

changchengx:
> I can't see the issue with that code? We're checking for access from the local device `cudaDev` for each peer `i`
>
Yes. It wants to whether the ```i``` could access to the memory allocated from ```cudaDev```. In the API document, it checks whether the ```device``` could access to the memory allocated from ```peerDevice```. So, I think the argument ```i``` should be assigned to the parameter ```device```.

kiskra-nvidia:
I'm guessing that in practice the order doesn't really matter but I agree with you @changchengx that the order you suggested is correct.

PR #1539, opened 2024-12-12
Fix logging on connection failure

nzmsv:
When a bootstrap socket connection fails, an error message is printed. Part of the message comes from ncclSocketToString.  That function calls getnameinfo to obtain a human-readable string.  However, if getnameinfo failed the function would return garbage from the stack.


kiskra-nvidia:
I agree that it doesn't look great that the return value from `getnameinfo` is being ignored, but could you outline in more detail the failure scenario here?  Was an uninitialized `ncclSocketAddress` being passed to `ncclSocketToString`?

kiskra-nvidia:
@nzmsv This should be fixed in NCCL 2.27.3 -- could you verify?

nzmsv:
Yes indeed - the logic in 2.27.3 is equivalent.  Thank you!

PR #1563, opened 2025-01-03
Comment of a Bootstrap Function  changed

LameloBally:
isFirstFromRoot Function from bootstrap.cc has changed to clearly describe what it does.


AddyLaddy:
Thanks, looks like that was a copy n paste of the comment from `nRankFromRoot()`

PR #1574, opened 2025-01-15
Fix the loop condition in ncclFindInterfaceMatchSubnet as it may not behave as expected when maxIfs > 1

fengquyoumo:
I am a newcomer learning NCCL, and I have come across this part of the code that seems potentially unreasonable.

When maxIfs > 1, we will only find the first IF that matches subnet.


kiskra-nvidia:
@fengquyoumo This should be fixed in NCCL 2.27.3 -- could you verify?

fengquyoumo:
I sincerely apologize for the delayed response.
The logic in 2.27.3 is equivalent. Thank you :)!

Issue #1573, opened 2025-01-10
nccl connection abort(between kubernetes pods): WARN NET/IB: read failed in ncclIbRoceGetVersionNum: Invalid argument


limu713:
We use nccl in kubernetes + rdma-device-plugin. 
Pods communicate by macvlan sub interface of roce hca. Different pod has different gid index. When run miprun between two pods, connection aborts. We trace nccl code and find that nccl tries to read file /sys/class/infiniband/$device/ports/$port_num/gid_attrs/types/$index which does not exist. Actually, its relative gid is 0000:0000:0000:0000:0000:0000:0000:0000 (cat /sys/class/infiniband/$device/ports/$port_num/gids/$index ).

Here are show_gids results in one pod. Each device has existing  gid of index 4,5,6,7. 

root@test-macvlan-pod-2:/# show_gids 
DEV	PORT	INDEX	GID					IPv4  		VER	DEV
---	----	-----	---					------------  	---	---
mlx5_0	1	4	fe80:0000:0000:0000:a4ed:ccff:fe8b:1994			v1	net1
mlx5_0	1	5	fe80:0000:0000:0000:a4ed:ccff:fe8b:1994			v2	net1
mlx5_0	1	6	0000:0000:0000:0000:0000:ffff:0a98:000c	10.152.0.12  	v1	net1
mlx5_0	1	7	0000:0000:0000:0000:0000:ffff:0a98:000c	10.152.0.12  	v2	net1
mlx5_1	1	4	fe80:0000:0000:0000:9459:9bff:fe54:7704			v1	net2
mlx5_1	1	5	fe80:0000:0000:0000:9459:9bff:fe54:7704			v2	net2
mlx5_1	1	6	0000:0000:0000:0000:0000:ffff:0a98:040c	10.152.4.12  	v1	net2
mlx5_1	1	7	0000:0000:0000:0000:0000:ffff:0a98:040c	10.152.4.12  	v2	net2
mlx5_2	1	4	fe80:0000:0000:0000:90ea:b5ff:fec5:3f24			v1	net3
mlx5_2	1	5	fe80:0000:0000:0000:90ea:b5ff:fec5:3f24			v2	net3
mlx5_2	1	6	0000:0000:0000:0000:0000:ffff:0a98:080c	10.152.8.12  	v1	net3
mlx5_2	1	7	0000:0000:0000:0000:0000:ffff:0a98:080c	10.152.8.12  	v2	net3
[...]

PR #1587, opened 2025-01-22
Fix for sparse GID tables (not starting at one)

jlamanna:
This fix is for supporting sparse GID tables. Especially in Kubernetes environments, a pod can end up with a GID table that does not start at "1". This arises when there is another pod on the host that has assigned an IP address to the NIC (usually when they are exposed through MACVLAN to multiple pods).

An example:

```
$ show_gids
DEV     PORT    INDEX   GID                                     IPv4            VER     DEV
---     ----    -----   ---                                     ------------    ---     ---
mlx5_0  1       6       0000:0000:0000:0000:0000:ffff:0b00:0401 11.0.4.1        v1      net1
mlx5_0  1       7       0000:0000:0000:0000:0000:ffff:0b00:0401 11.0.4.1        v2      net1
mlx5_0  1       8       fe80:0000:0000:0000:e4ea:71ff:feb9:4970                 v1      net1
mlx5_0  1       9       fe80:0000:0000:0000:e4ea:71ff:feb9:4970                 v2      net1
```

This patch fixes the existing code which has two problems for dealing with this:

1) The GID index loop can terminate prematurely because of an error
   being thrown in searching for the GID by testing an invalid GID
2) A non-valid GID could be selected (by just comparing address
   families), and then on subsequent tests, the ROCE call would fail,
   resulting in a valid GID being skipped.

### DIFF
--- a/src/transport/net_ib.cc
+++ b/src/transport/net_ib.cc
@@ -353,15 +353,13 @@ static ncclResult_t ncclUpdateGidIndex(struct ibv_context* context, uint8_t port
   bool gidCandidateMatchSubnet = matchGidAddrPrefix(usrFam, prefix, prefixlen, &gidCandidate);
 
   if (gidCandidateFam != gidFam && gidCandidateFam == usrFam && gidCandidateMatchSubnet) {
-    *gidIndex = gidIndexCandidate;
-  } else {
-    if (gidCandidateFam != usrFam || !validGid(&gidCandidate) || !gidCandidateMatchSubnet) {
-      return ncclSuccess;
-    }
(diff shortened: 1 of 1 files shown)


gcongiu:
Hi @jlamanna, this was already reported here: https://github.com/NVIDIA/nccl/issues/1573#issuecomment-2591509563

dmvevents:
Friendly check-in: this PR hasn't had activity in 472 days. I plan to revisit it next cycle. If it's no longer needed, let me know and I'll close; otherwise happy to incorporate any remaining feedback.

PR #1613, opened 2025-02-22
socket: handle short read of the magic

### LINKED ISSUE #1612: [bug] bootstrap may fail when connection close before sending 8 bytes
Bootstrap socket checks request magic (8 bytes) and will ignore if magic doesn't match.
It can greatly avoid failure caused by unexpected network traffic (like security scan).

However , when connection to peer listen socket is closed before 8 bytes are sent, it still will causes bootstrap failure.

ref log:
```
[0] misc/socket.cc:71 NCCL WARN socketProgress: Connection closed by remote peer 10.10.1.3<36138>
[0] NCCL INFO misc/socket.cc:80 -> 6
[0] NCCL INFO misc/socket.cc:474 -> 6
[0] NCCL INFO misc/socket.cc:628 -> 6
[0] NCCL INFO misc/socket.cc:731 -> 6
[0] NCCL INFO bootstrap.cc:701 -> 6
[0] NCCL INFO init.cc:1402 -> 6
[0] NCCL INFO group.cc:75 -> 6 [Async thread]
[0] NCCL INFO group.cc:423 -> 6
[0] NCCL INFO group.cc:573 -> 6
[0] NCCL INFO group.cc:106 -> 6
```

https://github.com/NVIDIA/nccl/blob/80f6bda4378b99d99e82b4d76a633791cc45fef0/src/misc/socket.cc#L474

please consider also ignore peer connections that disconnected before sending 8 bytes.

wenjianhn:
nmap will reset the TCP connection without sending any data during port scanning. As a result, the bootstrap will fail.

This patch fixes the issue by ignore connections that have not send the whole magic before closing.

Closes #1612

### DIFF
--- a/src/misc/socket.cc
+++ b/src/misc/socket.cc
@@ -461,6 +461,18 @@ static ncclResult_t socketSetFlags(struct ncclSocket* sock) {
   return ncclSuccess;
 }
 
+/* Ignore spurious connection and accept again */
+static void handleInvalidMagic(struct ncclSocket* sock) {
+  char line[SOCKET_NAME_MAXLEN+1];
+  WARN("handleInvalidMagic: didn't receive a valid magic from: %s",
+       ncclSocketToString(&sock->addr, line, 0));
+
+  close(sock->fd);
+  sock->fd = -1;
+  sock->state = ncclSocketStateAccepting;
+  sock->finalizeCounter = 0;
+}
+
 static ncclResult_t socketFinalizeAccept(struct ncclSocket* sock) {
   uint64_t magic;
   enum ncclSocketType type;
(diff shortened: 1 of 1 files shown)


gangxie112:
@kiskra-nvidia Can you help have a look at this issue? We hit it recently. It seems that previous fix missed something.

kiskra-nvidia:
@gangxie112 Are you saying that this PR doesn't fully solve the issue for you? Do you have specific instructions on how to reproduce?

I'm a bit surprised that this triggers during the bootstrap for you, given that that stage should take just a few seconds. Are you running nmap in a loop or something?

wenjianhn:
@kiskra-nvidia  we have verified that the PR does fix the issue.
cc @fishautumn

We reproduced the issue by 
```bash
IP=10.0.0.7
nmap -sT $IP -p 40000-60000
```

Since the bootstrap takes only a few seconds, we don't hit the issue often.

PR #1635, opened 2025-03-10
Fix for various issues in the graph module

eleanorTurintech:
The pr aims to fix some issues in the graph/ module. The details are as follows:

- graph/search.cc: removed an unreachable branch.
- graph/topo.cc: fixed an indexing issue when restarting the search.
- graph/tuning.cc: fixed the memory leakage issues when returning early.
- graph/xml.cc: fixed a bug when the xml value is wrapped in single quotes.

### DIFF
--- a/src/graph/search.cc
+++ b/src/graph/search.cc
@@ -328,8 +328,7 @@ ncclResult_t ncclTopoReplayGetGpu(struct ncclTopoSystem* system, struct ncclTopo
     *g = i;
     return ncclSuccess;
   }
-  if (*g == -1) return ncclInternalError;
-  return ncclSuccess;
+  return ncclInternalError;
 }
 
 ncclResult_t ncclTopoSearchRecGpu(struct ncclTopoSystem* system, struct ncclTopoGraph* graph, struct ncclTopoGraph* saveGraph, struct ncclTopoNode* gpu, int step, int backToNet, int backToFirstRank, int forcedOrder, int *time);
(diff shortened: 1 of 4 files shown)


src/graph/search.cc
```diff
@@ -328,8 +328,7 @@ ncclResult_t ncclTopoReplayGetGpu(struct ncclTopoSystem* system, struct ncclTopo
     *g = i;
     return ncclSuccess;
   }
-  if (*g == -1) return ncclInternalError;
-  return ncclSuccess;
+  return ncclInternalError;
```
eleanorTurintech, line 331: In the for-loop above, if a GPU with the specified rank is found, the function sets `*g` to the index of the GPU and return. If the for-loop completes, `*g` would remain as the initial value of `-1`, therefore the if-statement for checking `*g` value is redundant. It also means the final return statement will be unreachable. 

The fix removes the redundant if-statement and return `ncclInternalError` directly.

src/graph/xml.cc
```diff
@@ -51,10 +51,11 @@ ncclResult_t xmlGetValue(FILE* file, char* value, char* last) {
 #endif
   }
   int o = 0;
+  char quote = c;  // Remember which quote type we started with
   do {
     NCCLCHECK(xmlGetChar(file, &c));
     value[o++] = c;
-  } while (c != '"');
+  } while (c != quote);
```
eleanorTurintech, line 58: The fix allows the quote to be either `"` or `'`, and correctly matches the ending quote.

marksantesson:
Thanks, @eleanorTurintech ! I've upstreamed your change into what should be our next new release, v2.27.

kiskra-nvidia:
@eleanorTurintech This should be fixed in NCCL 2.27.3 -- could you verify?

Issue #1658, opened 2025-03-24
NCCL profiler proxyStep start/stop counts mismatch in plugin v3.


jingxu9x:
Thansk for you work on version 2.26.2, Now I can check the kernel events with the profiler. However, during testing I ran into a strange problem.

when I force nccl choice NET/Socket do allreduce in a 8-gpu machine, the proxyStep start event count is alway larger than stop event count. By all rights, this value should be equal.

I attach a patch file, hope it will help you find out the root cause.

The environment variables as follow:

```
export NCCL_P2P_DISABLE=1
export NCCL_P2P_DIRECT_DISABLE=0
export NCCL_SHM_DISABLE=1

export NCCL_PROFILE_EVENT_MASK=255
```

test cmd is
```
./all_reduce_perf -d float -b 1G -e 1G -f 2 -g 8 -c 1 -w 1 -n 1
```

[issue_reproduction.patch](https://github.com/user-attachments/files/19416763/issue_reproduction.patch)

gcongiu:
I don't think I am following. Can you attach a snap of the output you are getting from the profiler?

jingxu9x:
Yes, I also provide a script to reproduce this problem

![Image](https://github.com/user-attachments/assets/732abd6e-c976-497a-ab33-b7ed8eddfa08)

Ideally, this ProxyStepCnt should be 0. (because startEevent and stopEvent should be paired)

```
#! /bin/bash
set -xe

WORKDIR=/tmp

# apt-get update && apt install -y libibverbs-dev infiniband-diags openmpi-bin libopenmpi-dev net-tools

# build nccl
# git clone https://github.com/NVIDIA/nccl.git -b v2.26.2-1 $WORKDIR/nccl
# git -C $WORKDIR/nccl apply `pwd`/issue_reproduction.patch

pushd $WORKDIR/nccl
make -j src.build \
  TRACE=1 \
  NET_PROFILER=1 \
  DEBUG=1 \
  NVCC_GENCODE="-gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_89,code=sm_89 -gencode=arch=compute_90,code=sm_90"

# build profiler
make -C ext-profiler/example -j \
    TRACE=1 \
    NVCC_GENCODE="-gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_89,code=sm_89 -gencode=arch=compute_90,code=sm_90"
popd

# build nccl test
[...]

PR #1682, opened 2025-04-14
[nccl-profiling] fix bug on calling wrong id for proxy step in recvProxyProgress

jushg:
This bug seems to be the root cause of #1658

Our team profiler implementation started to have some proxyStep events without a recorded end timestamp with nccl v2.24.3 (working fine in nccl v2.23.4). 

I looked through the source code and narrowed down the issue to be because of this logical error in `recvProxyProgress`, since`doneStepId` is created outside of the while loop, but is used for all the steps inside the while loop --> we have proxy step events which are not updated properly. 

Let me know if more context is needed for verifying this fix

### DIFF
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -1479,7 +1479,6 @@ static ncclResult_t recvProxyProgress(struct ncclProxyState* proxyState, struct 
       struct ncclProxySubArgs* subGroup = args->subs+s;
       for (int i=0; i<subGroup->groupSize; i++) {
         struct ncclProxySubArgs* sub = subGroup + i;
-        int doneStepId = sub->done;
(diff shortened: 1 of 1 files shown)


gcongiu:
Thank you for the fix @jushg

jushg:
@gcongiu can you help me confirm if this is a correct fix? I want to back-port it to our internal NCCL repo if it make sense

gcongiu:
I can confirm the fix makes sense. I checked the git history and, indeed, the introduction of `doneStepId` altered the original behavior of the code w.r.t. v2.23.x. This fix will be part of v2.27

jushg:
Awesome, thank you for the confirmation! I will close this PR now.

Issue #1743, opened 2025-06-16
Compilation issue with gcc 14


albanD:
I am trying to compile PyTorch on Fedora 41 which comes with gcc 14.3.1.
I am using the latest cuda toolkit V12.9.86 (according to nvcc) as 12.8 doesn't support gcc 14.
When building with NCCL enabled there, I'm getting compilation errors as:

```bash
/usr/include/c++/14/type_traits(1610): error: "__is_nothrow_new_constructible" is not a function or static data member
      constexpr bool __is_nothrow_new_constructible
                     ^

/usr/include/c++/14/type_traits(1610): error: "constexpr" is not valid here
      constexpr bool __is_nothrow_new_constructible
      ^

/usr/include/c++/14/bits/stl_pair.h(1190): error: "__is_pair" is not a function or static data member
      inline constexpr bool __is_pair = false;
                            ^

/usr/include/c++/14/bits/stl_pair.h(1190): error: "constexpr" is not valid here
      inline constexpr bool __is_pair = false;
             ^

/usr/include/c++/14/bits/stl_pair.h(1193): error: __is_pair is not a template
      inline constexpr bool __is_pair<pair<_Tp, _Up>> = true;
                            ^

/usr/include/c++/14/bits/stl_pair.h(1193): error: "constexpr" is not valid here
      inline constexpr bool __is_pair<pair<_Tp, _Up>> = true;
             ^

6 errors detected in the compilation of "/home/albandes/local/pytorch/3.12_debug_source/build/nccl/obj/device/gensrc/reduce_scatter_minmax_bf16.cu".
```
Similar errors exist in `common.cu` and `reduce_scatter_minmax_f32.cu` as well.

Is that a know issue or might be related to the version being used within the PyTorch build process?

jsmolic:
Hi, I hit the same error on latest CentOS Stream where gcc-14 is also the default compiler. The cxx standard defaults to `c++11` in https://github.com/NVIDIA/nccl/blob/master/makefiles/common.mk#L77 which doesn't work with `gcc-14` , changing that to `-std=c++14` works around the compilation errors, so perhaps the default standard could be updated to 14 in the source?

cameronrutherford:
I ran into this compiling on Ubuntu22.04 with `gcc@14.3.0`, `cuda@12.9`. I am hoping that adding `CXXSTD="-std=c++20"` to my `configure` invocation fixes things, otherwise I might have to resort to installing with Spack.

[...]

PR #1751, opened 2025-06-23
makefiles: common.mk: update c++ standard to c++14

jsmolic:
Compilation fails using newer versions of gcc (gcc-14 specifically) due to the c++11 standard default set by the build system:

```
/usr/include/c++/14/type_traits(1610): error: "__is_nothrow_new_constructible" is not a function or static data member
      constexpr bool __is_nothrow_new_constructible
                     ^

/usr/include/c++/14/type_traits(1610): error: "constexpr" is not valid here
      constexpr bool __is_nothrow_new_constructible
      ^

/usr/include/c++/14/type_traits(1610): error: "__is_nothrow_new_constructible" is not a function or static data member
      constexpr bool __is_nothrow_new_constructible
                     ^

/usr/include/c++/14/type_traits(1610): error: "constexpr" is not valid here
      constexpr bool __is_nothrow_new_constructible
      ^

/usr/include/c++/14/type_traits(1610): error: "__is_nothrow_new_constructible" is not a function or static data member
      constexpr bool __is_nothrow_new_constructible
                     ^

/usr/include/c++/14/type_traits(1610): error: "constexpr" is not valid here
      constexpr bool __is_nothrow_new_constructible
      ^

/usr/include/c++/14/type_traits(1610): error: "__is_nothrow_new_constructible" is not a function or static data member
      constexpr bool __is_nothrow_new_constructible
                     ^

/usr/include/c++/14/type_traits(1610): error: "constexpr" is not valid here
      constexpr bool __is_nothrow_new_constructible
```

Update the standard to c++14 to allow compilation with modern toolchain versions.

Closes: https://github.com/NVIDIA/nccl/issues/1743

### DIFF
--- a/makefiles/common.mk
+++ b/makefiles/common.mk
@@ -76,7 +76,7 @@ $(info NVCC_GENCODE is ${NVCC_GENCODE})
 ifeq ($(shell test "0$(CUDA_MAJOR)" -ge 13; echo $$?),0)
   CXXSTD ?= -std=c++17
 else
-  CXXSTD ?= -std=c++11
+  CXXSTD ?= -std=c++14
 endif
 
 CXXFLAGS   := -DCUDA_MAJOR=$(CUDA_MAJOR) -DCUDA_MINOR=$(CUDA_MINOR) -fPIC -fvisibility=hidden \


kiskra-nvidia:
Yeah, there appears to be an incompatibility between nvcc and gcc 14 header files when running with `-std=c++11`. I agree with your fix (we're planning to include it in the next 2.27 update); I just hope nothing else will break on some old distros...

jsmolic:
I did a quick test on ubuntu with gcc-6 and it compiled without issues, gcc-5 on the other hand fails (even unrelated to this change) but hopefully no one is still using that

PR #1758, opened 2025-07-01
NCCL fails occasionally in Bootstrap failing to accept socket connections due to EINTR

saifhhasan:
At Meta, we've identified a recurring issue where some jobs experience intermittent failures during bootstrapping communication. Specifically, the socketTryAccept function encounters an error when the system returns an EINTR (interrupted system call) value.

With this code change, we handle EINTR so that `accept` can continue with occasional interrupts.

### DIFF
--- a/src/misc/socket.cc
+++ b/src/misc/socket.cc
@@ -441,7 +441,8 @@ static ncclResult_t socketTryAccept(struct ncclSocket* sock) {
   if (sock->fd != -1) {
     sock->state = ncclSocketStateAccepted;
   } else if (errno == ENETDOWN || errno == EPROTO || errno == ENOPROTOOPT || errno == EHOSTDOWN ||
-             errno == ENONET || errno == EHOSTUNREACH || errno == EOPNOTSUPP || errno == ENETUNREACH) {
+             errno == ENONET || errno == EHOSTUNREACH || errno == EOPNOTSUPP || errno == ENETUNREACH ||
+             errno == EINTR) {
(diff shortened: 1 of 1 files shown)


marksantesson:
Thanks! Your change has been merged. It should be in the v2.27.6 update.

stephenmsachs:
Thanks @saifhhasan for your contribution. I'll go ahead and closer this PR now since it has been merged into the next release.

kiskra-nvidia:
@saifhhasan Sorry for a late follow-up. What NCCL versions have you been experiencing this issue with?

kiskra-nvidia:
@saifhhasan Ping on this one. While your contribution was merged into 2.27.6 and is part of that release, I believe that this bug was in fact fixed in NCCL 2.27.3 and your patch shouldn't have been necessary. I'm planning to revert it for NCCL 2.28 unless I hear from you that the bug was observed with NCCL 2.27 releases prior to 2.27.6.

saifhhasan:
> @saifhhasan Ping on this one. While your contribution was merged into 2.27.6 and is part of that release, I believe that this bug was in fact fixed in NCCL 2.27.3 and your patch shouldn't have been necessary. I'm planning to revert it for NCCL 2.28 unless I hear from you that the bug was observed with NCCL 2.27 releases prior to 2.27.6.

Hi @kiskra-nvidia - Pardon me for delayed response, I was on leave for a while. We experienced this issue across all NCCL versions from 2.19 all the way to 2.25 .. If it is presented and fixed in 2.27.3 we can ignore skip this PR. Thank you for double checking.

PR #1782, opened 2025-07-15
support binding client socket address to specific ifname

baymaxhuang:
**background**
In nccl, when we use  NCCL_SOCKET_IFNAME to specify which IP interfaces to use for communication, only the listening socket (server side) will bind a specific ifname address.  The client socket would use the default address. It would cause some unexpected problems when there are link problems.

**Solution**
Add a peerAdd in ncclSocket to save peer address of the socket, and the local address is saved in addr. When a client connects to a remote socket (sock->peerAdd), the source address would also be bound to sock->addr.

**TODO**
Init rasnet client socket with a specific address that rasNetListeningSocket also uses. At present, ncclSocketInit uses nullptr and it would be bound to the default address of the system.

### DIFF
--- a/src/bootstrap.cc
+++ b/src/bootstrap.cc
@@ -258,7 +258,7 @@ static ncclResult_t setFilesLimit() {
 static ncclResult_t rootSend(union ncclSocketAddress* addr, uint64_t magic, union ringConnectInfo* info) {
   ncclResult_t res = ncclSuccess;
   struct ncclSocket sock;
-  NCCLCHECKGOTO(ncclSocketInit(&sock, addr, magic, ncclSocketTypeBootstrap), res, fail);
+  NCCLCHECKGOTO(ncclSocketInit(&sock, &bootstrapNetIfAddr, addr, magic, ncclSocketTypeBootstrap), res, fail);
   NCCLCHECKGOTO(ncclSocketConnect(&sock), res, fail);
   NCCLCHECKGOTO(socketSend(&sock, info, sizeof(union ringConnectInfo)), res, fail);
   NCCLCHECK(ncclSocketClose(&sock));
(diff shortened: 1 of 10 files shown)


gangxie112:
@kiskra-nvidia Can you help have a look at this?

kiskra-nvidia:
Sorry for the delay in getting back to you regarding this PR.

Could you explain the motivation for this change? NCCL currently relies on the OS kernel to bind a connecting socket to the correct interface, depending on the destination address. "It would cause some unexpected problems when there are link problems" doesn't really provide sufficient details. What sort of configuration doesn't work correctly? Is it perhaps one where multiple NICs of a single node are on the same subnet?

gangxie112:
> Sorry for the delay in getting back to you regarding this PR.
> 
> Could you explain the motivation for this change? NCCL currently relies on the OS kernel to bind a connecting socket to the correct interface, depending on the destination address. "It would cause some unexpected problems when there are link problems" doesn't really provide sufficient details. What sort of configuration doesn't work correctly? Is it perhaps one where multiple NICs of a single node are on the same subnet?

Thanks for the reply.  Actually, it's not a nccl's defect. We recently found that sglang might use the stale port, causing wrongly access to nccl's tcp port (It may be some service discovery problem of sglang). You know, nccl's socket error handling has some issue and could crash or hang. So, as a work around, we want to isolated nccl's tcp stack from the others in the cluster by using GPU hca instead. But some cloud provider limits the ip route of the RoCE hca used by GPU, and OS hca ordering could hardly unified. So, we need extact socket binding by specified the nccl_ib_hca.

kiskra-nvidia:
I still don't understand -- why doesn't `NCCL_SOCKET_IFNAME` meet your needs? Granted, it will bind the listening sockets, not the connecting ones, but under what circumstances is that not enough?

baymaxhuang:
This PR is no longer needed, thx.

PR #1789, opened 2025-07-22
NCCL Fast Init - CPU Optimizations for NCCL Initialization Large Scale

saifhhasan:
## Problem
At large scale 32K+ GPUs we start to see a significant initialization time coming from `ncclBuildRing` and `initTransportsRank` - often several dozens of seconds at 100K scale. This occurs because both of these functions performs nested loops of complexity O(N*N), which at 100K scale translates to 10B loop iterations, each loop iteration with multiple instructions.

We used CPU Profiling tools that helped us spot the loops that were taking excessive amount of time during the initialization phase.

Screenshot 2025-07-22 at 11 20 18 AM

Screenshot 2025-07-22 at 11 21 24 AM

## Observations
This changes optimizes the two functions to remove the overhead. And **enables NCCL to initialize fast at 100K scale**. We tested this at Meta at Scale and were able to observe following savings in numbers.
* At 96K (96*1024) scale `ncclBuildRing` saves 26s of busy CPU cycles. 
* At 48K (48*1024) scale `initTranportsRank` optimization saves of 11+ seconds

Most **notably this patch helps reduce CPU Utilization during job startup phase** which is also crucial for other ongoing operations at a job level (Checkpoint initialization, Model Loading, Data Fetching etc.). 

## Testing
At meta we've tested these fixes on our large scale clusters. On the top I'm also adding a simple stanalone benchmark to demonstrate the performance improvement of `ncclBuildRings` before and after. The numbers are as below for varying number of ranks from 1024 to 96K. With 16 rings, we save about 13s, for 32 rings it'll be about 52s of duration.

```
Run on (368 X 2396.4 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x368)
  L1 Instruction 64 KiB (x368)
  L2 Unified 512 KiB (x368)
  L3 Unified 16384 KiB (x1)
Load Average: 8.19, 7.24, 9.36
---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
//
// Without the Fix
//
BM_ncclBuildRings/1024              1648154 ns      1648085 ns          415
BM_ncclBuildRings/2048              6378142 ns      6377981 ns          112
BM_ncclBuildRings/4096             23255134 ns     23253167 ns           30
BM_ncclBuildRings/8192             90435922 ns     90432606 ns            8
BM_ncclBuildRings/16384           358064532 ns    358047348 ns            2
BM_ncclBuildRings/32768          1430027485 ns   1429897998 ns            1
BM_ncclBuildRings/65536          5720381498 ns   5719983433 ns            1
BM_ncclBuildRings/98304          12864594936 ns   12863500960 ns            1

//
// With the Fix in this PR
//
BM_ncclBuildRingsOptimized/1024       30666 ns        30665 ns        22790
BM_ncclBuildRingsOptimized/2048       60711 ns        60710 ns        11498
BM_ncclBuildRingsOptimized/4096      120775 ns       120771 ns         5795
BM_ncclBuildRingsOptimized/8192      243442 ns       243431 ns         2878
BM_ncclBuildRingsOptimized/16384     502392 ns       502383 ns         1398
BM_ncclBuildRingsOptimized/32768    1006667 ns      1006595 ns          696
BM_ncclBuildRingsOptimized/65536    2010650 ns      2010633 ns          345
BM_ncclBuildRingsOptimized/98304    3003351 ns      3003085 ns          233
```

### Test Binary Code
Can be compiled standalone by linking with gtest and google-benchmark
[nccl-build-ring-benchmark.cpp.txt](https://github.com/user-attachments/files/21373371/nccl-build-ring-benchmark.cpp.txt)

### DIFF
--- a/src/graph/rings.cc
+++ b/src/graph/rings.cc
@@ -4,6 +4,8 @@
  * See LICENSE.txt for license information
  ************************************************************************/
 
+ #include <vector>
+
 #include "core.h"
 
 void dumpLine(int* values, int nranks, const char* prefix) {
@@ -25,7 +27,7 @@ void dumpLine(int* values, int nranks, const char* prefix) {
   INFO(NCCL_INIT, "%s", line);
 }
 
-ncclResult_t ncclBuildRings(int nrings, int* rings, int rank, int nranks, int* prev, int* next) {
+int ncclBuildRings(int nrings, int* rings, int rank, int nranks, int* prev, int* next) {
   for (int r=0; r<nrings; r++) {
     char prefix[40];
     /*sprintf(prefix, "[%d] Channel %d Prev : ", rank, r);
@@ -33,8 +35,10 @@ ncclResult_t ncclBuildRings(int nrings, int* rings, int rank, int nranks, int* p
     sprintf(prefix, "[%d] Channel %d Next : ", rank, r);
     dumpLine(next+r*nranks, nranks, prefix);*/
 
+    std::vector<bool> rankBitSet(nranks, false);
     int current = rank;
     for (int i=0; i<nranks; i++) {
+      rankBitSet[current] = true;
       rings[r*nranks+i] = current;
       current = next[r*nranks+current];
     }
@@ -46,14 +50,7 @@ ncclResult_t ncclBuildRings(int nrings, int* rings, int rank, int nranks, int* p
     }
     // Check that all ranks are there
     for (int i=0; i<nranks; i++) {
-      int found = 0;
-      for (int j=0; j<nranks; j++) {
-        if (rings[r*nranks+j] == i) {
-          found = 1;
-          break;
-        }
-      }
-      if (found == 0) {
+      if (!rankBitSet[i]) {
         WARN("Error : ring %d does not contain rank %d", r, i);
         return ncclInternalError;
       }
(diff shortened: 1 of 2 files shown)


saifhhasan:
Hi @sjeaugey and @marksantesson - I've sent out a new PR for optimizing init times at large scale that surprised us at Meta. We believe it'll be of help to broader community. I'm happy to iterate with you to improve the PR based on your feedback.

stephenmsachs:
Hi @saifhhasan. Thanks for your contribution. I have a question about the  `initTransportsRank` changes the implementation: 
The original implementation checks every pair (i,j) of `comm->peerInfo[i]`, `comm->peerInfo[j]` and disables `nvlsRegSupport` if any two ranks sharing host & pid are found. The proposed implementation only checks if any `comm->peerInfo[i]` shares a process with current rank `comm->peerInfo[comm->rank]`. Since `comm->nvlsRegSupport` is process-local I believe this can lead to a different outcome.

stephenmsachs:
Hi @saifhhasan . We will have a similar change in `rings.cc` in the next release. This will bring down the complexity a little more. I would still appreciate clarification concerning your proposed changes in `init.cc`. Maybe I am just not understanding them correctly.

saifhhasan:
Thank you @stephenmsachs for helping take a look at the PR.

Regarding init.cc changes, what you present could be the case edge case. `comm->nvlsRegSupport` may have different values on various ranks with this change, but in practice due to execution scheduling we never run into this case. We may be able to overcome this by using `std::unordered_set<pair<hostHash, pidHash>>` to achieve O(N-LogN) complexity.

xiaofanl-nvidia:
This has been accepted and released in the latest 2.28.7 release. @saifhhasan

PR #1791, opened 2025-07-24
NCCL Fast Init - Improve Bootstrap AllGather by 2x at large scale

saifhhasan:
## Problem
The default specification of AllGather is to send to the next peer and receive from a previous one. In the case of RDMA this is the most optimal strategy. However, in the case of bootstrap we’re exchanging trivial data chunks  ~100 bytes) over TCP sockets from a process running on CPU. Thus saturating NIC line rate is not of a concern and most of our performance bottleneck comes from the number of steps performed for AllGather which equals to N-1 where N is number of processes.

## Solution
We propose to leverage bidirectional AllGather, where for each step we send and receive a unit of information to both of our peers. With this it’ll require us N/2 steps to conclude all gathers.

Screenshot 2025-07-23 at 6 27 03 PM

## Testing
* Run simple Collectives to ensure code-path is working as intended
* At Meta we ran this change at large scale in our test-setup showing notable improvement in init times. We ran bootstrap AllGather multiple times in a row (to warmup) and use a steady state number. Numbers on across all process falls within range of few 100ms 

NOTE: These numbers are used on a standalone test for collective initialization only. The actual numbers would be higher in a training workload as system will have more load on the CPU (based on our observations)

Before - 9.1s
```
NCCL INFO Init timings - ncclCommInitRankConfig: rank 46246 nranks 98304 total 57.96 (kernels 0.00, alloc 0.00, bootstrap 46.74, allgathers 9.12, topo 1.04, graphs 0.03, connections 0.39, rest 0.64)
```

After - 5.55s 
```
NCCL INFO Init timings - ncclCommInitRankConfig: rank 12356 nranks 98304 total 37.34 (kernels 0.00, alloc 0.00, bootstrap 29.56, **allgathers 5.55**, topo 1.14, graphs 0.03, connections 0.40, rest 0.67)
```

### DIFF
--- a/src/bootstrap.cc
+++ b/src/bootstrap.cc
@@ -4,6 +4,10 @@
  * See LICENSE.txt for license information
  ************************************************************************/
 
+#include <atomic>
+#include <future>
+#include <thread>
+
 #include "nccl.h"
 #include "core.h"
 #include "utils.h"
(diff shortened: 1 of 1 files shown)


sjeaugey:
Question about the background of this PR: in NCCL 2.23, for large scale init optimization, we implemented an [IB/RoCE based allgather](https://developer.nvidia.com/blog/new-scaling-algorithm-and-initialization-with-nvidia-collective-communications-library-2-23/#accelerated_bootstrap_operations) which is supposed to be way faster than the socket implementation. Was there a reason why you didn't want to use that implementation and preferred to optimize the socket implementation instead?

saifhhasan:
That's a good point. We kind of did this when NCCL-2.21 was around. But it is something that we can try out now. However we may not be able to test it upfront due to resources. TCP Solution we were able to test and validate using the CPU Emulation based test runs with mocked Cuda/IB. 

Though I believe, this can also give some improvement to IB/RoCE based allGather too, however the absolute improvement won't be of too much importance at 100K scale. But for TCP the absolute gain would be quite noticeable.

xiaofanl-nvidia:
This has been accepted and released in the latest 2.28.7 release. @saifhhasan

stephenmsachs:
I am going to close this since the [bidirectional ring](https://github.com/NVIDIA/nccl/blob/master/src/bootstrap.cc#L1049) has been released with v2.28.7. Thanks again for you input.

PR #1811, opened 2025-08-15
NCCL 2.28.1-1 Early Access

marksantesson:
## NCCL 2.28.1-1 Early Access

This is an Early Access release. It contains new features and APIs that may be changed before the official General Access release. In particular, the Device API is not frozen. It also may contain performance degradation that is not representative of the General Access release.

See the Known Issues section for problems that we are still working on.

### Device API
- Provide a device-side API for integrating NCCL communication into application kernels.
- This is currently supported for NVLink/P2P communication, including multicast (NVLS) and low-latency (LL) protocol.
- Device API signature and functionalities are subject to change.

### Symmetric memory improvements
- Support for aggregating symmetric operations.
- Reimplement symmetric kernels using device API.

### New Host APIs
- Introduce new host collective APIs: ncclAlltoAll, ncclScatter, ncclGather.

### CE (Copy Engine) Collectives
- Reduce SM utilization for alltoall, scatter, gather, and allgather within a single (MN)NVL domain.
- Free up SM capability for the application.
- To enable the feature, register buffers into symmetric windows and use the NCCL_CTA_POLICY_ZERO flag in the communicator config_t.

### Plugins
- Network
  - App-aware Network plugin. NCCL passes information about communication operations to be executed on the network     end point. This allows for better tuning of network end points and their use in the plugins. 
  - Improve handling of physical and virtual network devices and load/unload.
  - Network plugin version 11 - add explicit context and communication ID support for per communicator init/finalize.
- Profiler
  - Add support for API events (group, collective, and kernel launch) and for tracking kernel launches in the profiler plugin.
  - Add Inspector Profiler Plugin for NCCL Collective Metric Collection.
- Tuner
  - Expose NCCL tuning constants at tuner initialization via ncclTunerConstants_v5_t.
  - Add NVL Domain Information API.
- Support multiple plugin types from a single shared object.
- Improve lifecycle management and error handling.
- Clean up diagnostics during plugin loading.

### CMake support
- Adds a CMake build system as an alternative to existing Makefiles. 

Add new option `NCCL_MNNVL_CLIQUE_ID=-2` which will use rack serial number to partition the MNNVL clique. This will limit NVLink domains to GPUs within a single rack.

NVLSTree Tuning Fixes. Update tuning data for H100, GB200-NV72.

Known Issues that are still being worked on
- ncclDevCommCreate hangs in case of multiple GPUs per thread.
- pkg.build doesn’t work with Cmake.
- Large-scale CE collectives have known performance degradation.
- Backward compatibility layer for network plugins doesn’t work with AWS EFA plugin.
- Various performance regressions, some only affecting specific combinations of hardware and algorithms. Do not expect that the current performance is indicative of what the GA release will have.


alokprasad:
@marksantesson do we have some performance improvement/degradation result available for sm vs ce implementation of collectives,

sjeaugey:
I'm not sure we have number to share, but the point of the CE implementation is to free SMs. 

The peak bandwidth may be a bit higher since CE use 256B payload vs 128B for SM operations. While the latency can be worse because multiple CE operations are longer to launch than a single kernels.

But overall, the performance difference (benchmark-wise) is not huge, while the difference in SM usage is very significant. Comparing performance from the perf test would be mostly missing the point.

If NCCL operations are large enough and overlapped with compute kernels, there is a high chance that this feature would improve the performance of compute kernels significantly and make a difference in end-to-end run time. If NCCL operations are fully exposed, i.e. they run along on the GPU, it will probably not make a big difference (and could even be worse depending on the size).

alokprasad:
@sjeaugey thats useful info, btw is there any NVIDIA tools that shows SM and CE utilization ( directly or indirectly)
quick search shows NVIDIA DCGM (Data Center GPU Manager) ? NCU?

PR #1844, opened 2025-09-15
Allocate nvtxGlobals_t as small data

pranavk:
Some shipped CUDA libraries (eg: cusolver) use PC32 relocations to reference nvtxGlobals_v3. This breaks some clients that allocate all data in large sections (use -mcmodel=medium -mlarge-data-threshold=0).

Annotate nvtxGlobals_v3 properly to avoid breaking such clients.

### DIFF
--- a/src/include/nvtx3/nvtxDetail/nvtxImpl.h
+++ b/src/include/nvtx3/nvtxDetail/nvtxImpl.h
@@ -70,6 +70,13 @@ extern "C" {
 #pragma GCC visibility push(hidden)
 #endif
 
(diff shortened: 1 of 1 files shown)


dmvevents:
Friendly check-in: this PR hasn't had activity in 236 days. I plan to revisit it next cycle. If it's no longer needed, let me know and I'll close; otherwise happy to incorporate any remaining feedback.

pranavk:
it's still needed if someone is building their binaries with a mixture of small and medium/large code model.

PR #1846, opened 2025-09-16
xml.cc: fix memory leak

liangxs:


sjeaugey:
Thanks. I fixed the indentation (tab vs spaces) but other than that, it's going to be merged to the next release.

stephenmsachs:
This has been [fixed](https://github.com/NVIDIA/nccl/blob/master/src/graph/xml.cc#L645). Thanks again for your contribution.

PR #1870, opened 2025-10-09
fix: nccl_net.h path error

zzwzds:
a small bug(head file path error)

### DIFF
--- a/src/include/gin/gin_host.h
+++ b/src/include/gin/gin_host.h
@@ -9,7 +9,7 @@
 
 #include "allocator.h"
 #include "nccl.h"
-#include "nccl_net.h"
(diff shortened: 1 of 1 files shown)


stephenmsachs:
Thanks for reporting. I am unable to re-create the issue. I see that `include/plugin` is added to the header search path [here](https://github.com/NVIDIA/nccl/blob/v2.28_gin_preview/src/Makefile#L63).

zzwzds:
> Thanks for reporting. I am unable to re-create the issue. I see that `include/plugin` is added to the header search path [here](https://github.com/NVIDIA/nccl/blob/v2.28_gin_preview/src/Makefile#L63).

thanks,i did not saw this

PR #1881, opened 2025-10-28
plugin/net.cc: Fix segfault in ncclGin initialization

hz-cheng:
If ncclGinIbGdaki.devices() fails after init() succeeds, pluginLib->ncclGin is left at -1, leading to a segfault when calling pluginLib->ncclGin->init().


stephenmsachs:
Thanks a lot for reporting. We are looking into this right now.

hz-cheng:
Correct: 
In my case, `ncclGinIbGdaki.devices(&ndev) == ncclSuccess` is true, but `ndev > 0` is false. If either of these conditions is not met, it will cause a segmentation fault.

xiaofanl-nvidia:
This fix has been merged and will be released in 2.29. Thank you for the contribution @hz-cheng

hz-cheng:
> This fix has been merged and will be released in 2.29. Thank you for the contribution @hz-cheng

You're welcome, and I will close this PR.

PR #1912, opened 2025-11-17
cmake: open version.mk relative to repo root

sielicki:
## Description

under FetchContent usage, CMAKE_SOURCE_DIR does not refer to the nccl project root until project() is invoked, but we need to parse the version before we can call project(). Avoid by using CMAKE_CURRENT_LIST_DIR instead, which remains stable between project() calls.

## Related Issues

N/A

## Changes & Impact

N/A, build annoyance.

## Performance Impact

N/A


xiaofanl-nvidia:
@nv-udeodhar to take a look.

PR #1971, opened 2026-01-05
fix: Invalid nullptr initialization for CUmemGenericAllocationHandle when compiling with Clang

jushg:
## Description

CUmemGenericAllocationHandle is defined as unsigned long long (not a pointer type). Initializing it with nullptr is invalid in C++ since nullptr cannot be implicitly converted to an integer type. This causes a compilation error with Clang:

```
error: cannot initialize a member subobject of type 'CUmemGenericAllocationHandle' 
(aka 'unsigned long long') with an rvalue of type 'std::nullptr_t'
```

### Why this was not caught with Makefile build? 

NVIDIA's Makefile build uses NVCC with GCC as the host compiler, which historically allows implicit nullptr to integer conversions as a GCC extension (with only a warning). Clang rejects this as a hard error per the C++ standard.

## Changes & Impact

This was the  original code in v2.28, only change in v2.29 to nullptr, so this PR just revert that.

### DIFF
--- a/src/transport/net_ib/gdaki/gin_host_gdaki.cc
+++ b/src/transport/net_ib/gdaki/gin_host_gdaki.cc
@@ -283,7 +283,7 @@ class GdakiGlobalGPUBufferTable {
   uint32_t *get_rkeys_d() { return this->rkeys_hd_mhandle.gpu_buf; }
 
   GdakiGlobalGPUBufferTable()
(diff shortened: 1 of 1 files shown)


lindsayad:
would be great to see this merged

teojgo:
This has been fixed in [v2.29.7-1](https://github.com/NVIDIA/nccl/releases/tag/v2.29.7-1).

teojgo:
Closing this since the fix is already in place.

PR #2027, opened 2026-02-26
Fix buffer overflow in parseStringList

### LINKED ISSUE #2026: [Issue]: Stack Buffer Overflow in `parseStringList`

Application crash

### Share Your Debug Logs

_No response_

We identified a stack buffer overflow in NCCL's configuration string parsing via fuzzing. It is reachable via the public NCCL API and reproducible with ASAN instrumentation.

## Summary

| # | Type | CWE | Function | File:Line |
|---|------|-----|----------|-----------|
| 1 | Stack Buffer Overflow | 121 | `parseStringList` | utils.cc:168 |

The bug is triggered when an application initializes NCCL with a crafted environment variable or config file entry.

This is a sample application to reproduce the bug, when compiled with ASAN:

```cpp
#include <nccl.h>
#include <stdio.h>

int main(void) {
    ncclUniqueId id;
    ncclResult_t res = ncclGetUniqueId(&id);
    printf("ncclGetUniqueId: %d\n", res);
    return 0;
}
```
To trigger the bug with an environment variable:

```bash
NCCL_COMM_ID=$(python3 -c "print('A'*128)") ./poc
```

Alternatively, to trigger it via configuration file:

```bash
python3 -c "print('NCCL_COMM_ID=' + 'A'*128)" > ~/.nccl.conf
./poc
```

## Stack Buffer Overflow in parseStringList

The `netIf` structure defines a 64-byte `prefix` buffer:

```c
// src/include/utils.h
struct netIf {
  char prefix[64];
  int port;
};
```

`parseStringList` copies each character into `prefix` without a bounds check:

```c
// src/misc/utils.cc:167-170
} else {
    ifList[ifNum].prefix[ifC] = c;
    ifC++;
}
```

The loop index `ifC` is incremented on every non-delimiter character but never compared against the buffer size. When the input contains more than 63 characters without a `,` or `:` delimiter, the write overflows the 64-byte `prefix` array into adjacent stack memory.

The bug can be triggered via environment variables (examples: `NCCL_COMM_ID`, `NCCL_SOCKET_IFNAME`).

These variables can also be injected via NCCL configuration files (`~/.nccl.conf`, `/etc/nccl.conf`, or a custom path in `$NCCL_CONF_FILE`), which are read during initialization and applied via `ncclOsSetEnv`.

### ASAN Output

```
==14==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7b1d7a7f1064 at pc 0x556781cfac69 bp 0x7ffe993a35e0 sp 0x7ffe993a35d8
WRITE of size 1 at 0x7b1d7a7f1064 thread T0
    #0 0x556781cfac68 in parseStringList(char const*, netIf*, int) /src/nccl/src/misc/utils.cc:168:33
    #1 0x556781d0067a in ncclSocketGetAddrFromString(ncclSocketAddress*, char const*) /src/nccl/src/misc/socket.cc:213:9
    #2 0x556781dcebbc in bootstrapNetInit() /src/nccl/src/bootstrap.cc:115:13
    #3 0x556781c19907 in initOnceFunc() /src/nccl/src/init.cc:147:3
    #4 0x7f1d7cb404de in __pthread_once_slow (/lib/x86_64-linux-gnu/libpthread.so.0+0x114de) (BuildId: 9753720502573b97dbac595b61fd72c2df18e078)
...
```

## Patch

A patch is provided in an accompanying PR which adds a bounds check before writing into `prefix`. Characters beyond the limit are silently discarded.

## Notes

This bug was found as part of follow-on research from DARPA's AI Cyber Challenge (AIxCC), where Trail of Bits built Buttercup, a Cyber Reasoning System that combines static analysis, fuzzing, and large language models to find and fix vulnerabilities.

### NCCL Version

v2.29

### Your platform details

_No response_

### Error Message & Behavior

_No response_

reytchison:
Fixes: NVIDIA/nccl#2026

## Description

Adds a bounds check in parseStringList to fix a buffer overflow.

## Related Issues

#2026

### DIFF
--- a/src/misc/utils.cc
+++ b/src/misc/utils.cc
@@ -165,8 +165,10 @@ int parseStringList(const char* string, struct netIf* ifList, int maxList) {
         ifNum++; ifC = 0;
       }
     } else {
-      ifList[ifNum].prefix[ifC] = c;
-      ifC++;
+      if (ifC < sizeof(ifList[ifNum].prefix) - 1) {
+        ifList[ifNum].prefix[ifC] = c;
+        ifC++;
+      }
     }
     ptr++;
   } while (ifNum < maxList && c);


xiaofanl-nvidia:
@reytchison your commit has been merged to v2.30u1 here: https://github.com/NVIDIA/nccl/commit/cb04f02320c24ead52261f71a8597db13930afa7

PR #2028, opened 2026-02-27
Fix wrong cumemhandle initialization to nullptr.

larroy:
## Description

While building pytorch I noticed a compile error in the constructor GdakiGlobalGPUBufferTable given that cumemhandle is wrongly initialized as nullptr instead of 0.

## Changes & Impact

Minor fix

## Performance Impact

None


larroy:
Looks like the fix is already in as per @marksantesson

Issue #2029, opened 2026-02-27
[Issue]: gin.putValue() Type Casting Issue in Proxy Mode for Float Types


myh1222:
### How is this issue impacting you?

Data corruption

### Share Your Debug Logs

The `buildGfd()` function in `src/include/nccl_device/gin/proxy/gin_proxy.h` incorrectly handles floating-point types when encoding inline values. The current implementation performs a **numeric conversion** instead of preserving the **bit pattern**, causing `gin.putValue()` to send incorrect data when using `float` or `double` types in Proxy mode.

When `T = float`, the expression `(uint32_t)srcVal` performs a **numeric conversion**:
- `float 100.0f` → `(uint32_t)100.0f` = **100** (decimal value)
- Expected: the IEEE 754 bit pattern **0x42C80000**

This is incorrect because RDMA transfers should preserve the binary representation of the value, not its numeric interpretation.

`template <typename T>
__global__ void TestKernel(ncclWindow_t recvwin, struct ncclDevComm devComm) {
    ncclGin gin{devComm, 0};
    T value = static_cast<T>(100.0);

    gin.putValue(
        ncclTeamWorld(devComm), peer,
        recvwin, 0,
        value,
        ncclGin_SignalInc{0}
    );
}`

### NCCL Version

2.29.2 and higher

### Your platform details

h20

myh1222:
NCCL 2.29.7-1 already fixed the issue, so this commit is no longer needed.

PR #2030, opened 2026-02-27
Fix gin.putValue() fails to send correct floating-point values in Proxy mode

myh1222:
## Summary

The `buildGfd()` function in `src/include/nccl_device/gin/proxy/gin_proxy.h` incorrectly handles floating-point types when encoding inline values. The current implementation performs a **numeric conversion** instead of preserving the **bit pattern**, causing `gin.putValue()` to send incorrect data when using `float` or `double` types in Proxy mode.

## Affected Component

- File: `src/include/nccl_device/gin/proxy/gin_proxy.h`
- Function: `buildGfd()`
- Lines: 74-81 (before fix)

## Root Cause

The problematic code performs direct C-style casts on the template parameter `T srcVal`:
When `T = float`, the expression `(uint32_t)srcVal` performs a **numeric conversion**:
- `float 100.0f` → `(uint32_t)100.0f` = **100** (decimal value)
- Expected: the IEEE 754 bit pattern **0x42C80000**

This is incorrect because RDMA transfers should preserve the binary representation of the value, not its numeric interpretation.


myh1222:
[https://github.com/NVIDIA/nccl/issues/2029](url)

myh1222:
NCCL 2.29.7-1 already fixed the issue, so this commit is no longer needed.

PR #2037, opened 2026-03-09
[Feature Preview] Introduce elastic buffer, gin.get() and other GIN optimizations

xiaofanl-nvidia:
This is an Early Access release. It contains new features and APIs that may be changed before the official General Access release. It also may contain performance degradation that is not representative of the General Access release.

## Elastic Buffer
"Elastic" buffers in this context refers to a contiguous VA space that contains cuMem segments that are backed by either GPU or CPU memory. This is useful for cases where an application runs out of GPU memory and wants to use CPU memory as an extension, all using the same VA for convenience.

## gin.get
Adds non-blocking gin.get to fetch data from a remote peer. Users can ensure the get is complete (and data is visible) using gin.flush

## Other GIN optimizations
- Add a **per-context resource sharing mode** for GIN to control the resource sharing scope used by GDAKI DOCA verbs, defaulting to GPU scope. This enables selecting CTA-scoped sharing when needed while keeping existing behavior unchanged by default.
- New API to **update SL/TC** after devComm created, and new lower level API to reset tracking and memory for verbs.


xiaofanl-nvidia:
Closing this preview since everything is on https://github.com/NVIDIA/nccl/tree/v2.30 already.

PR #2040, opened 2026-03-09
nccl_ep: Low-Latency kernel memory footpring optimization

artpol84:
## Description

Reduce LL kernels memory consumption by including the top-k indices into the token message payloads.

On dispatch, this allows to avoid maintaining a separate buffer space per local-expert/remote-rank pair and instead have one space for each remote rank.
This reduces the memory overhead from O(E x B x H) down to O(N x B x H) where E - number of experts, N - number of ranks, B - batch size, and H - token hidden dimension

On combine, the top-k indices are used to reduce the communication buffer from O(E x B x H) to O(K x B x H).

## Related Issues

N/A

## Changes & Impact

*Changes:* Reorganize NCCL EP communication buffer layout.
*Impact:* Order of magnitude reduction in memory consumption.

## Performance Impact

No impact observed

### DIFF
--- a/contrib/nccl_ep/device/low_latency.cu
+++ b/contrib/nccl_ep/device/low_latency.cu
@@ -31,11 +31,35 @@ __forceinline__ __device__ bool isRankMasked(int* rankMask, int rank) {
 }
 __device__ __constant__ bool dP2pDisabled = false;
 
+__device__ __forceinline__ int getCommId(int hashKey) {
+    return hashKey / MAX_NCCL_GIN_CTX_PER_COMM;
+}
+
+__device__ __forceinline__ int getCtxId(int hashKey) {
+    return hashKey % MAX_NCCL_GIN_CTX_PER_COMM;
+}
+
+__device__ __forceinline__ int getLocalExpertIdx(int expertIdx, int numLocalExperts) {
+    return (expertIdx >= 0) ? expertIdx % numLocalExperts : -1;
+}
+
+__device__ __forceinline__ int getExpertRankIdx(int expertIdx, int numLocalExperts) {
+    return (expertIdx >= 0) ? expertIdx / numLocalExperts : -1;
+}
+
+__device__ __forceinline__ void syncSmGroup(int groupIdx, int nThreads) {
+    asm volatile("bar.sync %0, %1;" :: "r"(groupIdx), "r"(nThreads));
+}
+
+#define SYNC_DISP_SEND_COPY 1
+#define SYNC_DISP_SEND_COMM(idx) (SYNC_DISP_SEND_COPY + idx)
+#define SYNC_DISP_RECV_COPY(topK, idx) (SYNC_DISP_SEND_COMM(topK) + idx)
+
 __forceinline__ __device__ uint64_t ncclGetP2pPtr(const uint64_t& dstPtr,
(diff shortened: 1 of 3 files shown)


xiaofanl-nvidia:
@jskrobola can you help start the mirror?

artpol84:
@xiaofanl-nvidia, @sb17v 

Update: This change was tested with:
* ep_bench microbenchmark, and
* internal vLLM integration

showing no performance degradation compared to pre-optimization code

tingmingzhong:
Hi, I tested this on an H100 and observed a 2x performance regression in low-latency dispatch after communication buffer optimization on intra-node 8-GPU benchmarks.

## Environment
GPU: 8x NVIDIA H100 (Compute Capability 9.0, single node, NVLink enabled)
NCCL Version: 2.29.7+cuda13.1

## Benchmark Configuration

```
Algorithm:  LOW_LATENCY
Ranks:      8
Tokens:     128
Hidden:     7168
Top-k:      8
Experts:    256 (local: 32)
Warmup:     10 iters
Benchmark:  50 iters
Dtype:      BF16
```
## Benchmark Results

| Metric | Baseline  | communication  buffer optimization | Delta |
|--------|---------------------|---------------------------|-------|
| **Dispatch avg** | **67.90 us** | **140.17 us** | **+106%, 2.06x slower** |
| Dispatch throughput | 214.29 GB/s | 103.81 GB/s | -51.6% |
| Combine avg | 104.70 us | 86.87 us | -17.0% (improved) |
| Combine throughput | 138.98 GB/s | 167.50 GB/s | +20.5% |
| **Total (D+C) avg** | **172.61 us** | **227.04 us** | **+31.5%, 1.32x slower** |

## Root Cause Analysis
###  `rankArrivedCnt` global synchronization barrier
All warp groups must wait for **all** `numLocalExperts` (=32) counts from the same source rank to arrive before any can proceed:

```cpp
// After: global barrier per source rank
numRecvTokens = waitForRecvTokensRelaxed(...);
atomic_add_release_global(rankArrivedCnt + srcRank, 1);
while (ld_acquire_sys_global(rankArrivedCnt + srcRank) != numLocalExperts);
```

This serializes the receive pipeline — even if one expert's data is ready, it must wait for all 32 experts' counts to arrive.

artpol84:
Thank you,

Yes this is a known issue on NVLink domain. We have another implementation that is being tested now. It should be available in the next update.
My original comment wasn't precise as I only tested with RDMA domain initially.

My analysis, however, showed that root cause was in a very slow Dispatch receive phase implementation (`findNthToken` function specifically) that has to scan all tokens of a rank and cherry-pick the needed ones for a specific expert:
https://github.com/NVIDIA/nccl/pull/2040/changes#diff-c363ea67a1e5800319d26f21c8fe84ec784a2fc432269e9a36d915041305fcbcR352

Replacing this function with a homogeneous pass over the incoming tokens actually improved Dispatch latency.

artpol84:
Closing this to avoid future confusion

artpol84:
Update:

aparently there new version is available as part of:
https://github.com/NVIDIA/nccl/tree/v2.30u1
branch.

The new version is slightly slower for NVLink-only (not 2x) but shows better results for RDMA domain
I'm still collecting the data

PR #2042, opened 2026-03-09
nccl_ep: Remove unused tensor type enum

artpol84:
Clean the codebase removing the deadcode


contrib/nccl_ep/include/nccl_ep.h
```diff
@@ -32,8 +32,6 @@ typedef enum {
     NCCL_EP_TENSOR_TAG_RECV_EXPERT_COUNTER_DEVICE = 5,
     // Tensor containing tokens received per expert (pinned host memory)
     NCCL_EP_TENSOR_TAG_RECV_EXPERT_COUNTER_HOST = 6,
-    // Tensor containing per-expert token counts
-    NCCL_EP_TENSOR_TAG_TOKENS_PER_EXPERTS = 7,
```
sb17v, line 36: I think we need to fix the following in the dispatch API if we are removing this:
```
//   local_tensors - [IN,OUT] Array of pointers to preallocated tensors, with information that is local to the rank.
//                            LL mode: accepts 1 optional local tensor:
//                                    NUM_TOKENS_PER_EXPERTS: [OUT] a 1D tensor of unsigned int [num_experts]
//                                    that contains the number of tokens received by each expert on this rank.
```

artpol84: This enum is a deadcode, the two enums above duplicate it and are actually used:
https://github.com/NVIDIA/nccl/pull/2042/changes#diff-0ed431b7dc072f7ce49c9470b66954b5beba57b7fdf066e641d0b9867012fe0eL32

artpol84: No API changes required. This PR is a code cleanup only

Issue #1916, opened 2025-11-21
[RFE]: Add ncclNotifyTag API for user-defined profiler metadata injection


dmitry-monakhov:
### Please provide the below details to ensure we understand your needs

# Feature Request: User Metadata Tagging API for NCCL Profiler

## Summary

Add `ncclNotifyTag(uint64_t tag, ncclComm_t comm, cudaStream_t stream)` to inject user-defined metadata into profiler event stream, enabling phase-aware profiling and debug instrumentation with <0.001% overhead.

## Motivation

Modern ML training frameworks divide iterations into distinct phases (forward pass, backward pass, optimizer step), each with different communication patterns and performance characteristics. Understanding per-phase performance is critical for optimization, but current NCCL profiler API provides no mechanism for applications to annotate these semantic boundaries.

### The Phase Attribution Problem

NCCL profilers typically use asynchronous event processing: NCCL operations generate events that are queued and processed by a background thread. This creates a timing gap between when operations occur and when they're attributed to training phases.

**Problem**: When the training framework switches phases (e.g., backward -> optimizer), events still in the queue from the previous phase get misattributed to the new phase. This is especially severe for short phases.

**Optimizer phase challenge**: The optimizer step is particularly problematic because:
1. **Very short duration** (10-200ms, vs 1000ms+ for forward/backward)
2. **Communication-intensive** (AllReduce gradients across all ranks)
3. **Most latency-sensitive** (directly impacts training throughput)

With typical profiler queue lag (~100ms of events), optimizer metrics can be 50-90% corrupted with backward-phase operations, making accurate optimizer analysis nearly impossible.

**Current workarounds** have severe limitations:
- **Synchronous queue flush**: 10%+ training slowdown (unacceptable)
- **External correlation (NVTX)**: Cannot correlate with queued events
- **Post-hoc attribution**: Requires complex timestamp correlation, error-prone

### What We Need

A lightweight mechanism for applications to inject phase markers directly into the profiler event stream, ensuring correct attribution with negligible overhead.

## Proposed API

```c
ncclResult_t ncclNotifyTag(
    uint64_t tag,           // User-defined metadata (arbitrary uint64)
    ncclComm_t comm,        // Communicator (may be NCCL_COMM_NULL)
    cudaStream_t stream     // CUDA stream (may be 0)
);
```

Profiler plugin extension:
```c
typedef enum {
    ncclProfilerEvent_NcclOp = 0,
[...]

PR #2045, opened 2026-03-10
Add ncclNotifyTag API for user-defined profiler-v7

dmonakhov:
# Add ncclNotifyTag API for user-defined profiler annotations (v7)

## Motivation

Training frameworks (PyTorch, Megatron, etc.) need a way to correlate NCCL
collective performance with application-level phases -- forward pass, backward
pass, optimizer step, data loading, etc. Today, profiler plugins see a flat
stream of collectives with no application context.

`ncclNotifyTag` lets users inject lightweight string annotations into the
profiler event stream. Plugins can then attribute bandwidth/latency metrics
to specific training phases, enabling:

- Per-phase collective performance dashboards (Prometheus/Grafana)
- Training step boundaries in Chrome traces
- Straggler detection scoped to specific communication patterns

## Original feature request
- https://github.com/NVIDIA/nccl/issues/1916

## API

```c
ncclResult_t ncclNotifyTag(const char* tag, ncclComm_t comm, cudaStream_t stream);
```

- `tag`: null-terminated string, max 32 bytes including terminator (NCCL_TAG_MAX_LEN)
- `comm`: communicator to annotate
- `stream`: unused (reserved for future use)
- Returns `ncclInvalidArgument` if `tag` or `comm` is NULL

The call is synchronous and lightweight -- no CUDA operations, no allocations.

## Usage example

```c
// Annotate training phases for profiler plugins
ncclNotifyTag("forward", comm, stream);
ncclAllReduce(sbuf, rbuf, N, ncclFloat, ncclSum, comm, stream);

ncclNotifyTag("backward", comm, stream);
ncclAllReduce(sbuf, rbuf, N, ncclFloat, ncclSum, comm, stream);
```

### Profiler plugin API v7

Extends v6 with a new event type `ncclProfileUserTag = (1 << 15)` and a new
descriptor field:

```c
struct {
  const char* tag;  // pointer valid only during startEvent/stopEvent
} userTag;
```

NCCL core delivers UserTag as a start/stop event pair. The tag pointer is valid
only for the duration of the call -- plugins must copy the string if they need
to retain it. NCCL core does zero string copying itself.

Backward compatibility: v6 and earlier plugins receive a v7 shim that filters
out UserTag events and clears the UserTag activation bit. No changes needed
in existing plugins.

### Plugin version chain

```
ncclProfiler_v7 (native)
  -> ncclProfiler_v6 (shim: filters UserTag events)
    -> ncclProfiler_v5 (shim: filters UserTag + CE events)
      -> v4 -> v3 -> v2 -> v1
```

### Example plugin (plugins/profiler/example/)

Records UserTag as Chrome trace instant events:
```json
{"name": "UserTag", "cat": "USER_TAG", "ph": "i", "s": "g",
 "args": {"tag": "forward", "rank": 0}}
```

Uses pool-based allocation (default pool size: 8, configurable via
`NCCL_PROFILE_USER_TAG_POOL_SIZE`).

### Inspector plugin (plugins/profiler/inspector/)

Captures the active tag per communicator at collective insertion time (not
completion time) for deterministic attribution. The tag appears in:

- JSON output: `"user_tag": "forward"` in `coll_perf` records
- Prometheus output: `user_tag="forward"` label on metrics

No lock needed for tag capture -- both UserTag and collective startEvent run
on the same NCCL submission thread, serialized per communicator.

## Testing 

Dockerized integration test with 13 test cases (not included in this PR):

- **Example plugin (6 tests)**: epoch tags, rapid fire (pool exhaustion),
  long tag truncation, empty tag, NULL comm, NULL tag
- **Inspector JSON (6 tests)**: same scenarios, validates `user_tag` in JSON
- **Inspector Prometheus (1 test)**: sustained collectives for 35s, validates
  `user_tag` label in .prom output

All tests pass on H200 with CUDA 13.0.

### DIFF
--- a/plugins/profiler/example/event.h
+++ b/plugins/profiler/example/event.h
@@ -302,6 +302,15 @@ struct ceBatch {
   struct ceBatch* pollerNext;  // For poller tracking list
 };
 
+struct userTag {
+  uint64_t type;
+  struct context* ctx;
+  int rank;
+  char tag[NCCL_TAG_MAX_LEN];
+  double startTs;
+  double stopTs;
+};
+
 struct groupApi {
   uint64_t type;
   struct context* ctx;
@@ -392,6 +401,11 @@ struct context {
   int ceBatchPoolBase;
   int ceBatchPoolIndex;
   struct ceBatch* ceBatchPool;
+
+  int userTagPoolSize;
+  int userTagPoolBase;
+  int userTagPoolIndex;
+  struct userTag* userTagPool;
 };
 
 template <typename T>
--- a/plugins/profiler/example/nccl/profiler.h
+++ b/plugins/profiler/example/nccl/profiler.h
@@ -30,6 +30,8 @@ enum {
   ncclProfileCeColl         = (1 << 12), // CE collective operation
   ncclProfileCeSync         = (1 << 13), // CE synchronization operation
   ncclProfileCeBatch        = (1 << 14), // CE batch operation
+  // User-defined tag events (v7)
+  ncclProfileUserTag        = (1 << 15), // User-defined tag annotation
 };
 
 typedef enum {
@@ -84,7 +86,11 @@ typedef ncclProfilerEventState_t ncclProfilerEventState_v3_t;
 typedef ncclProfilerEventState_t ncclProfilerEventState_v4_t;
 typedef ncclProfilerEventState_t ncclProfilerEventState_v5_t;
 typedef ncclProfilerEventState_t ncclProfilerEventState_v6_t;
+typedef ncclProfilerEventState_t ncclProfilerEventState_v7_t;
 
+#define NCCL_TAG_MAX_LEN 32
+
+#include "profiler_v7.h"
 #include "profiler_v6.h"
 #include "profiler_v5.h"
 #include "profiler_v4.h"
@@ -93,10 +99,10 @@ typedef ncclProfilerEventState_t ncclProfilerEventState_v6_t;
 #include "profiler_v1.h"
 #include "profiler_net.h"
 
-// Use v6 as default to support CE events
-// v5 and earlier versions are still supported for backward compatibility
-typedef ncclProfiler_v6_t ncclProfiler_t;
-typedef ncclProfilerEventDescr_v6_t ncclProfilerEventDescr_t;
-typedef ncclProfilerEventStateArgs_v6_t ncclProfilerEventStateArgs_t;
+// Use v7 as default to support UserTag events
+// v6 and earlier versions are still supported for backward compatibility
+typedef ncclProfiler_v7_t ncclProfiler_t;
+typedef ncclProfilerEventDescr_v7_t ncclProfilerEventDescr_t;
+typedef ncclProfilerEventStateArgs_v7_t ncclProfilerEventStateArgs_t;
 
 #endif // end include guard
--- a/plugins/profiler/example/nccl/profiler_v7.h
+++ b/plugins/profiler/example/nccl/profiler_v7.h
@@ -1,0 +1,151 @@
+/*************************************************************************
+ * SPDX-FileCopyrightText: Copyright (c) 2026 Poolside Inc & AFFILIATES
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * See LICENSE.txt for more license information
+ *************************************************************************/
+
+#ifndef PROFILER_V7_H_
+#define PROFILER_V7_H_
+
+#include "profiler_v6.h"
+
+// Extend v6 descriptor with UserTag support
+typedef struct {
+  uint64_t type;                // event type descriptor
+  void* parentObj;              // pointer to the profiler parent object
+  int rank;                     // originating rank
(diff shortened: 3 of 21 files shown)


gcongiu:
Thank @dmonakhov.

It seems to me like the notify API only exists to let the application talk to the profiler through NCCL. The profiler API design has the plugin form to avoid interface proliferation in NCCL. The original idea was that any new interface needed to interact with the profiler could be exposed through the profiler plugin API rather than the NCCL API.

Can't your plugin expose a tag interface directly to the application without going through NCCL? For example, a way for the application to talk to the plugin using an annotation API could be to use NVTX. I think Pytorch Kineto already uses NVTX annotations in NCCL.

gcongiu:
@dmonakhov we are considering a different approach where the collective API calls will be extended with a configuration param (e.g., `ncclAllReduce(..., ncclConfParam_t* param)`). The param can contain the communication phase tag as attribute and correctly propagate it to the profiler without any need for stream synchronization between one phase and the next. This will require you to adopt the new APIs in your code, will this be a problem? Do you have any concerns/comments about this proposal?

dmitry-monakhov:
@gcongiu 
Hm.. Interesting. Please correct me if I am wrong, you want   ncclConfParam_t  act as  some generic container similar to standard go-lang context  or open telemetry baggage.  Both contests proven to works.  If we can attach KV list to any nccl-op, and it will be visiable to profiler it would be awesome. So it seems like a better idea then my original proposal. Because give more flexibility for user to implement tracing at any scale. This likely result a lot of modifications in our code, but it is doable and this is for good reason.

yfguo:
@dmitry-monakhov 
I would like to follow up on the proposed interface design. We are thinking something like this. There can be a `ncclCollConfig_t` struct which contains things that can be set at per-collective call level. If we add the uint64_t field for user tag, we can "tag" a specific collective call.
```cpp
ncclCollConfig_t forward_pass_cfg = NCCL_COLL_CONFIG_INITIALIZER;
ncclCollConfig_t backward_pass_cfg = NCCL_COLL_CONFIG_INITIALIZER;
forward_pass_cfg.userProfilerTag = FORWARD_TAG;
backward_pass_cfg.userProfilerTag = BACKWARD_TAG;

// run allreduce for forward pass
ncclAllReduceConfig(sbuf, rbuf, N, ncclFloat, ncclSum, comm, stream, &forward_pass_cfg);
// same cfg can be reused for other calls
ncclAllGatherConfig(..., &forward_pass_cfg);
// regular collective call (without config) is not tagged.
ncclAllReduce(...); // run as untagged

// run allreduce for backward pass
ncclAllReduceConfig(sbuf, rbuf, N, ncclFloat, ncclSum, comm, stream, &backward_pass_cfg);
```
Do you think this would satisfy your use case?

dmonakhov:
> @dmitry-monakhov I would like to follow up on the proposed interface design. We are thinking something like this. There can be a `ncclCollConfig_t` struct which contains things that can be set at per-collective call level. If we add the uint64_t field for user tag, we can "tag" a specific collective call.
> 
> ```c++
> ncclCollConfig_t forward_pass_cfg = NCCL_COLL_CONFIG_INITIALIZER;
> ncclCollConfig_t backward_pass_cfg = NCCL_COLL_CONFIG_INITIALIZER;
> forward_pass_cfg.userProfilerTag = FORWARD_TAG;
> backward_pass_cfg.userProfilerTag = BACKWARD_TAG;
> 
> // run allreduce for forward pass
> ncclAllReduceConfig(sbuf, rbuf, N, ncclFloat, ncclSum, comm, stream, &forward_pass_cfg);
> // same cfg can be reused for other calls
> ncclAllGatherConfig(..., &forward_pass_cfg);
> // regular collective call (without config) is not tagged.
> ncclAllReduce(...); // run as untagged
> 
> // run allreduce for backward pass
> ncclAllReduceConfig(sbuf, rbuf, N, ncclFloat, ncclSum, comm, stream, &backward_pass_cfg);
> ```
> 
> Do you think this would satisfy your use case?

@yfguo

Ack. It will works,   uint64_t is wide enough to implement any encoding scheme.
Your approach indeed cleaner, because of explicit parameter is passed. 
From integration point it require some ground work, as majority of users use NCCL indirectly via pytorch nowadays :). It seems it is reasonably straight forward.  We can use existing thread-local trick like https://github.com/pytorch/pytorch/blob/main/c10/util/ThreadLocalDebugInfo.h#L61 to set UserProfilerTagDebugInfo , and all later ill be used during collective calls

yfguo:
@dmitry-monakhov The user profiler tag changes have landed in the `dev` branch as these commits:
69e7c2c0c06f7744d7bde0c24e566a2b2ec84b99
0dab26429c5a0ef15339fe0285809721b67302bf
deb1f27a074afb3eb65106419168814655624563

Please give it a try.

xiaofanl-nvidia:
This will be released as part of 2.31 shortly. Closing.

@dmonakhov please let us know if you see any issues when using this feature! Thank you for your RFE, which motivated us to add the collConfig feature.

PR #2050, opened 2026-03-12
Fix nccl_ep cuda shim to only get used for CUDA version below 13.1

sb17v:
## Description

Fix the preprocessor directive for CUDA compat shims in nccl_ep so shims are enabled for CUDA < 13.1 and disabled for 13.1+


artpol84:
@xiaofanl-nvidia addressed and can be closed

xiaofanl-nvidia:
Merged to master.

Issue #2001, opened 2026-01-26
[Issue]: nccl-tests fail with wrong transfers


angainor:
### How is this issue impacting you?

Data corruption

### Share Your Debug Logs

[ncclDebug.gpu-1-7.192846.txt](https://github.com/user-attachments/files/24858151/ncclDebug.gpu-1-7.192846.txt)
[ncclDebug.gpu-1-7.192847.txt](https://github.com/user-attachments/files/24858150/ncclDebug.gpu-1-7.192847.txt)
[ncclDebug.gpu-1-7.192848.txt](https://github.com/user-attachments/files/24858147/ncclDebug.gpu-1-7.192848.txt)
[ncclDebug.gpu-1-7.192849.txt](https://github.com/user-attachments/files/24858149/ncclDebug.gpu-1-7.192849.txt)
[ncclSystem.txt](https://github.com/user-attachments/files/24858148/ncclSystem.txt)

### Steps to Reproduce the Issue

I'm testing `nccl` using `nccl-tests` on a GH200 system with Slingshot 11 and  `aws-ofi-nccl` plugin. I compile the libraries against CUDA 12.6 using Cray Programing Environment. Starting from `nccl-2.27.3` I am seeing occasional `#wrong` transfers reported by `all_reduce_perf`. I did not reproduce those problems with `nccl-2.26.6`. Interestingly, the issues happen _mostly_ for message size `1048576` and around 30% of the time. Most runs finish without errors.

I tried with a number of `libfabric` and `aws-ofi-nccl`, including newest ones available (`libfabric-2.4.0` and `aws-ofi-nccl-1.17.3`). This has no impact. The only trigger I found is switching NCCL version to greater than `2.27.3` (I also tried the newest `nccl-2.29.2` and see the same issue).

### NCCL Version

2.27.3 and higher

### Your platform details

HPE Cray EX system, GH200 nodes with 4 GPUs, Slingshot 11 interconnect between compute nodes, bare metal. Happens when using cross-node transfers. Tested with 1 GPU per node, and 4 GPUs per node.

```
nvidia-smi topo -m
	GPU0	GPU1	GPU2	GPU3	CPU Affinity	NUMA Affinity	GPU NUMA ID
GPU0	 X 	NV6	NV6	NV6	0-71	0		4
GPU1	NV6	 X 	NV6	NV6	72-143	1		12
GPU2	NV6	NV6	 X 	NV6	144-215	2		20
[...]

PR #2053, opened 2026-03-13
graph/tuning: disable LL128 on PATH_P2C inter-node paths (GH200 corruption fix)

ryanhankins:
On GH200 (Grace Hopper), the inter-node NIC->GPU receive path is classified as PATH_P2C (PCIe->NVLink-C2C).  LL128 requires that when the receiver observes the flag at the end of a 128-byte chunk, all preceding bytes in that chunk are already visible in GPU memory.  That ordering guarantee is not met on GH200: the NIC writes directly into GPU HBM via PCIe->NVLink-C2C, and the two 64-byte halves of the 128-byte chunk can become visible to the GPU independently.  If the flag half (last 8 bytes) becomes visible before the data half, the consumer reads stale or incomplete data, resulting in silent corruption.

Previously the condition to enable LL128 on Hopper/Blackwell was:

  typeInter <= PATH_PXN

This included PATH_P2C because numerically P2C < PXN.  The fix changes the condition to:

  typeInter <= PATH_PXB || typeInter == PATH_PXN

which explicitly excludes PATH_P2C.  LL128 remains enabled for NVLink-switch fabrics (PATH_PXN) and closer intra-node paths where the ordering guarantees hold.  LL and Simple are unaffected and were verified clean on GH200.

Verified:
  NCCL_PROTO=Simple  - clean
  NCCL_PROTO=LL      - clean
  NCCL_PROTO=LL128   - corrupt (fixed by this change)
  NCCL_LL128_C2C=0   - clean (env-var workaround, superseded by this fix)

## Description

  graph/tuning: disable LL128 on PATH_P2C inter-node paths (GH200 corruption fix)

  On GH200 (Grace Hopper), the NIC->GPU receive path is classified as PATH_P2C (PCIe->NVLink->C2C). NCCL's LL128 protocol requires that when the receiver sees the flag byte at
  the end of a 128-byte chunk, all preceding data in that chunk is already visible in GPU memory. That guarantee does not hold on GH200: the two 64-byte halves of the chunk
  can become visible independently across the PCIe->C2C boundary, causing silent data corruption.

  Root cause: the previous condition typeInter <= PATH_PXN inadvertently included PATH_P2C (numerically P2C < PXN).

  Fix: change the condition to typeInter <= PATH_PXB || typeInter == PATH_PXN, explicitly excluding PATH_P2C. LL128 stays enabled for NVLink-switch fabrics and tighter intra-node paths. LL and Simple are unaffected.

  Verified clean on GH200: with SIMPLE,LL,LL128 protocols.

Reference any related issues or PRs:

https://github.com/NVIDIA/nccl/issues/2001

## Performance Impact

N/A

### DIFF
--- a/src/graph/tuning.cc
+++ b/src/graph/tuning.cc
@@ -481,8 +481,15 @@ ncclResult_t ncclTopoTuneModel(struct ncclComm* comm, int minCompCap, int maxCom
     if (pEnable == 2 && p == NCCL_PROTO_LL128) {
       pEnable = 1;
       if (ncclParamLl128C2c() && minCompCap >= 90) {
-        // Enable LL128 by default only on Hopper/Blackwell for all connections up to P2C and PXN.
-        pEnable &= (graphs[a]->typeInter <= PATH_PXN);
+        // Enable LL128 on Hopper/Blackwell for PXN inter-node paths (e.g. NVLink switch
+        // fabrics) where write ordering is guaranteed. Explicitly exclude PATH_P2C even
+        // though P2C < PXN: on GH200 the NIC writes to GPU HBM via PCIe->NVLink-C2C and
+        // PCIe posted writes through the C2C interconnect do not guarantee sequential
+        // visibility at 128-byte granularity, so the LL128 flag (last 8B of a 128B chunk)
+        // can become visible to the GPU before the preceding data bytes, causing corruption.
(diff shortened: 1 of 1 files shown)


ryanhankins:
We have resolved this by disabling the eager protocol, forcing all traffic to go via GPUDirect.

PR #2059, opened 2026-03-20
Fix /dev/shm/nccl-K error.

lunixbochs:
## Description

With memoryless NUMA domains (e.g. if you don't populate all of the memory channels on an EPYC CPU), NCCL IPC has a type confusion that results in this error:
```
[rank1]: Error while attaching to shared memory segment /dev/shm/nccl-Ќ (size 0), error: No such file or directory (2)
```

This happens with memoryless NUMA domains, which would cause the received cuMem SHM descriptor union to be interpreted as a /dev/shm file path. This PR checks the union tag and turns it into a clear warning.

## Related Issues

I have a more detailed writeup here: https://github.com/pytorch/pytorch/issues/152302#issuecomment-4094534698

## Changes & Impact

N/A

## Performance Impact

N/A


sjeaugey:
Thanks for the report. I have the feeling that we could have fixed the issue though, rather than failing.
@xiaofanl-nvidia I think we should investigate further.

PR #2061, opened 2026-03-22
Enable strict C++ compiler warnings and -Werror by default

maxwbuckley:
## Summary
- Enable `-Wextra`, `-Wshadow=local`, `-Wnon-virtual-dtor`, `-Woverloaded-virtual`, `-Wimplicit-fallthrough` in the base CXXFLAGS for both Makefile and CMake build systems
- Enable `-Werror` by default (`WERROR=1`) so no new warnings can be introduced going forward
- Fix all existing warnings across 79 source files to achieve a zero-warning build

## Warning categories fixed
| Warning | Count | Fix |
|---|---|---|
| `-Wmissing-field-initializers` | ~343 | `= {0}` → `= {}`; complete partial aggregate inits |
| `-Wunused-parameter` | ~180 | Comment out names (`type /*name*/`); `(void)param` for conditionally-used params |
| `-Wshadow=local` | ~34 | Rename inner variables; prefix macro-internal vars with `_` |
| `-Wunused-function` | ~10 | Add `inline` to static header functions |
| `-Wtype-limits` | 5 | Remove always-false `unsigned < 0` comparisons |
| `-Wempty-body` | 4 | Add braces around conditionally-empty `TRACE()` macro bodies |
| `-Wimplicit-fallthrough` | 2 | Add `__attribute__((fallthrough))` |
| Third-party NVTX | ~192 | `#pragma GCC system_header` on vendored header |

## Design decisions
- **`-Wshadow=local`** instead of full `-Wshadow`**: Full `-Wshadow` produces ~1700 warnings from constructor parameters shadowing members (idiomatic C++). `-Wshadow=local` catches the genuinely dangerous local-shadows-local cases only.
- **`WERROR=0` escape hatch**: Developers can still build with `make WERROR=0` during prototyping if needed.
- **Vendored NVTX headers**: Suppressed via `#pragma GCC system_header` since modifying third-party code is impractical.

## Test plan
- [x] Full clean build with CUDA 12.9 on RTX 5090 — zero C++ warnings, library links successfully
- [x] `astyle` formatter run on all modified files
- [x] Tuner plugin unit tests: 15/15 passed
- [x] All 10 example programs in `docs/examples/` build cleanly under strict flags

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ option(DEBUG "Enable debug build" OFF)
 option(ASAN "Enable Address Sanitizer" OFF)
 option(UBSAN "Enable Undefined Behavior Sanitizer" OFF)
 option(TRACE "Enable tracing" OFF)
-option(WERROR "Treat warnings as errors" OFF)
+option(WERROR "Treat warnings as errors" ON)
 option(PROFAPI "Enable profiling API" ON)
 option(NVTX "Enable NVTX" ON)
 option(RDMA_CORE "Enable RDMA core" OFF)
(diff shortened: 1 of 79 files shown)


xiaofanl-nvidia:
Hi @maxwbuckley thank you for your contribution. However, this is a sweeping change on a fast moving code base so we will need to do this in a phase manner and interleave with other feature development efforts. 

I'm opening up an internal RFE to track the general suggestion to fix all compile time warnings and adding it to our roadmap.

PR #2067, opened 2026-03-24
[NCCL EP] Fix HT dispatch bandwidth metrics and add IB bandwidth reporting in `ep_bench.cu`

phu0ngng:
## Description

Fix the HT mode dispatch bandwidth metric from bidirectional (`rdma_send + total_recv`) to unidirectional receive (`total_recv / time`), labeled "Recv BW (RDMA+NVL)". The receive-side metric directly answers
  when the expert GEMM can start — it is the number that determines training step latency. The previous bidirectional sum inflated reported numbers by ~2× relative to HybridEP's published figures.

Also fix the IB bandwidth calculation to count remote-node tokens only (excluding local-node traffic that goes via NVLink), and add it as a separate "IB BW" output line for multi-node runs. This matches
  `test_hybrid_ep.py`'s NVL/IB split and enables direct apple-to-apple comparison between NCCL EP and HybridEP benchmarks.

## Related Issues

NA

## Changes & Impact

Apple to Apple comparison to HybridEP

## Performance Impact

NA

### DIFF
--- a/contrib/nccl_ep/ep_bench.cu
+++ b/contrib/nccl_ep/ep_bench.cu
@@ -930,21 +930,27 @@ HighThroughputBytes calculateHighThroughputBytes(
         }
     }
 
-    // Recv side: simulate each remote rank's routing to count tokens that hit our local experts.
-    // Uses the same randperm seed (rank + 42) as the main topk generation loop.
+    // Recv side: simulate ALL source ranks' routing to count unique (src_rank, token) pairs
+    // where at least one of the k selected experts belongs to myRank.
+    // This matches HybridEP's dispatched_hidden.numel() * 2 methodology exactly.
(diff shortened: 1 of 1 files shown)


artpol84:
Thanks @phu0ngng!

Let's take it internally, we currently have some of these fixes in place in the internal repo.

xiaofanl-nvidia:
@phu0ngng is this PR fully contained within this commit at this point? https://github.com/NVIDIA/nccl/commit/292ddf006551247dea9221c2160f8122273b7a48
If there are additional improvements and fixes, I can mirror your github MR to internal code base. This way you don't need to maintain both copies.

phu0ngng:
@xiaofanl-nvidia, the changes in this PR were included upstream already. I will close this PR.

Issue #1744, opened 2025-06-16
Support Building NCCL with CMake?


BwL1289:
We forked [NCCL](https://github.com/eugo-inc/nccl-cmake) to support CMake and Clang-only builds. 

Is there interest in getting this merged upstream?

CC @kiskra-nvidia

BwL1289:
@kiskra-nvidia bumping this. Is there interest?

nv-udeodhar:
@BwL1289 We have been investigating CMake enablement for the next release (v2.28). We will compare this PR to our internal implementation.

BwL1289:
Thanks. Is there a notion that clang will be supported?

BwL1289:
@nv-udeodhar it looks like v2.28 has been released in preview. Was our code used in the CMake implementation?

BwL1289:
@nv-udeodhar just a heads up, your CMake implementation does not work. We've decided to maintain our fork and have confirmed (clang only) that it works at runtime. Thanks.

nv-udeodhar:
Can you please attach the CMake build log?
Can you please use this command line on the latest Github codebase.
`rm -rf build && cmake -DCMAKE_CUDA_ARCHITECTURES=80 -B build/ && cmake --build build --
parallel`
Please note the top level CMake file was recently added on the master branch.

BwL1289:
> Can you please attach the CMake build log?

I no longer have the build log

> Please note the top level CMake file was recently added on the master branch.

We built from this commit.

> Can you please use this command line on the latest Github codebase.
rm -rf build && cmake -DCMAKE_CUDA_ARCHITECTURES=80 -B build/ && cmake --build build -- parallel

We spent the last week updating our CMake fork so we won't be using NCCL's.

PR #2075, opened 2026-03-29
Fix atomic alignment

372046933:
## Description

Below is the smallest reproducible code. the `__attribute__((packed))` on `ncclGinProxyGfd_t` makes the alignment of ncclGinProxyGfd_t equals to 1. The packed struct here is not needed for two reasons.
1. We will do atomic operations on the address later. packed breaks the alignment requirement of atomics
2. `ncclGinProxyQword_t` is already packed and size of it is 8 bytes. So size of `ncclGinProxyGfd_t` is 64 bytes. and `packed` is redundant here.

```c++
#include <atomic>

typedef union {
  uint64_t raw;
  struct {
    uint64_t v : 1;
    uint64_t resv : 63;
  } __attribute__((packed)) flag;
} ncclGinProxyQword_t;


static_assert(sizeof(ncclGinProxyQword_t) == sizeof(uint64_t),
              "sizeof(ncclGinProxyQword_t) != sizeof(uint64_t)");

typedef struct __attribute__((packed)) {
  ncclGinProxyQword_t qword[8];
} ncclGinProxyGfd_t;

static_assert(alignof(ncclGinProxyGfd_t) == 1,
              "Incorrect packed implementation");

void f( ncclGinProxyGfd_t *q) {
  ncclGinProxyQword_t qword;
  __atomic_load(&q[0].qword[0].raw, &qword.raw, __ATOMIC_RELAXED);
}
```

gcc does not give any warning. But clang does.
```
test_atomic.cc:24:3: warning: misaligned atomic operation may incur significant performance penalty; the expected alignment (8 bytes) exceeds the actual alignment (1 bytes) [-Watomic-alignment]
   24 |   __atomic_load(&q[0].qword[0].raw, &qword.raw, __ATOMIC_RELAXED);
      |   ^
1 warning generated.

```

## Related Issues

If we are going to support clang, this packed must be removed, otherwise `libatomic` needs to be linked.
https://github.com/NVIDIA/nccl/issues/1744
## Changes & Impact

This PR does not affect GCC. For clang, the generated code does not need link to `libatomic`

## Performance Impact
No difference for GCC. Significant performance improvement on clang since libatomic call is eliminated

### DIFF
--- a/src/include/nccl_device/gin/proxy/gin_proxy_device_host_common.h
+++ b/src/include/nccl_device/gin/proxy/gin_proxy_device_host_common.h
@@ -118,7 +118,7 @@ typedef enum {
   ncclGinProxyGfdQwords = 8,
 } ncclGinProxyGfdQwordIdx_t;
 
-typedef struct __attribute__((packed)) {
+typedef struct {
   ncclGinProxyQword_t qword[ncclGinProxyGfdQwords];
 } ncclGinProxyGfd_t;
 static_assert(sizeof(ncclGinProxyGfd_t) == 64,


372046933:
Or
```c++
typedef struct alignas(8) __attribute__((packed)) {
  ncclGinProxyQword_t qword[8];
} ncclGinProxyGfd_t;
```
IMHO, removing these attribute makes it simple

372046933:
@dmvevents We are using clang to compile NCCL. Can this PR be merged?

PR #2081, opened 2026-03-31
Add bilingual architecture learning guides

tianhao909:
## Description

Add a bilingual, source-oriented documentation track for NCCL internals under
`docs/`, covering communicator initialization, topology search, tuning,
transport selection, collective execution, and performance-model intuition.

## Related Issues

None.

## Changes & Impact

- add `docs/README.md` as a repository-local docs entry point
- add `docs/en/` deep-dive guides for quick start, architecture, collective
  execution, topology/tuning, math/performance, and source mapping
- add `docs/zh/` mirrored guides with the same structure for Chinese readers
- add a root `README.md` link to the new learning guides
- docs-only change; no API or runtime behavior changes

## Performance Impact

Docs-only change. Validated with `git diff --check` and `make -j src.build`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,12 @@ NCCL (pronounced "Nickel") is a stand-alone library of standard communication ro
 
 For more information on NCCL usage, please refer to the [NCCL documentation](https://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/index.html).
 
+## Repository-local Learning Guides
+
+For source-oriented walkthroughs maintained in this repository, see
+[`docs/README.md`](docs/README.md). For runnable examples, see
+[`docs/examples/README.md`](docs/examples/README.md).
+
 ## Build
 
 Note: the official and tested builds of NCCL can be downloaded from: https://developer.nvidia.com/nccl. You can skip the following build steps if you choose to use the official builds.
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,39 @@
+<!--
+  SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+
+  See LICENSE.txt for more license information
+-->
+
+# NCCL Repository Learning Guides
+
+This directory complements the official NVIDIA documentation with
+source-oriented, repository-local walkthroughs.
+
+## Quick links
+
+- English deep dive: [docs/en/README.md](en/README.md)
+- Runnable examples: [docs/examples/README.md](examples/README.md)
+
+```mermaid
+flowchart TD
+    Docs[docs/] --> EN[en/: source-driven architecture guides]
+    Docs --> EX[examples/: runnable sample programs]
+```
+
+The English track is optimized for readers who want to answer questions such as
+"why did NCCL pick ring instead of tree?" or "which file should I open after
+`ncclAllReduce`?". A mirrored Chinese track will follow in a second commit on the
+same branch so the final PR becomes bilingual.
+
+## What is different from the official documentation?
+
+The official user guide explains how to use NCCL as a library. This repository
+track explains how the current source tree is organized and how the runtime
+actually arrives at its decisions.
+
+## Suggested starting point
+
+1. Read the [English landing page](en/README.md).
+2. Continue with [quick-start.md](en/quick-start.md).
+3. Keep [examples](examples/README.md) open for runnable reference code.
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,6 +19,7 @@ source-oriented, repository-local walkthroughs.
 ```mermaid
 flowchart TD
     Docs[docs/] --> EN[en/: source-driven architecture guides]
+    Docs --> ZH[zh/: mirrored Chinese deep dives]
     Docs --> EX[examples/: runnable sample programs]
 ```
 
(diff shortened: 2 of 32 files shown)


dssgabriel:
If there is interest in this PR, I would be willing to write a French translation.

sjeaugey:
Really interesting work. I read quickly through it and the explanations seemed correct. There are approximations, but that's the point, it's not a 100% precise documentation, the goal is to understand the big picture.

I would not put that documentation at the root of the doc/ directory though. The reason why we created the doc/ folder was to be able to have many sub-directories for various types of documentation.

But I agree that we should create a doc/README.md which would point to all the available documentation (user guide, examples, implementation).

Note that the main issue with implementation documentation is that it requires heavy maintenance. The code evolves rapidly, and documentation can quickly become outdated. So it would be good to mention at the beginning of the documentation that this was based on NCCL version X.

Also, this is a lot of material so I'm afraid it would take us time to review everything in details and check for accuracy everywhere, adding more if needed.

Yet, I feel this would be very useful for many developers.

So, to accelerate the inclusion of this doc, while making it clear this is not coming from NVIDIA but added as a useful guide, would you be ok moving it to e.g. `doc/contrib/tianhao909` (or any directory name in `doc/contrib/`), then have doc/README.md point to it directly, so that users looking for an implementation documentation would find it immediately?

xiaofanl-nvidia:
Please don't close this. We are discussing where to host this on github such as docs/contrib, align with other externally contributed docs.

marksantesson:
Hi @tianhao909 , thanks for your patience! We've just [released](https://github.com/NVIDIA/nccl/commit/3ee68fbf777cb7446277645d3d3a143c5ade3404) to the [v2.30u1 branch](https://github.com/NVIDIA/nccl/commits/v2.30u1) some new directories for hosting material just like this. Can you please modify your commit to conform to the new structure? Please put these files in `docs/contrib/architecture/learning`, with an appropriate link in `docs/contrib/README.md`. I also have some suggestions from reading through that you may or may not want to implement (I only read the english version):

* Please take credit for your good work. You can list yourself as anonymous if you like, but it would be nice to have some indication that this was not the NCCL team. We want to highlight community contributions and make sure we aren't given the credit for others' hard work.
* The README.md says "code that exists in this branch". I think it would be better to say "code that exists in this _library_", unless I'm misunderstanding your point.
* In both architecture.md and topology-and-tuning.md you mention that it "asks each transport" whether it can connect. It does not ask the profiler transport (see mentions of `NTRANSPORTS`, and the collNet transport always says "no", so it might be worth clarifying that. On the other hand, I can understand not wanting to go into that level of detail for a high level architecture overview.
* Also, the nearby "selectTransport" diagram is broken in both pages as I view it. Apparently this is caused by the `Select[selectTransport(...)]` part, where `(...)` has special meaning. Putting it in quotes should fix it: `Select[“selectTransport(...)”]`
* I think it would be helpful if (when you discuss different aspects of init such as topology, paths, and connections) you would mention which modes of NCCL_DEBUG_SUBSYS would print them out.

That's it! If you care to make those changes then we should be able to take it right in. I'd also be happy to make them myself, but I can't adjust the chinese portion to match... Thanks again for your contribution!

tianhao909:
Thanks a lot @marksantesson for the detailed review! I've reworked the PR to fit the new structure. Summary of changes:

- **New location**: moved everything to `docs/contrib/architecture/learning/` (with `en/` and `zh/`), and added a link under a new "Architecture and Internals" section in `docs/contrib/README.md`. Per `docs/contrib/CONTRIBUTING.md` I renamed the per-folder `README.md` index files to `index.md` and dropped the previous root `README.md` addition. The branch is now rebased onto `v2.30u1`.
- **Credit / not NVIDIA**: every page and the contrib entry now state this is a community contribution by tianhao909, not from the NCCL team. The commit is signed off accordingly.
- **"this branch" -> "this library"**: fixed in the index pages (and swept the whole set).
- **"asks each transport" clarification**: clarified that `selectTransport(...)` only iterates the `NTRANSPORTS` real transports and does not ask the profiler transport, and that the CollNet transport's `canConnect` always returns "no" for a p2p edge (CollNet is set up via a separate path), so the practical winners are P2P/SHM/NET. Applied in both `architecture.md` and `topology-and-tuning.md`.
- **Broken selectTransport diagram**: fixed by quoting the node labels, e.g. `Select["selectTransport(...)"]`, in both pages.
- **NCCL_DEBUG_SUBSYS**: added notes/tables mapping init aspects to subsystems (topology/paths/search -> `GRAPH`, tuning -> `TUNING`, connection/transport -> `INIT` with `NET`/`P2P`/`SHM`, bootstrap -> `BOOTSTRAP`, NVLS -> `NVLS`), all with `NCCL_DEBUG=INFO`.
- **Version**: all pages now state they are based on NCCL v2.30u1 (2.30.6).

The Chinese pages were updated to match. Please take another look, and thanks again!

tianhao909:
Thanks @marksantesson and @NCCL-Bot! The mirror failure was about DCO + rebase, so I've fixed both:

- **Rebased onto the latest `v2.30u1`.** The branch was behind the target; it is now rebased on top of the current `v2.30u1` tip with no conflicts.
- **DCO sign-off.** The commit kept `Signed-off-by: tianhao909 <tianhaofu@foxmail.com>`, but it also carried a `Co-authored-by: Copilot <...>` trailer that cannot provide its own sign-off. I removed that co-author trailer so every author on the commit is properly signed off.
- **Docs hardening (no content meaning changed).** While here I made the remaining Mermaid diagrams robust on GitHub: multi-line node labels in `architecture.md` and `topology-and-tuning.md` (EN/ZH) are now quoted with `<br/>` line breaks, matching the earlier `selectTransport("...")` fix.

The branch now has a single, signed-off commit on top of the latest `v2.30u1`. Could you please re-run `/mirror dev` when you have a moment? Thanks again for the help!

xiaofanl-nvidia:
We merged the contribution here: https://github.com/NVIDIA/nccl/commit/86fa684d3bd5b6ac3414d6d024fdc6f65441cb81
Thank you!!

PR #2093, opened 2026-04-06
GIN: implement reset-without-zeroing for signals and counters

bhasunit:
## Description

Replace signal/counter reset-by-zeroing in GIN with an offset-based scheme.
Instead of writing zero to the underlying atomic, resetSignal and
resetCounter snapshot the current value into a per-context offset
array. Subsequent reads and waits subtract the offset, making the
signal/counter appear reset without modifying the NIC-visible value.

## Related Issues

None

## Performance Impact

Improved performance on p5en with GIN.


kgioioso:
To me there are a few design options:

1. Remove `ResetSignal` from the backend API. All backends share the same reset implementation, which caches the current value of the pointer (read via `GetSignalPtr`) as an offset
2. Add `ReadSignal` to the backend API (currently, we implement `ReadSignal` with `GetSignalPtr`). All backends must fully own signals. Backends can choose whether they want to reset the signal or cache an offset
3. Add offset logic to GIN, but keep `ResetSignal` as part of the backend API. That is what this MR is doing.

To me, option 1 is the cleanest, assuming all backends want to use the offset logic for reset. 2 seems slightly cleaner if we want backends to have different logic, but I'm not sure that's needed. IMO option 3 splits the logic up in uninituitive ways.

@bhasunit can you opine?  I'd also like to hear from @jbachan

kgioioso:
Separately, I have concerns that we are changing the public API. It seems like we don't need to change the public API for index-based signals because we store the offsets in the GIN context.

For VA signals, I don't think it's fair to ask users to keep track of the offsets. We should ask backends to implement a proper reset for these types of signals. @bhasunit my understanding is this is possible for you all to do, but will impact performance. Is my understanding correct?

jbachan:
2. Is my favorite. Backends that can reset shouldn't be burdened with an extra load. It still invalidates GetSignalPtr though. I think we would change GetSignalPtr to return both a ptr and a bias value, so that when dereferencing the ptr into a value the value should not be interpreted until having the bias subtracted off first. The reset-capable backends would return the constant 0 for the bias leading to zero additional SASS.

kgioioso:
@jbachan IMO ReadSignal would replace GetSignalPtr. What is the use case of GetSignalPtr if we have ReadSignal?

jbachan:
ReadSignal implies finding the pointer each call. Locating that pointer might involve a few hops through internal metadata structs. At least that was the case with GDAKI last I checked. GetSignalPtr allows locating the ptr to be done once and then reused (e.g. in a waiting spin loop). Prove to me that overheads of ReadSignal would be hoisted out of the spin loop and I have no argument against the simplicity of ReadSignal.

kgioioso:
I think we have different ideas for option 2:

> Add ReadSignal to the backend API (currently, we implement ReadSignal with GetSignalPtr). All backends must fully own signals. Backends can choose whether they want to reset the signal or cache an offset

I mean this to be `gin.waitSignal` and `gin.readSignal` etc with `ncclGinApi_ReadSignal`. 

I think you are proposing implementing `gin.waitSignal` with a new `ncclGinApi_GetSignalPtr`, which also returns a (possibly 0) offset. I like that option. In that case, I don't see much point to adding `ncclGinApi_ReadSignal`; `gin.readSignal` can still be implemented with the new `ncclGinApi_GetSignalPtr`

kgioioso:
@bhasunit and I synced offline. The VA Signal issue remains open.

@bhasunit to put everything in writing:
* For indexed signals, we'll change `ncclGinApi_GetSignalPtr` to return a ptr and an offset. For GDAKI, that offset will always be 0
* Since only CPU Proxy needs it, the offset arrays will be allocated by the backend, not by NCCL. For CPU Proxy, I believe that logic will go in `src/transport/net_ib/gin.cc`
* We'll wait to hear from @jbachan to decide whether ReadSignal is needed.

bhasunit:
Ended up with too many merge conflicts trying to rebase to 2.30u1. Will close this request and open another one against 2.30u1.

PR #2121, opened 2026-04-20
Fix uninitialized comm->netDeviceType leading to CUDA illegal memory access

david-abr:
## Description

178b6b759074597777ce13438efb0e0ba625e429 added netDeviceType and initialized it in src/graph/topo.cc: https://github.com/NVIDIA/nccl/blob/178b6b759074597777ce13438efb0e0ba625e429/src/graph/topo.cc#L780

6aae37927840b2bd7b7d42d2f0050e75f88ee97f refactored the topo.cc loop but dropped the assignment of netDeviceType: https://github.com/NVIDIA/nccl/blob/6aae37927840b2bd7b7d42d2f0050e75f88ee97f/src/graph/topo.cc#L1151

This means that when using a net plugin that reports NCCL_NET_DEVICE_UNPACK and registering user buffers, comm->netDeviceType still reads as NCCL_NET_DEVICE_HOST, so the guard in sendrecv_reg.cc lets registration through, loadRecvConn sets both NetDeviceUnpack and NetRegMode, waitPeer() nulls the src pointer, and ncclNetDeviceUnpack dereferences NULL, causing CUDA illegal memory access.

### DIFF
--- a/src/plugin/net.cc
+++ b/src/plugin/net.cc
@@ -214,6 +214,10 @@ static ncclResult_t ncclNetPluginAssignToComm(struct ncclComm* comm, int pluginI
     comm->ncclNet = netPluginLibs[pluginIndex].ncclNet;
     comm->ncclNetVer = netPluginLibs[pluginIndex].ncclNetVer;
     comm->netPluginIndex = pluginIndex;
+    ncclNetProperties_t props;
+    if (comm->ncclNet->getProperties(0, &props) == ncclSuccess) {
+      comm->netDeviceType = props.netDeviceType;
+    }
     netPluginLibs[pluginIndex].ncclNetPluginRefCount++;
     *isAssigned = true;
     INFO(NCCL_INIT|NCCL_NET, "Assigned NET plugin %s to comm", netPluginLibs[pluginIndex].ncclNet->name);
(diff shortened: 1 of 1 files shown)


david-abr:
@sjeaugey this is my first contribution to NCCL so let me know if I'm doing something wrong here -- would you be willing to review this, or help me find someone who can?

sjeaugey:
Looks reasonable to me. CC @xiaofanl-nvidia

xiaofanl-nvidia:
Working on review & merge now..

src/device/prims_simple.h
```diff
@@ -525,7 +525,7 @@ class Primitives<
             flags |= DirectRead;
           }
         }
-        if (netRegFlag) {
+        if (netRegFlag && !(flags & NetDeviceUnpack)) {
```
KaimingOuyang: This line is not necessary. There is a check in https://github.com/NVIDIA/nccl/blob/6da422082f910a8dd230f7e42e26ece4dc37bccc/src/register/coll_reg.cc#L348

Can we delete this change?

david-abr: yep done

xiaofanl-nvidia:
@david-abr this should be ready to merge, except one of the commits is missing DCO. Could you help squash the two commits into one? Thanks!

david-abr:
@xiaofanl-nvidia sorry about the latency here. Yep I just rebased the two commits together.

david-abr:
@xiaofanl-nvidia bump here?

david-abr:
@KaimingOuyang @xiaofanl-nvidia one more bump?

david-abr:
@sjeaugey @KaimingOuyang @xiaofanl-nvidia one more bump attempt? Let me know if I should close the PR

sjeaugey:
Sorry about the delay. This is a genuine bug so we should fix it.
CC @xiaofanl-nvidia

PR #2123, opened 2026-04-21
fix num_recv_tokens calculate error in nccl ep bench

yilin-void:
## Description
When I was running ep_bench in HT mode, I encountered an “invalid argument” error, and it looks like this code error is what’s causing the problem.
This must match the allocation formula in nccl_ep.cc:998 `ep_group->max_recv_tokens = nRanks * max_tokens_per_rank`
Using num_local_experts here (instead of nRanks) overshoots when num_experts > nRanks^2, causing cudaMemcpyAsync at nccl_ep.cc:2090 to read past the IPC buffer (invalid argument).

### DIFF
--- a/contrib/nccl_ep/ep_bench.cu
+++ b/contrib/nccl_ep/ep_bench.cu
@@ -1982,7 +1982,7 @@ int main(int argc, char* argv[]) {
             fflush(stdout);
         }
     } else {
-        num_recv_tokens = config.max_tokens_per_rank * num_local_experts;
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
@yilin-void Thanks for catching this! But I think we recently fixed this on ToT. 
This seems to be the commit: https://github.com/NVIDIA/nccl/commit/2679c654fd56c8c5bbf2880cd6a31132e41018c1
You can monitor our active development on https://github.com/NVIDIA/nccl/tree/v2.30u1
Closing for now.

PR #2080, opened 2026-03-31
Enable P2P transport for AMD systems with >2 GPUs at PHB level

voipmonitor:
## Summary

On AMD multi-socket systems, GPUs on the same NUMA node connect through separate PCIe root complexes under the same PCIe Host Bridge (`PATH_PHB`). The default P2P level (`PATH_PXB`) disables P2P for these paths, forcing NCCL to use shared memory (SHM) transport instead of direct P2P — with **24-42% bandwidth loss**.

The existing AMD exception (`paths.cc:328`) enables `PATH_SYS`-level P2P but only for ≤2 GPUs. This patch extends it to allow `PATH_PHB`-level P2P for >2 GPUs, enabling same-socket P2P while remaining conservative (no cross-socket P2P change).

**Code change** (1 line logic change in `src/graph/paths.cc`):
```c
// Before:
if ((arch == X86 && vendor == AMD) && GPU.count <= 2) p2pLevel = PATH_SYS;

// After:
if (arch == X86 && vendor == AMD) {
    p2pLevel = (GPU.count <= 2) ? PATH_SYS : PATH_PHB;
}
```

## Benchmark results

**System**: Dual-socket AMD EPYC 9575F (Turin, Zen 5), 4x NVIDIA RTX PRO 6000 (Blackwell) on same NUMA node

**Both stock and patched use NCCL 2.29.7+cuda13.2** built from the same master commit (3619159), with the only difference being this patch. nccl-tests rebuilt against 2.29.7.

**Transport change**: `SHM/direct/direct` → `P2P/direct pointer`

### Throughput (all_reduce_perf, Ring, -g 4, -n 500, bus bandwidth in GB/s)

| Size | Stock (SHM) | Patched (P2P) | Improvement |
|------|-------------|---------------|-------------|
| 256K | 11.32 GB/s  | 15.32 GB/s    | **+35%**    |
| 512K | 13.12 GB/s  | 18.55 GB/s    | **+41%**    |
| 1M   | 20.01 GB/s  | 27.88 GB/s    | **+39%**    |
| 2M   | 24.62 GB/s  | 34.89 GB/s    | **+42%**    |
| 4M   | 30.77 GB/s  | 39.67 GB/s    | **+29%**    |
| 16M  | 36.18 GB/s  | 44.92 GB/s    | **+24%**    |
| 128M | 37.64 GB/s  | 46.60 GB/s    | **+24%**    |

### Latency (all_reduce_perf, -g 4, -n 1000)

[...]

PR #2036, opened 2026-03-08
Fix AMD inter-CPU bandwidth and LL128 protocol for multi-socket EPYC

voipmonitor:
## Summary

- **AMD inter-socket bandwidth is hardcoded to 16 GB/s for all generations**, while Intel has 4 model-specific tiers (BDW 6 → ERP 40 GB/s). Modern AMD EPYC (Milan, Genoa, Turin) Infinity Fabric delivers 32-48+ GB/s, causing NCCL to severely underestimate available bandwidth on AMD dual-socket systems.

- **LL128 protocol is unconditionally disabled for PATH_SYS connections.** On AMD multi-socket systems, inter-GPU paths always traverse the SMP interconnect (PATH_SYS), forcing fallback to SIMPLE protocol. The only workaround is `NCCL_GRAPH_FILE` with manually crafted XML overriding path types and speeds.

### 1. AMD CPU generation detection (`graph.h`, `topo.cc`)
Added CPUID Family-based model detection for AMD, mirroring existing Intel logic:

| Generation | CPUID Family | NCCL familyId | Model constant |
|---|---|---|---|
| Zen 1/2 (Naples/Rome) | 17h | 0x8F | `AMD_ZEN12` |
| Zen 3/4 (Milan/Genoa) | 19h | 0xAF | `AMD_ZEN34` |
| Zen 5 (Turin) | 1Ah | 0xBF | `AMD_ZEN5` |

### 2. Generation-specific inter-socket bandwidth (`topo.h`, `topo.cc`)
Replaced flat `AMD_BW = 16.0` with per-generation values:

| Generation | Bandwidth | Rationale |
|---|---|---|
| Zen 1/2 | 16.0 GB/s | Unchanged (2x xGMI 1.0/2.0) |
| Zen 3/4 | 32.0 GB/s | 3-4x xGMI 3.0/4.0 links |
| Zen 5 | 48.0 GB/s | Enhanced Infinity Fabric |

### 3. LL128 over PATH_SYS on AMD (`tuning.cc`)
For Hopper/Blackwell GPUs on AMD systems: enable LL128 when `typeInter == PATH_SYS` and `bwInter >= 32 GB/s` (Zen 3+). This allows low-latency protocols over Infinity Fabric instead of forcing SIMPLE protocol fallback.

## Motivation

On AMD EPYC dual-socket systems with Hopper GPUs, NCCL:
1. Reports only 16 GB/s SYS bandwidth (vs actual 32-48+ GB/s)
[...]

PR #2127, opened 2026-04-22
[Graph] Improve AMD Turin no-XML ring graph selection on AMD Turin

voipmonitor:
## Summary
This draft PR improves no-XML intra-node ring graph selection on dual-socket AMD Turin systems.

It addresses a pathological case where NCCL auto graph selection falls back to a low-bandwidth 8-GPU ring (`Pattern 4`, `bw 15/15`, `type SYS/PIX`, `nChannels 2`) unless `NCCL_GRAPH_FILE` is used.

## Motivation
On an 8x RTX PRO 6000 Blackwell server with dual-socket AMD Turin, the application-level cold-prefill path was much slower without `NCCL_GRAPH_FILE`, even though transport stayed on `P2P/CUMEM`.

The problematic graph shape was reproducible in NCCL debug logs and consistently showed the same low-bandwidth full-8-GPU ring choice.

Environment notes for the reproducer:
- dual-socket AMD Turin
- 8x RTX PRO 6000 Blackwell
- GPU P2P is force-enabled via driver modprobe on this machine
- the app-level reproducer also enables a PCIe custom allreduce path (`VLLM_ENABLE_PCIE_ALLREDUCE=1`)
- `NCCL_P2P_LEVEL=SYS`

## What this changes
This stack does three things:

1. Improve graph comparison/search behavior so that multi-channel candidates are preferred when aggregate bandwidth is equal.
2. Increase the Zen5 intra-node bandwidth model from `32.0` to `40.0`.
3. Add a narrow AMD Turin pathological-ring promotion path for the bad auto-selected 8-GPU intra-node ring.

The promotion path is intentionally gated to a very specific case:
- AMD Zen5
- intra-node ring
- 8 GPUs
- `nChannels == 2`
- `typeIntra == PATH_SYS`
- `typeInter == PATH_PIX`
- `bwIntra <= 15.0`
- `bwInter <= 15.0`

## Reproducer
The application-level reproducer is a single cold prefill request through vLLM using:
- `moonshotai/Kimi-K2.5`
- `lightseekorg/kimi-k2.5-eagle3-mla`
- `TRITON_MLA`
- FP8 KV cache
- `DCP=4`

The measurement is TTFT to the first streamed token on the first request only.

## Results
Fair rerun from the same build environment:
- no-XML with this PR stack: `ttft=1.073736s`, `8005` prompt tokens, `7455.28 tok/s`
- XML baseline: `ttft=1.070525s`, `8005` prompt tokens, `7477.64 tok/s`

These are effectively at parity.

For reference, the broken no-XML path before the fix was much slower on the same reproducer:
- old no-XML: `ttft=9.138378s`, `8005` prompt tokens, `875.98 tok/s`

## Non-goal / additional validation
Forcing `8` channels was tested separately and was not the fix. It produced only noise at the app level and could be worse on synthetic collective microbenchmarks.

## Dependency note
This draft stacks on the existing AMD work:
- #2036
- #2080

Those are still required for the full setup used here.

### DIFF
--- a/src/graph/paths.cc
+++ b/src/graph/paths.cc
@@ -339,8 +339,13 @@ ncclResult_t ncclTopoCheckP2p(struct ncclComm* comm, struct ncclTopoSystem* syst
 
   int arch, vendor, model;
   NCCLCHECK(ncclTopoCpuType(system, &arch, &vendor, &model));
-  // Allow P2P between pairs of GPU devices on AMD systems
-  if ((arch == NCCL_TOPO_CPU_ARCH_X86 && vendor == NCCL_TOPO_CPU_VENDOR_AMD) && system->nodes[DEV].count <= 2) p2pLevel = PATH_SYS;
+  // Allow P2P on AMD systems: SYS level for ≤2 GPU devices (original behavior),
+  // PHB level for >2 to enable same-socket P2P through the PCIe Host Bridge.
+  // Without this, GPUs under separate root complexes on the same NUMA node
+  // (PATH_PHB) fall back to shared memory transport, losing 24-46% bandwidth.
+  if (arch == NCCL_TOPO_CPU_ARCH_X86 && vendor == NCCL_TOPO_CPU_VENDOR_AMD) {
+    p2pLevel = (system->nodes[DEV].count <= 2) ? PATH_SYS : PATH_PHB;
+  }
 
   // User override
   NCCLCHECK(ncclGetUserP2pLevel(&p2pLevel));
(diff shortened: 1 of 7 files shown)


hz-cheng:
Great work! I tested this PR in my environment, and it can solve my problem.

But I found that if I don't limit the maximum number of channels with `NCCL_MAX_NCHANNELS`, it tries to create 40 channels, which seems to be too many.

Another question, Zen5 intra-node bandwidth is at least 192GB/s, Why do you set it to 40.0GB/s instead?

xiaofanl-nvidia:
++ @thomasgillis @marksantesson to review and mirror once it's ready.

marksantesson:
@voipmonitor , can you please rebase this onto current `dev` branch? Some of the changes from your previous change appear duplicated here, so it would be good to see what is left.

Also, I'm concerned that this seems to be a fix targeted to your environment, not a general fix. In particular, I'd rather the algorithm found the right result rather than patching it up at the end. I'm happy to work with you to find out why it isn't getting the right solution normally after you've rebased.

Thanks!

PR #2132, opened 2026-04-24
Align GIN cross-NIC default with classic graph search semantics

ngoyal2707:
## Summary

`ncclTopoCheckCrossNicSupport` currently returns `true` whenever
`NCCL_CROSS_NIC != 0`, which means the default value `2` commits the
GIN init path to `NCCL_GIN_CONNECTION_FULL` (any-to-any across NICs).
Unlike the classic graph search, GIN has no retry/fallback path, so
this choice is load-bearing and not correctable after init.

The classic graph search interprets the same default `NCCL_CROSS_NIC=2`
differently (`src/graph/search.cc:1011-1015, 1157-1163`): it starts with
`graph->crossNic = 0` (rail-aligned) and only retries with cross-NIC if
the rail-aligned search can't satisfy bandwidth/latency. On rail-aligned
fabrics the first attempt succeeds and cross-NIC is never used.

This change aligns GIN's default with that classic behavior by requiring
an explicit opt-in (`NCCL_CROSS_NIC=1`) to advertise cross-NIC support
to GIN. `NCCL_CROSS_NIC=0` and `NCCL_CROSS_NIC=2` both keep GIN in
`CONNECTION_RAIL`.

## Effect

| `NCCL_CROSS_NIC` | Before | After |
|------------------|--------|-------|
| unset (default 2) | `CONNECTION_FULL` | `CONNECTION_RAIL` |
| `=0` | `CONNECTION_RAIL` | `CONNECTION_RAIL` |
| `=1` | `CONNECTION_FULL` | `CONNECTION_FULL` |

Classic graph search behavior is unchanged (it reads `ncclParamCrossNic()`
directly, not through this helper).

### DIFF
--- a/src/graph/search.cc
+++ b/src/graph/search.cc
@@ -998,7 +998,7 @@ float sm100SpeedArrayInter[] = { 96.0, 86.0, 80.0, 48.0, 45.1, 42.0, 40.0, 30.0,
 #define NSPEEDSINTER_SM100 (sizeof(sm100SpeedArrayInter)/sizeof(float))
 
 ncclResult_t ncclTopoCheckCrossNicSupport(bool* supported) {
-  *supported = (ncclParamCrossNic() != 0);
+  *supported = (ncclParamCrossNic() == 1);
   return ncclSuccess;
 }
 


thomasgillis:
@ngoyal2707 thanks for this change. However we do stand by our current implementation choice, and here is why:
per our documentation, NCCL_CROSS_NIC=2 means "Try to use the same NIC for the same ring/tree, but still allow for the use of different NICs if it would result in a better performance.". 

In the context of GIN, this is interpreted as the device API is allowed to request full connectivity to achieve the best performance. Accepting your change would prevent such a device API usage, which would break the compliance to our documentation.

ngoyal2707:
Thanks as discussed offline, I think itd be good to root cause why is NCCL trying cross rail self connection in the first place if kernels are not supposed to do any cross rail comms

sjeaugey:
I think we understand that well. We create resources for each ncclDevCommCreate, but the GIN group is shared between all devComms. So if CROSS_NIC is not set to 0, the GIN group will include all peers, in case a later call to ncclDevCommCreate will ask for a fully connected GIN.

So when we create the GIN resources, we are creating connections based on the GIN group, which has all peers. Which is semantically fine, as GIN communication along rails will still work fine, so we met the requirement (just provided more than requested).

Arguably though, if the requirement was to only connect GIN in rail mode, we are wasting time and resources by providing a fully connect GIN. And we're also exposing ourselves to failure on systems where NCCL_CROSS_NIC does not reflect the actual connectivity.

In any case, we're exploring a couple of options to only connect peers on the same rail when asked for rail-only GIN, to save time, resources, and avoid errors.

ngoyal2707:
Thanks @sjeaugey and @thomasgillis 
makes sense to me

PR #2142, opened 2026-04-28
Replace strerror() with thread-safe ncclStrerror wrapper

randomizedcoder:
## Description

Replace ~48 thread-unsafe `strerror()` calls with a new `ncclStrerror()` inline wrapper that uses `strerror_r()`, handling both GNU and POSIX variants at compile time.

`strerror()` returns a pointer to a shared static buffer. Concurrent calls from NCCL's worker threads can corrupt error messages or cause data races. This is the most widespread MT-unsafe function in the codebase — 46 call sites flagged by static analysis.

### Static analysis findings

| Tool | Rule | Findings |
|------|------|----------|
| **semgrep** | `strerror-thread-unsafe` | 46 instances across codebase |
| **clang-tidy** | `concurrency-mt-unsafe` | strerror included in 836 total MT-unsafe findings |

### What changed

**New wrapper** in `src/include/checks.h`:

```cpp
static inline const char* ncclStrerror(int errnum, char* buf, size_t buflen) {
#if (_POSIX_C_SOURCE >= 200112L) && !defined(_GNU_SOURCE)
  if (strerror_r(errnum, buf, buflen) != 0)
    snprintf(buf, buflen, "Unknown error %d", errnum);
  return buf;
#else
  return strerror_r(errnum, buf, buflen);
#endif
}
```

Each call site adds a stack-local `char errBuf[256]` and passes it to `ncclStrerror()`. No shared state, no behavioral change for single-threaded callers.

**14 files updated** (48 conversions total):

- `src/include/checks.h` — wrapper + 9 error-checking macros
- `src/os/linux.cc` — ~20 conversions in socket/shm functions
- `src/misc/ibvwrap.cc` — 5 instances in IBV macros
- `src/misc/mlx5dvwrap.cc` — 3 instances in MLX5DV macros
- `src/include/ibvwrap.h` — 2 instances in inline functions
- `src/proxy.cc`, `src/os/linux_ipcsocket.cc`, `src/graph/xml.cc`, `src/ras/client.cc`, `src/transport/net_ib/connect.cc`, `src/transport/net_socket.cc`, `src/ras/ras.cc`, `src/plugin/profiler.cc`, `plugins/profiler/inspector/inspector.cc` — 1-3 instances each

### Commits

1. **test:** unit tests with source verification (intentionally FAIL before fix)
2. **fix:** `ncclStrerror` wrapper + all 48 call site conversions (4/4 tests PASS)

### Testing

```bash
cd tests/c && make clean && make test
# 4/4 PASS (known values, zero-buffer edge case, thread safety, source verification)
```

Tests are self-contained (no CUDA/GPU required).

### Note on style

While working on `linux.cc`, we noticed several C-style casts (`(char*)`, `(int*)`) and `NULL` literals that could be modernized to `static_cast`/`reinterpret_cast` and `nullptr`. We kept this PR focused on the thread-safety fix per the contributing guidelines, but are happy to submit a follow-up style PR for those improvements if the team is interested.

### DIFF
--- a/plugins/profiler/inspector/inspector.cc
+++ b/plugins/profiler/inspector/inspector.cc
@@ -434,9 +434,10 @@ static bool ensureDir(char* workdir) {
     if (mkdir(workdir, mode) == 0) {
       return true; // Directory created successfully
     } else {
+      { char errBuf[256];
       INFO_INSPECTOR(
         "NCCL Inspector: failed to create dump directory %s: %s", workdir,
-        strerror(errno));
+        strerror_r(errno, errBuf, sizeof(errBuf))); }
       return false;
     }
   }
@@ -512,8 +513,9 @@ inspectorDumpThread::~inspectorDumpThread() {
           TRACE_INSPECTOR("NCCL Inspector: Cleaned up Prometheus file %s",
                           deviceFlushEntries[i].filename);
         } else {
+          { char errBuf[256];
           INFO_INSPECTOR("NCCL Inspector: Failed to cleanup Prometheus file %s: %s",
-                         deviceFlushEntries[i].filename, strerror(errno));
+                         deviceFlushEntries[i].filename, strerror_r(errno, errBuf, sizeof(errBuf))); }
         }
       }
     }
--- a/src/graph/xml.cc
+++ b/src/graph/xml.cc
@@ -392,7 +392,8 @@ ncclResult_t ncclTopoGetXmlFromFile(const char* xmlTopoFile, struct ncclXml* xml
   FILE* file = fopen(xmlTopoFile, "r");
   if (file == NULL) {
     if (warn) {
-      INFO(NCCL_GRAPH|NCCL_ENV, "Could not open XML topology file %s : %s", xmlTopoFile, strerror(errno));
+      char errBuf[256];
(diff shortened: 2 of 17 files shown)


sjeaugey:
Thanks for the PR. It is true that per its man page, strerror is not thread-safe. In practice, it is thread safe most of the time.

Basically, if the error is known, then strerror returns a pointer to a static array with all error codes, which is not going to be modified (it's read-only data) and therefore it's not going to be corrupted by another call to strerror.

The only reason for strerror to not be thread-safe is when the error is not known, in which case strerror will return "Unknown error XXX", which is generated at call time in a static buffer, which can be overwritten by the next call to strerror, if that next call is also called with an unknown error.

In other words, the only consequence of this issue is for the "Unknown error XXX" error message to be potentially incorrect. Sure, it could happen in case of severe memory corruption, but frankly I'm not sure it justifies the extra code complexity. The main advantage is probably that it silences static analysis tools.

Regarding the note on style, the host-side code of NCCL was historically deliberately written with C-style (while the CUDA device-side code was C++ style) to allow for code sharing with other communication libraries, which are predominantly written in C (MPICH, Open-MPI, UCX, etc). Over time though, it is true that some parts of NCCL were rewritten not following that convention, making it a mix of C and C++ style, which is not ideal.

PR #2143, opened 2026-04-28
fix: replace sprintf/strcpy with bounds-checked snprintf/strncpy

randomizedcoder:
## Issues addressed

**Buffer overflow risk from unbounded string operations:**
- `sprintf()` writes with no size limit — any format expansion exceeding the buffer is undefined behavior
- `strcpy()` copies with no bounds check — source strings longer than the destination overflow silently

These are flagged by flawfinder (CWE-120, CWE-134), cpplint (`runtime/printf`), and gcc `-Wformat-truncation`.

## Changes

**Commit 1 — `test: add snprintf/strncpy unit tests with source verification`**
Tests first (TDD ordering). 25 table-driven test cases covering:
- `snprintf` boundary conditions, return-value overflow, incremental buffer fill, PCI bus ID format
- `strncpy` null-termination guarantees, Unix socket path (108B) and `MAX_STR_LEN` (256B) boundaries
- Source-verification tests that grep changed files for `snprintf`/`strncpy` presence (intentionally FAIL on unfixed code)

**Commit 2 — `fix: replace sprintf/strcpy with bounds-checked snprintf/strncpy`**
Mechanical replacements across 9 files:
| File | Change |
|------|--------|
| `src/debug.cc` | `strcpy` → `strncpy` for timestamp format strings |
| `src/graph/xml.h` | `strcpy` → `strncpy` for XML attribute key/value copying |
| `src/misc/socket.cc` | `sprintf` → `snprintf` for socket address formatting |
| `src/misc/utils.cc` | `sprintf` → `snprintf` for PCI bus ID formatting |
| `src/os/linux_ipcsocket.cc` | `strcpy` → `strncpy` for Unix domain socket paths |
| `src/plugin/gin.cc` | `strcpy` → `strncpy` for GIN plugin name copying |
| `src/plugin/net.cc` | `strcpy` → `strncpy` for net plugin name copying |
| `src/plugin/profiler.cc` | `sprintf` → `snprintf` with `size_t` position counter |
| `src/ras/client.cc` | `strcpy` → `strncpy` for RAS client hostname/port/handshake |

Pattern for all `strncpy` replacements:
```c
strncpy(dst, src, sizeof(dst)-1);
dst[sizeof(dst)-1] = '\0';
```

## Test plan
- [ ] Verify `snprintf` return values are handled correctly (no silent truncation bugs)
- [ ] Verify all `strncpy` calls have explicit null-termination
- [ ] Run `tests/c/test_buffer_overflow` — all 25 cases should PASS
- [ ] Build with `-Wall -Wextra` — no new warnings

### DIFF
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -218,15 +218,19 @@ static void ncclDebugInit() {
         ncclDebugTimestampFormat[i+j] = ' ';
         ncclDebugTimestampMaxSubseconds *= 10;
       }
-      strcpy(ncclDebugTimestampFormat+i+ncclDebugTimestampSubsecondDigits, tsFormat+i+replaceLen);
+      strncpy(ncclDebugTimestampFormat+i+ncclDebugTimestampSubsecondDigits, tsFormat+i+replaceLen,
+              sizeof(ncclDebugTimestampFormat)-i-ncclDebugTimestampSubsecondDigits);
+      ncclDebugTimestampFormat[sizeof(ncclDebugTimestampFormat)-1] = '\0';
       break;
     }
   }
   if (ncclDebugTimestampSubsecondsStart == -1) {
     if (strlen(tsFormat) < sizeof(ncclDebugTimestampFormat)) {
-      strcpy(ncclDebugTimestampFormat, tsFormat);
+      strncpy(ncclDebugTimestampFormat, tsFormat, sizeof(ncclDebugTimestampFormat)-1);
+      ncclDebugTimestampFormat[sizeof(ncclDebugTimestampFormat)-1] = '\0';
     } else {
-      strcpy(ncclDebugTimestampFormat, "[%F %T] ");
+      strncpy(ncclDebugTimestampFormat, "[%F %T] ", sizeof(ncclDebugTimestampFormat)-1);
+      ncclDebugTimestampFormat[sizeof(ncclDebugTimestampFormat)-1] = '\0';
     }
   }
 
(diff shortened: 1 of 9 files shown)


sjeaugey:
I agree with most fixes in `src`.

Regarding the tests though, it would have been good to review Claude's changes before posting the PR. If everyone posts changes directly from Claude without proper review, then the burden is on **us** to sort what's good from what's not good, and you are not adding a lot of value.

randomizedcoder:
G'day,

Thanks for the feedback.

To ensure the code change fixes the problem, I used a TDD approach, so the first commit does detect the issue, and then the second commit fixes it.  The point of the tests is to be table driven, so that as a reviewer you can see all the combinations and permutations are well covered.  Tables mean it is easy to add/remove more tests, and hopefully encourage future maintainers to consider all the cases.  This certainly wasn't a single shot Claude.  Claude is lazy, there's no way it would have created table driven tests without a lot of encouragement.

With that said, happy to remove the tests, or to reduce them.  Please let me know what you would prefer.

Thanks,
Dave

sjeaugey:
Okay. So maybe it's just that I can't make sense of the tests.

It seems to me that many tests are not testing NCCL code at all, they are testing the behavior of snprintf and strncpy. So I'm confused, as those tests should have passed before the change, and they're not really testing the change.

Am I missing the point?

As for tests actually checking the NCCL code, it seems that they're just doing a grep for sprintf and strcpy. Not sure why we needed to write code for that?

randomizedcoder:
Tests removed
https://github.com/NVIDIA/nccl/pull/2143/changes/ac473b9ede24f61bc2d3a38c04521651523c7784

sjeaugey:
Would it be possible to just force-push the branch with only the second commit, to keep the history clean?

randomizedcoder:
Single commit
https://github.com/NVIDIA/nccl/pull/2143/changes/3f95965d15fbb25519cef0fa61ce38396b42a9cd

sjeaugey:
Hum, looks like the DCO check failed. Does your commit have the same `Author` and `Signed-off-by` emails?

randomizedcoder:
Fixed!  Sorry about that!

PR #2161, opened 2026-05-07
nccl_ep: Fix integer overflow in inter-node N2N warp at 8+ nodes

mozarhua:
## Description

Fix integer overflow in inter-node N2N warp at 8+ nodes encountered when running NCCL EP benchmark using high throughput mode. 

The `dense_dst_offset` variable in the inter-node N2N warp of the dispatch kernel was declared as int, causing signed 32-bit overflow when computing RDMA write destination offsets at 8 nodes. The offset calculation `(rank_in_remote * max_tokens_per_dest * bytes_per_entry + chunk_offset * bytes_per_entry)` exceeds `INT32_MAX` (~2.1 GB) with large GIN buffers, producing a negative value that sign-extends to a bogus 64-bit address. Later in `net.put()`, RDMA write failed with this bogus address falling out of the MR region.

## Changes & Impact

This commit changes `dense_dst_offset` from `int` to `size_t`.


xiaofanl-nvidia:
Merged here - https://github.com/NVIDIA/nccl/commit/7c308bfed7d066d46e4c30806c355b2cfe6dadef Closing.

Issue #2165, opened 2026-05-08
[Issue]: GIN may choose wrong NIC when net plugin is used


fishautumn:
### How is this issue impacting you?

Lower performance than expected

### NCCL Version

2.30.4

### Error Message & Behavior

We ran NCCL GIN with nccl-rdma-sharp net plugin, but found it chosed wrong NIC (cross NUMA, and resulting in quite low bandwidth).

Here is the detail:
```
ncclCommInitRankFunc
├── commAlloc
│   ├── ncclNetInit
│   │   └── ncclNetPluginInit
│   │       └── net->init => ncclIbInit (net plugin)
│   │           └── nccl_p2p_ib_init    (save to plugin ncclIbDevs[], original order)
│   └── ncclGinInit
│       └── ncclGinPluginInit
│           └── gin->init => ncclGinIbInit
│               └── ncclGinIbInitType
│                   └── ncclIbInitDevices (nccl internal ncclIbDevs[], PCI order)
└── initTransportsRank
    └── ncclTopoGetSystem
        ├── ncclTopoProcessNet(gin)
        |   └── ncclTopoPopulateNics
        |       └── gin->getProperties    (nccl internal ncclIbDevs[], PCI order)
        └── ncclTopoProcessNet(net)
            └── ncclTopoPopulateNics
                └── net->getProperties    (plugin ncclIbDevs[], original order)
```

plugin and gin used the same NIC set, but with different order.
As `$NCCL_IB_DEVICE_PCI_ORDER` is on by default, gin sorted the NIC list by PCI address, while the net plugin doesn't.

Bellow logs of 2nd `ncclTopoPopulateNics` indicates the differences:
```
2026-05-07 19:22:05 slave34:6072:6072 [0] NCCL INFO TOPO/NET : Changing NCCL RDMA Plugin v11 dev index from 4 to 3
2026-05-07 19:22:05 slave34:6072:6072 [0] NCCL INFO TOPO/NET : Changing NCCL RDMA Plugin v11 dev index from 3 to 4
```

After the 2nd  `ncclTopoPopulateNics`, the net dev index in topo reflects the plugin `ncclIbDevs[]`

But when gin setups IB connection
```
ncclDevCommCreate
└── ncclCommGroupRegisterSymmetric
    └── ncclDevrCommCreateInternal
        └── ncclGinConnectOnce
            ├── ncclTopoGetLocalGinDevs
[...]

PR #2166, opened 2026-05-09
fix: use gin_dev attribute for GIN network devices

Outsider3911:
## Description

Separate GIN and net plugin dev indices via dedicated gin_dev XML attribute.

## Related Issues

https://github.com/NVIDIA/nccl/issues/2165

## Performance Impact

The net plugin and GIN will use their respective dev indices.


xiaofanl-nvidia:
++ @thomasgillis to take a look. If it looks good, please mirror it for merging.

Issue #2062, opened 2026-03-22
[Issue]: Potential BUG: Memory Visibility and Race Condition in bootstrapNetInit (Identified by RustCC)


yunquleonliu:
### How is this issue impacting you?

Data corruption

### Share Your Debug Logs

I have developed a Rust C/C++ library that applies Rust abstractions on CC code. 

### NCCL Version

NCCL 2.28.7 (package revision 1)

### Your platform details

Title: [Potential Bug] Memory Visibility and Race Condition in bootstrapNetInit (Identified by RustCC)

Description:
A potential concurrency risk was identified in src/bootstrap.cc during a scan by RustCC (a specialized C++ profiler/static analysis tool based on Rust-style memory safety rules). The implementation of bootstrapNetInit uses a Double-Checked Locking (DCL) pattern that lacks the necessary atomic synchronization required by the C++11 (and later) memory model.

Location:
src/bootstrap.cc
Lines: 86-96

C++
86  static char bootstrapNetIfName[MAX_IF_NAME_SIZE+1];
87  static union ncclSocketAddress bootstrapNetIfAddr;
88  static int bootstrapNetInitDone = 0;  // <--- Risk: Plain int
...
94  if (bootstrapNetInitDone == 0) {
95      std::lock_guard<std::mutex> lock(bootstrapNetMutex);
96      if (bootstrapNetInitDone == 0) {
Technical Analysis:
The variable bootstrapNetInitDone is a plain int accessed outside the mutex at line 94. While this pattern is common, it poses two significant risks in modern C++:

Instruction Reordering: Without std::atomic and proper memory barriers, the compiler or CPU could reorder the store bootstrapNetInitDone = 1 before the actual initialization of bootstrapNetIfName or bootstrapNetIfAddr. A concurrent thread could see bootstrapNetInitDone == 1, skip the lock, and attempt to use uninitialized networking data.

Memory Visibility: On non-x86 architectures (e.g., ARM64/Grace-Hopper), the update to bootstrapNetInitDone might not be immediately visible to other cores without an explicit memory barrier, leading to unnecessary contention.

Tool Report (RustCC):

Rule: TCC-CONC-001 (Unsynchronized Access in DCL)

Confidence: HIGH

Interpretation: Violation of memory safety policy. The code relies on implicit hardware ordering which is not guaranteed across all supported NCCL platforms.

Suggested Fix:
[...]

PR #2168, opened 2026-05-10
bootstrap: Fix data race in network init flag

ankohuu:
## Description

The bootstrapNetInitDone variable in bootstrapNetInit was declared as int, 
causing a data race when concurrent callers read it before taking bootstrapNetMutex
while another thread writes it after completing bootstrap network interface initialization.

This commit changes bootstrapNetInitDone from int to std::atomic<int> and
uses acquire/release ordering around the double-checked initialization path.

## Related Issues

#2062

## Changes & Impact

Use an atomic boot initial flag

### DIFF
--- a/src/bootstrap.cc
+++ b/src/bootstrap.cc
@@ -15,6 +15,7 @@
 #include "ras.h"
 #include <mutex>
 #include "os.h"
+#include <atomic>
 #include <thread>
 #include <chrono>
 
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
++ @KaimingOuyang to take a look. If it's good, please mirror for merging.

KaimingOuyang:
I am not sure why it can cause race condition. `bootstrapNetInit()` is only called once through `std::call_once(initOnceFlag, initOnceFunc)` so that only 1 thread should access `bootstrapNetInitDone`. Can you elaborate the race condition in your case?

PR #1834, opened 2025-09-05
fix #1808: do not retry on bad magic when proxy and ras net accept

fishautumn:
fix #1808

Issue #1808, opened 2025-08-15
[Issue]: nccl receiving an external TCP request causes the proxy thread's ncclProxyService to hang


YeSho-cpp:
### How is this issue impacting you?

Application hang

### Share Your Debug Logs

2025-08-12 09:37:02.216313 worker-9 >> :168:1228 [2] NCCL INFO Connected all trees
2025-08-12 09:37:02.25487 worker-9 >> :169:1234 [3] NCCL INFO Connected all trees
2025-08-12 09:37:02.259538 worker-9 >> :170:1232 [4] NCCL INFO Connected all trees
2025-08-12 09:37:02.259607 worker-9 >> :171:1235 [5] NCCL INFO Connected all trees
2025-08-12 09:37:02.259932 worker-9 >> :167:1230 [1] NCCL INFO Connected all trees
2025-08-12 09:37:02.259934 worker-9 >> :166:1233 [0] NCCL INFO Connected all trees
2025-08-12 09:37:02.259936 worker-9 >> :173:1229 [7] NCCL INFO Connected all trees
2025-08-12 09:37:02.259938 worker-9 >> :172:1231 [6] NCCL INFO Connected all trees
2025-08-12 09:37:03.463948 worker-9 >> [2025-08-12 01:37:03] :166:1226 [0] misc/socket.cc:484 NCCL WARN socketFinalizeAccept: wrong magic 74656d2f20544547 != 7970663b906d70b9
2025-08-12 09:37:14.267026 worker-9 >> :166:1835 [0] NCCL INFO Channel 00/0 : 80[0] -> 87[7] via P2P/CUMEM
2025-08-12 09:37:14.28165 worker-9 >> :168:1836 [2] NCCL INFO Channel 03/0 : 82[2] -> 90[2] [send] via NET/IBext_v10/10/GDRDMA
2025-08-12 09:37:14.28173 worker-9 >> :168:1836 [2] NCCL INFO Channel 11/0 : 82[2] -> 90[2] [send] via NET/IBext_v10/10/GDRDMA
2025-08-12 09:37:14.288816 worker-9 >> :171:1837 [5] NCCL INFO Channel 04/0 : 85[5] -> 93[5] [send] via NET/IBext_v10/13/GDRDMA
2025-08-12 09:37:14.288886 worker-9 >> :171:1837 [5] NCCL INFO Channel 12/0 : 85[5] -> 93[5] [send] via NET/IBext_v10/13/GDRDMA
[...]

PR #2174, opened 2026-05-12
bootstrap, socket: handle bad magic without blocking caller

Alkaid-Benetnash:
### Overview
This PR extends the bad-magic handling in #1808 / PR #1834 (commits `8fdd98c` and follow-up `7310803`) to the bootstrap layer and the async-mode progress path, both of which remain vulnerable on `v2.30u1`.

**Provenance and verification status.** This PR is derived from a corresponding fix we developed and verified in RCCL ([ROCm/rccl#2177](https://github.com/ROCm/rccl/pull/2177)) where the same hang reproduced reliably at scale and is no longer observed after the fix. We have not independently observed or reproduced the equivalent hang in NCCL, and the NCCL patch in this PR has not been tested at production scale. We are submitting it upstream because the relevant code paths in NCCL `v2.30u1` are structurally identical to the RCCL paths we patched, and the residual gaps described below are present in current `v2.30u1` HEAD.

**Branch target.** Targeting `v2.30u1` rather than `master` because the prior bad-magic fixes were merged to `v2.30u1`; this follows the same convention.

### Why the existing fixes are not enough

Issue #1808 / PR #1834 added `ncclSocketStateBadMagic` and a `retryOnBadMagic` parameter to `ncclSocketAccept`, then set `retryOnBadMagic=false` at the proxy and ras callsites so the do-while inside `ncclSocketAccept` exits on bad-magic events instead of looping forever.

That change covers proxy and ras only. Two remaining gaps:

**Gap 1: Bootstrap accept callsites (sync-mode monopolization)**

Four `ncclSocketAccept` callsites in `bootstrap.cc` (`bootstrapRoot`, `socketRingConnect`, `bootstrapInit`, `socketAccept`) retain the default `retryOnBadMagic=true`. When a non-NCCL peer hits the bootstrap listen socket, the do-while resets state from `BadMagic` back to `Accepting` after each event and continues. Once the bad-magic queue is drained, `accept()` returns `EAGAIN` (the listen socket is non-blocking), state stays `Accepting`, and the loop spins. The bootstrap thread is trapped inside `ncclSocketAccept` and cannot service legitimate peer connects. A single stuck rank cascade-hangs the whole job's `commInit`.

**Gap 2: Async-mode `ncclSocketReady` callers (silent stall)**

`socketFinalizeAccept` transitions to `BadMagic` on magic mismatch. `socketProgressState`'s if-cascade has no case for `BadMagic` — the only `BadMagic` reset lives inside `ncclSocketAccept`'s do-while and only fires while that loop runs. Async-mode callers re-enter `socketProgressState` via `ncclSocketReady` (e.g. transport accept path); if state lands on `BadMagic` there, `ncclSocketReady` returns `*running=0` indefinitely. The thread is not stuck (it continues servicing other work), but the specific per-`recvComm` accept stage stops making progress and any global PG depending on that peer-pair handshake hangs.

This is a regression introduced by `8fdd98c`'s addition of the `BadMagic` terminal state. Pre-`8fdd98c` code had no `BadMagic` state to get stuck at.

## Related Issues

#1808 / PR #1834

### Added a`bootstrap.cc` — `bootstrapAccept` wrapper

```c
static ncclResult_t bootstrapAccept(struct ncclSocket* sock, struct ncclSocket* listenSock,
                                    volatile uint32_t* abortFlag) {
  while (1) {
    if (abortFlag != NULL && COMPILER_ATOMIC_LOAD(abortFlag, std::memory_order_acquire) != 0) {
      return ncclInternalError;
    }
    NCCLCHECK(ncclSocketInit(sock));
    NCCLCHECK(ncclSocketAccept(sock, listenSock, /*retryOnBadMagic=*/false));
    if (sock->state != ncclSocketStateBadMagic) return ncclSuccess;
    INFO(NCCL_INIT|NCCL_NET, "bootstrap: rejected bad-magic peer connection on listen sock, retrying accept");
    (void)ncclSocketClose(sock);
    usleep(1000);
  }
}
```

Pass `retryOnBadMagic=false` so `ncclSocketAccept` returns control to this wrapper after each bad-magic event; close the rejected socket; yield 1 ms (avoid 100% CPU busy-spin under sustained flood); honor `abortFlag` between iterations. The four bootstrap callsites are updated to call this helper.

### `misc/socket.cc` — `socketProgressState` BadMagic recovery

```c
static ncclResult_t socketProgressState(struct ncclSocket* sock) {
  if (sock->state == ncclSocketStateBadMagic) {
    sock->state = ncclSocketStateAccepting;
  }
  if (sock->state == ncclSocketStateAccepting) {
    NCCLCHECK(ncclOsSocketTryAccept(sock));
  }
  ...
}
```

Makes the if-cascade self-healing: any caller (sync or async) that re-enters `socketProgressState` with `state=BadMagic` gets state reset to `Accepting` and progresses normally on the next poll. Sync callers that pass `retryOnBadMagic=false` (proxy, ras, the new `bootstrapAccept`) are unaffected: their state only becomes `BadMagic` after `socketProgressState` returns inside the loop, the do-while exits, and the caller observes `BadMagic` before the next `socketProgressState` call.

([ROCm/rccl#2177](https://github.com/ROCm/rccl/pull/2177)) has more details on the observed symptoms without the fix.
And init hang disappears with this PR.

## Performance Impact

I expect there to be no performance concerns regarding this PR. Init hang fix is not on the perf critical path.

### DIFF
--- a/src/bootstrap.cc
+++ b/src/bootstrap.cc
@@ -272,6 +272,27 @@ struct extInfo {
 #define NET_HANDLE(h, rank)    ((h) + (rank * NCCL_NET_HANDLE_MAXSIZE))
 #define BOOTSTRAP_HANDLE(h, i) ((struct ncclBootstrapHandle*)((char*)h + i * NCCL_UNIQUE_ID_BYTES))
 
+// Bootstrap-side accept wrapper. ncclSocketAccept's default behavior
+// (retryOnBadMagic=true) loops internally on bad-magic events, which can
+// monopolize CPU and starve legitimate peer connects when a non-NCCL TCP
+// peer hits the bootstrap listen socket. We pass retryOnBadMagic=false and
+// drive the retry from here: close the rejected socket, yield 1ms, retry.
+// abortFlag is honored between iterations.
+static ncclResult_t bootstrapAccept(struct ncclSocket* sock, struct ncclSocket* listenSock,
+                                    volatile uint32_t* abortFlag) {
+  while (1) {
+    if (abortFlag != NULL && COMPILER_ATOMIC_LOAD(abortFlag, std::memory_order_acquire) != 0) {
+      return ncclInternalError;
+    }
+    NCCLCHECK(ncclSocketInit(sock));
+    NCCLCHECK(ncclSocketAccept(sock, listenSock, /*retryOnBadMagic=*/false));
+    if (sock->state != ncclSocketStateBadMagic) return ncclSuccess;
+    INFO(NCCL_INIT|NCCL_NET, "bootstrap: rejected bad-magic peer connection on listen sock, retrying accept");
+    (void)ncclSocketClose(sock);
+    usleep(1000);
+  }
+}
+
 static ncclResult_t rootSend(union ncclSocketAddress* addr, uint64_t magic, union ringConnectInfo* info) {
   ncclResult_t res = ncclSuccess;
   struct ncclSocket sock;
@@ -310,8 +331,8 @@ static void* bootstrapRoot(void* rargs) {
   /* Receive addresses from all ranks */
   do {
     struct ncclSocket sock;
-    NCCLCHECKGOTO(ncclSocketInit(&sock), res, out);
-    NCCLCHECKGOTO(ncclSocketAccept(&sock, listenSock), res, out);
+    // No abortFlag: bootstrapRoot is a detached thread without a comm.
+    NCCLCHECKGOTO(bootstrapAccept(&sock, listenSock, /*abortFlag=*/NULL), res, out);
     NCCLCHECKGOTO(socketRecv(&sock, &info, sizeof(info)), res, out);
     NCCLCHECKGOTO(ncclSocketClose(&sock), res, out);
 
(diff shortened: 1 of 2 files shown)


Alkaid-Benetnash:
cc @kiskra-nvidia who reviewed the prior badmagic PR.

kiskra-nvidia:
Thank you for this PR!

I agree that the change to handle bad magic is not fully correct, but neither is the fix being proposed here (the socket state needs to be reset in the _callers_ of `socketProgressState`, not in that function itself -- i.e., we need to move that code to `ncclSocketReady`). Anyway, I'll take your proposed change to `socket.cc` and iterate on it internally, as I see more issues with our current code...

I don't understand the motivation for changes to `bootstrap.cc`. `ncclSocketAccept` calls are _supposed_ to block for non-async sockets when `retryOnBadMagic` is `true` (which it is by default), until a valid connection is established. Are you saying that a bad magic issue during bootstrap results in a hang that's unrecoverable, i.e., an invalid connection followed by a valid one does _not_ recover the state (`ncclSocketAccept` does not return a fully established socket)? If so, then that would indicate a bug in the socket code, and we should fix it there, not add workarounds to the callers.

kiskra-nvidia:
@Alkaid-Benetnash Please see commit dd40473 (to be included in the next 2.30 release update) that should make the handling of bad magic more consistent. Hopefully it fixes the issues you've come across, but if not, then let's iterate on it.

xiaofanl-nvidia:
Closing.

PR #2182, opened 2026-05-18
Pin GIN proxy thread to NUMA-local CPU set

ngoyal2707:
The GIN proxy thread (ncclGinProgress) currently inherits its CPU affinity from
whatever thread calls ncclGinDevCommSetup. That happens after ncclCommInitRank
returns, so the temporary NUMA pin established during init has been restored,
and the GIN proxy ends up with the caller's affinity.

The traditional proxy thread (ncclProxyService) is spawned during init while
the calling thread is still NUMA-pinned, so it inherits the correct affinity
automatically.

This change saves comm->cpuAffinity on ginState before spawning the proxy
thread, and applies it at the top of ncclGinProgress to bring GIN parity with
the traditional proxy.

### DIFF
--- a/src/gin/gin_host.cc
+++ b/src/gin/gin_host.cc
@@ -42,6 +42,9 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
 
 void* ncclGinProgress(struct ncclGinState* ginState_) {
   struct ncclGinState* ginState = (struct ncclGinState*)ginState_;
+  if (ncclOsCpuCount(ginState->cpuAffinity)) {
+    ncclOsSetAffinity(ginState->cpuAffinity);
+  }
   while (1) {
     std::unique_lock<std::mutex> lock(ginState->mutex);
     if (ginState->ginProgress == 1) {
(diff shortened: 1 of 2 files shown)


src/gin/gin_host.cc
```diff
@@ -42,6 +42,9 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
 
 void* ncclGinProgress(struct ncclGinState* ginState_) {
   struct ncclGinState* ginState = (struct ncclGinState*)ginState_;
+  if (ncclOsCpuCount(ginState->cpuAffinity)) {
+    ncclOsSetAffinity(ginState->cpuAffinity);
```
davidozog, line 46: Side question: I know we don't do it elsewhere (e.g., `ncclProxyService`), but should we check the `ncclResult_t` here and report it through `ginState->asyncResult`?  Or is best-effort ok for cpu affinity?

src/include/gin/gin_host.h
```diff
@@ -24,6 +25,7 @@ struct ncclGinStateDevComm {
 };
 
 struct ncclGinState {
+  ncclAffinity cpuAffinity;
```
davidozog, line 28: This no longer mirrors `ncclGinState` in `gin_host_win_stub.h`.  That may not matter now... but since it's just a stub maybe we should just keep them mirrored for now?

davidozog (approved):
Looks like a good catch to me, thanks @ngoyal2707.

Just a couple questions for @ngoyal2707 @kgioioso.  I'm good with mirrorring the MR to target v2.30u1 if you are.

src/include/gin/gin_host.h
ngoyal2707: @davidozog Thanks for reviewing, as discussed offline, please feel free to commandeer and add changes that are needed

xiaofanl-nvidia:
Merged here https://github.com/NVIDIA/nccl/commit/80e715f91f8ebd6f8782019573e571c86abb5921 Thanks for your contribution! @ngoyal2707

PR #2205, opened 2026-06-04
Fix rma proxy connect cleanup

wanglei875:
## Summary
Clean up partially initialized RMA proxy resources when `ncclRmaProxyConnectOnce()` fails.

The failure path now closes:
- the current listen communicator if `listen()` succeeded
- any created RMA proxy contexts
- any established RMA collective communicators

Also routes `rmaProxyCtxs` allocation through the existing failure cleanup path.

## Testing
- `git diff --check -- src/rma/rma_proxy.cc`
- `make -j$(nproc) src.build` compiled `rma/rma_proxy.cc`; full build later failed on an existing missing `plugin_cleanup.h` include in `plugin/rma/rma_v13.cc`.


xiaofanl-nvidia:
++ @jynv to review and mirror if it looks good.

PR #2208, opened 2026-06-04
dev_runtime: preserve GIN resource requirements

wanglei875:
## Summary

`deepCopyDevCommRequirements()` copies `ncclDevResourceRequirements` nodes before queuing device communicator creation work, but it only preserved the buffer fields.

This change also preserves the GIN resource fields:
- `ginSignalCount`
- `ginCounterCount`
- `outGinSignalStart`
- `outGinCounterStart`

These fields are later used to assign GIN signal/counter ranges and compute the total GIN resource counts.
This is consistent with outBufferHandle and outMultimemHandle, which are already preserved across the async task boundary.
## Testing

- Ran `git diff --check`
- Full build not run: `nvcc` is not available in this environment

### DIFF
--- a/src/dev_runtime.cc
+++ b/src/dev_runtime.cc
@@ -969,6 +969,10 @@ static ncclResult_t deepCopyDevCommRequirements(struct ncclDevCommRequirements c
     (*dstRes)->bufferSize = rr->bufferSize;
     (*dstRes)->bufferAlign = rr->bufferAlign;
     (*dstRes)->outBufferHandle = rr->outBufferHandle;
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
++ @kiskra-nvidia to review and mirror.

kiskra-nvidia (approved):
This looks good.

I was thinking at first that it would be a headache, because we added those fields in 2.28.7, and the structure concerned is not versioned. However, we do not provide backwards compatibility to versions prior to 2.29.2, so we don't need to do anything extra.

xiaofanl-nvidia:
This has been merged to dev branch. Closing.

PR #2212, opened 2026-06-04
Replace reduce with allreduce in NCCL4Py allreduce test

jdinan:
## Description

The test is supposed to run allreduce, but was calling reduce. Without this change, the print statement produces incorrect output.

## Related Issues

None

## Changes & Impact

No impact

## Performance Impact

Correctness only


jdinan:
Not a bug -- reduce is overloaded: with root specified it's NCCL reduce, without root it's allreduce.

Issue #2196, opened 2026-05-26
[RFE]: Misleading RING_LL suffix in AllGather/ReduceScatter kernel names


baptiste-bt:
Hello, we noticed that for AllGather and ReduceScatter collectives using the RING algorithm, the kernel name always includes the `RING_LL` suffix, whatever the actual NCCL proto chosen for a specific collective. From similar issues, I understand that this helps reduce NCCL binary size, but I still find this a bit misleading and I was wondering what you thought about:

1. Not using any suffix by default instead of always including `LL` even when a different proto is chosen
2. Adding the possibility to make this configurable at build time to allow builds from source that can afford to specialize kernel names

On our side, we mostly monitor collectives at scale through `nsys` or `torch.profiler` during training, which directly display kernel names using the wrong proto, making us think that NCCL was using LL. Using `NCCL_DEBUG_SUBSYS=TUNING` was helpful to debug this, but too verbose to always include in training runs.

Related issues:
- https://github.com/NVIDIA/nccl/issues/469
- https://github.com/NVIDIA/nccl/issues/502

cc @sjeaugey

sjeaugey:
For reference, a NCCL kernel named ncclDevKernel_allreduce_ring_LL is a kernel which can execute any operation (any function: broadcast, reduce, allreduce, send/recv), with any algorithm and any protocol, but where one combination in particular is inlined in the kernel to avoid a high-latency function call, in this case, allreduce_ring_LL. The kernel we pick is generally based on the first (or only) operation which will be executed, and we pick the version which has the same function and algorithm, but potentially not the protocol that's inlined because we only generate inline kernels for low-latency operations, i.e. using the LL protocol. The reason why we only generate inline kernels for the LL protocol is binary size; we used to generate all protocols but at some point NCCL was getting too large so we had to compromise.

So, although the name doesn't say anything about what this kernel is executing, it is a good hint about the first operation (especially when there is only one) and even which algorithm it uses. And for that reason, over time, users have used the kernel name to interpret their nsys traces and see where their different operations were. I'm not blaming them for doing that, I do the same when I need to. It's convenient. But also misleading at first.

All this may change at some point. But in the meantime, and as long as we only generate _LL kernels, we could drop _LL from the name indeed. That would cause less confusion ... although it would make users rely even more on the name to know what's being executed, making it even more confusing if we still have cases where the name doesn't match.

sjeaugey:
This patch seems to work, but I haven't tested it extensively ...

```
[...]

PR #2214, opened 2026-06-05
docs: clarify profiler kernel names

Rohithmatham12:
## Summary
- Document that NCCL profiler kernel names may use representative generated kernel names.
- Point users to `NCCL_DEBUG_SUBSYS=TUNING` when they need the exact selected algorithm and protocol.

Addresses #2196.

## Testing
- `git diff --check`

This is a documentation-only change.


xiaofanl-nvidia:
++ @armratner @gcongiu to review and mirror.

PR #2215, opened 2026-06-05
register: restore CUDA device on deregister early exits

Jonathan-0256:
## Summary

`commDeregister()` saves the caller's current CUDA device and switches to
`comm->cudaDev` before looking up and releasing the registration handle.

Two early return paths skipped restoring the saved device:

- invalid deregistration handle
- registration still referenced by local/graph users

As a result, `ncclCommDeregister()` could leave the caller's thread with
`comm->cudaDev` as the current CUDA device.

This change restores the saved CUDA device on all paths after switching
devices. It also fixes the `CommCheck` API name used by the deregister paths.

## Testing

- Ran `git diff --check`


xiaofanl-nvidia:
++ @KaimingOuyang to review and mirror. CC @AddyLaddy

PR #2224, opened 2026-06-08
Remove misleading NCCL kernel protocol suffix

sanrise:
## Description

Removes the misleading protocol suffix from generated kernel symbols. Every kernel is a universal dispatcher that runs the tuned protocol at runtime through `ncclDevFuncTable`; only the LL body is inlined (for binary size), so a collective tuned to Simple or LL128 still shows up under a `_LL` kernel in `nsys` / `torch.profiler`, and trace-based analysis attributes performance to the wrong protocol. This drops the protocol component from generated `ncclDevKernel_*` symbols so the name carries only collective + algorithm (both accurate). The protocol-specific `ncclDevFunc*` dispatch entries are left intact, so runtime behavior is unchanged. (Same direction as the maintainer suggestion on #2196.)

## Related Issues

Part 2 of 2 for #2196. Depends on / should merge after the task protocol trace metadata PR: https://github.com/NVIDIA/nccl/pull/2223

## Changes & Impact

- `src/device/generate.py`: generated `ncclDevKernel_*` symbols no longer include the protocol component.
- No runtime or dispatch change; `ncclDevFunc*` entries are unchanged.
- Kernel symbol names change, so any tool that parses the protocol out of the kernel name will stop seeing it (that is the intent; the protocol is exposed via the metadata PR instead).

## Performance Impact

- None expected; this is a name-only generator change.
- Generator output verified: host table has protocol-neutral `ncclDevKernel_AllReduce_Sum_u8_RING`, the device body still uses `NCCL_PROTO_LL`, and no `_RING_LL` host-table symbol remains. Standalone master build passes.

### DIFF
--- a/src/device/generate.py
+++ b/src/device/generate.py
@@ -67,6 +67,12 @@ make ONLY_FUNCS="AllReduce Sum f32 RING LL128"
 def paste(sep, *args):
   return sep.join(x for x in args if x is not None)
 
+# Kernel names intentionally omit proto. The kernel body may dispatch a
+# different tuned protocol through ncclDevFuncTable, so a proto suffix would be
+# a cosmetic signal rather than the executed protocol.
+def kernel_sym(*fn):
+  return paste("_", *fn[0:4])
+
 func_pattern = sys.argv[2:3]
 if func_pattern and func_pattern[0]:
   import re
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
++ @armratner @gcongiu

MoraruMaxim:
@sanrise we just close this PR.

PR #2225, opened 2026-06-09
dev_runtime: free requirements on DevCommCreate failure

wanglei875:
## Description

When ncclDevCommCreate deep-copies the user-provided requirements into the async task, later failures before the task is enqueued only free the task wrapper.

Free the copied requirements on the failure path so the copied resourceRequirementsList/teamRequirementsList storage is released as well.

Validation:
- git diff --check
- source-level invariant check for the ncclDevCommCreate failure path

### DIFF
--- a/src/dev_runtime.cc
+++ b/src/dev_runtime.cc
@@ -1547,7 +1547,10 @@ exit:
   cudaSetDevice(saveDev);
   return ret;
 fail:
-  free(task);
+  if (task) {
+    freeDevCommRequirements(task->reqs);
+    free(task);
+  }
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
@marksantesson can you review and mirror if reasonable? Thanks!

marksantesson:
@wanglei875 , thanks for your contribution! It has been [committed](https://github.com/NVIDIA/nccl/commit/385b4a8e3c119c95eb7f678e1f47f3421cc9f72e) to the dev branch.

PR #2227, opened 2026-06-09
net_ib: free GIN barrier buffer on failure

wanglei875:
## Description

ncclGinIbP2PBarrier allocates a temporary dummy buffer and then calls ncclGinIbAllGather. If the allgather fails, the existing NCCLCHECK path returns before freeing the buffer.

Use a local cleanup label so the dummy buffer is released on both success and failure paths.

Validation:
- git diff --check
- source-level invariant check for the ncclGinIbP2PBarrier cleanup path


xiaofanl-nvidia:
++ @jynv to mirror if reasonable.

PR #2228, opened 2026-06-09
bound remote device count in ib connect and accept

jmestwa-coder:
Peer-supplied ndevs in the IB connection metadata is used to index fixed-size arrays without a range check:
- ncclIbConnectImpl stores remMeta.devs[i] into remDevs[]/rkeys[] for i up to remMeta.ndevs
- ncclIbAcceptImpl does the same on the receiver side
- those arrays are sized NCCL_IB_MAX_DEVS_PER_NIC, so a peer reporting ndevs above that count reads and writes past the end
Validate the count is within range right after the metadata is received, on both sides.

### DIFF
--- a/src/transport/net_ib/connect.cc
+++ b/src/transport/net_ib/connect.cc
@@ -1025,6 +1025,11 @@ ib_connect:
 
   memcpy(&remMeta, stage->buffer, sizeof(ncclIbConnectionMetadata));
 
+  if (remMeta.ndevs <= 0 || remMeta.ndevs > NCCL_IB_MAX_DEVS_PER_NIC) {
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
++ @thomasgillis to mirror if reasonable.

jmestwa-coder:
thanks. trigger is a remote ib peer sending ncclIbConnectionMetadata with ndevs above NCCL_IB_MAX_DEVS_PER_NIC during the connect/accept handshake. the storage loops then read remMeta.devs[i] and write remDevs[]/rkeys[] past the 8-entry arrays, so the guard rejects it on both sides before the loop runs. happy to share a small standalone asan repro of the loop if that helps with mirroring.

jmestwa-coder:
any update?

Issue #2136, opened 2026-04-27
[Issue]: GIN main qps load balance problem when use a lag port with queue affinity policy


baymaxhuang:
### How is this issue impacting you?

Lower performance than expected

### Share Your Debug Logs

Before patch:
```
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=0, main_qpn=0x11db5, remote_rank=0, remote_qpn=0x11eb9
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=1, main_qpn=0x11db7, remote_rank=1, remote_qpn=0x1a86a
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=2, main_qpn=0x11db9, remote_rank=0, remote_qpn=0x11eba
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=3, main_qpn=0x11dbb, remote_rank=1, remote_qpn=0x1a86e
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=4, main_qpn=0x11dbd, remote_rank=0, remote_qpn=0x11ebb
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=5, main_qpn=0x11dbf, remote_rank=1, remote_qpn=0x1a872
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=6, main_qpn=0x11dc1, remote_rank=0, remote_qpn=0x11ebc
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=7, main_qpn=0x11dc3, remote_rank=1, remote_qpn=0x1a876
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=8, main_qpn=0x11dc5, remote_rank=0, remote_qpn=0x11ebd
2026-04-27 10:04:20 test02:307183:307183 [0] NCCL INFO [0] Connected main QP: qp_idx=9, main_qpn=0x11dc7, remote_rank=1, remote_qpn=0x1a87a
[...]

Issue #2137, opened 2026-04-27
[Issue]: ginConnectionType in ncclDevCommCreate requirements not taking effect


huakainomore:
### How is this issue impacting you?

set ginConnectionType to NCCL_GIN_CONNECTION_RAIL, but actually use NCCL_GIN_CONNECTION_FULL.
Lead to created more than expected QPs .

### Steps to Reproduce the Issue

in  [EP v2](https://github.com/deepseek-ai/DeepEP/tree/epv2-release), the ginConnectionType field is set in ncclDevCommRequirements_t to NCCL_GIN_CONNECTION_RAIL before calling ncclDevCommCreate, However, the connection type setting does not take effect as expected, because 
in [src/gin/gin_host.cc/ncclGinConnectOnce](https://github.com/NVIDIA/nccl/blob/v2.30.4-1/src/gin/gin_host.cc), the ginConnectionType set to comm->globalGinSupport rather than requestedConnectionType, we do a patch to fix it [patch](https://github.com/user-attachments/files/27127970/0001-feat-fix-ginConnectionType-in-ncclGinConnectOnce.patch)
```code
    if (num_ranks > 1 and get_env("EP_DISABLE_GIN", 0) == 0) {
        reqs.ginContextCount = num_allocated_qps;
        reqs.ginExclusiveContexts = true;
        reqs.ginQueueDepth = 1024;
        reqs.ginTrafficClass = sl_idx;
        // Customized RDMA barrier needs extra signals
        reqs.ginSignalCount = num_ranks + 2 * 2;
        reqs.ginConnectionType = allow_hybrid_mode ? NCCL_GIN_CONNECTION_RAIL: NCCL_GIN_CONNECTION_FULL;
    }
    NCCL_CHECK(ncclDevCommCreate(comm, &reqs, &dev_comm));
```

### NCCL Version

v2.30.4-1+cuda13.1

huakainomore:
```
- ginState->ginConnectionType = requestedConnectionType;
+  ginState->ginConnectionType = comm->globalGinSupport;
```
i saw [b3aeda4](https://github.com/NVIDIA/nccl/commit/b3aeda4a9a8abd360e6ee0381a99b0ab2779d213) have a modification. Could I ask for the reason for this modification? @sjeaugey  many thanks!
it seems set reqs.ginConnectionType = NCCL_GIN_CONNECTION_RAIL not taking effect unless set NCCL_CROSS_NIC=0.

[...]

PR #2230, opened 2026-06-09
fix gin proxy load balance problem when nic is a lag port

YeSho-cpp:
## Description

Fix  a QP load balance problem when using ROCE LAG (mode 4, 802.3 AD) with queue affinity policy by GIN proxy

## Related Issues

https://github.com/NVIDIA/nccl/issues/2136

## Performance Impact

make gin proxy qp can full use of the dual uplink port of bonding ib device


xiaofanl-nvidia:
@YeSho-cpp did the recent rankStride change regress this case and your PR fixes the regression? 

++ @sjeaugey , @jynv to mirror if the fix is reasonable.

YeSho-cpp:
> @YeSho-cpp did the recent rankStride change regress this case and your PR fixes the regression?
> 
> ++ @sjeaugey , @jynv to mirror if the fix is reasonable.

These are two separate issues. The recent rankStride fix addresses connectionType in ncclDevCommCreate not taking effect in gin proxy (https://github.com/NVIDIA/nccl/issues/2137). The previous PR (https://github.com/NVIDIA/nccl/commit/a93cc034b7faf4200f24d3721a65bb9d5f5fb67c) fixed the issue in gdaki mode, while this new PR resolves the load balancing problem in proxy mode.

Issue #2035, opened 2026-03-05
[Issue]: Low performance with NVLSTree algo on B300


GeofferyGeng:
### How is this issue impacting you?

Lower performance than expected

NVLSTree:
2 Nodes ==> 800GBps +
4/8/16 Nodes ==> 200GBps +
Ring:
2/4/8/16 ==> 760GBps +

### Share Your Debug Logs

Nope

### NCCL Version

2.29.7+cuda13.1

### Your platform details

16 nodes , B300 \* 8 + CX8 \* 8 per Node

xiaofanl-nvidia:
This is a known issue and we are working on a fix. ++ @tarudoodi for viz.

GeofferyGeng:
@xiaofanl-nvidia Understood. Thank you for the update and looking forward to the fix. I'll leave this issue open unitll the fix.

teojgo:
@GeofferyGeng can you test using the upcoming release branch: https://github.com/NVIDIA/nccl/tree/v2.30

teojgo:
FYI, the release including the fix is out: https://github.com/NVIDIA/nccl/releases/tag/v2.30.3-1.

47oo:
@GeofferyGeng  Can you tell me what ib switch is ,800G 2u OR 4u ,and multi plane open or close .I have i issue when i use multi plane ,only run 200GB witl all_reduce_perf, thx

GeofferyGeng:
@47oo Sorry for the late reply. Exactly we use eth switch so we must split cx8 to 2*400Gbps. Consider the situation, you may use single qp on multi-plane?

GeofferyGeng:
@teojgo Sorry for the late reply. We retested with 2.30.4, results is 380GBps busbw with 4 nodes which is similar to Tree algo 370GBps. Is this expected?

47oo:
@GeofferyGeng we found the reason becauseof nvidia-open gpu driver,thx

tarudoodi:
> [@teojgo](https://github.com/teojgo) Sorry for the late reply. We retested with 2.30.4, results is 380GBps busbw with 4 nodes which is similar to Tree algo 370GBps. Is this expected?

@GeofferyGeng that is low performance for BusBW, but is close to the algo BW at 4 nodes.

- Can you share how you are running the test?
- Are you seeing improved perf at 2 nodes with v2.30.4? We improved that too.
[...]

PR #2244, opened 2026-06-23
net/ib: deterministic sub-device order in fused vNICs (fix cross-plane QP pairing under SR-IOV multi-plane RoCE)

koiker:
## Problem

When two or more physical IB/RoCE devices are fused into one virtual NIC (`NCCL_IB_MERGE_NICS` /
`NCCL_NET_FORCE_MERGE`), `ncclIbMakeVDeviceInternal` stores the sub-devices in **PCI/BDF enumeration
order**. QP-to-sub-device pairing inside a fused vNIC is **positional** (`qpIndex % ndevs` locally,
and the remote's advertised `devIndex`), so both endpoints must agree on sub-device order.

For **SR-IOV VFs of a dual-port / multi-plane NIC** (e.g. a ConnectX-8 whose two 400G ports land on
two RoCE planes, exposed as VFs to Kubernetes pods), the relative BDF order of the plane-0 VF vs the
plane-1 VF is **not stable across nodes**. So `slot0<->slot0` positional pairing can connect
**plane 0 on one node to plane 1 on the peer**. Those VFs are on different GID subnets, so:

- the QP can't reach RTR → `ibv_modify_qp failed with 110 Connection timed out`, or
- when a route exists, traffic is cross-plane and bandwidth collapses.

This is independent of `NCCL_IB_HCA` `:rail:plane` tags (they fix cross-*rail* grouping, not
intra-vNIC sub-device order) and of `NCCL_IB_SUBNET_AWARE_ROUTING` (that selects which *merged
device* matches a remote subnet; it does not reorder sub-devices *within* a fused vNIC).

## Fix

Sort the sub-devices of each fused vNIC **by device name** before building the merged device. Device
names are stable and identical across nodes, so the order is node-independent and positional pairing
becomes plane-consistent. ~14 lines, no new env vars, default single-NIC behavior unchanged.

## Validation

8× **B300**, dual-port **ConnectX-8**, RoCEv2 with two planes, on **Kubernetes + NVIDIA Run:ai**;
RoCE exposed as **SR-IOV VFs** via the **NVIDIA Network Operator** (switchdev) + **NV-IPAM** +
**Multus** (one VF per rail × plane). 4-node `all_reduce_perf` (32 GPU, p0+p1):

| Config | busbw @16 GiB | QP timeouts |
|---|---|---|
| single-plane (8 rails) | ~382 GB/s | 0 (half fabric idle) |
| dual-plane, unsorted (stock) | fail / ~89 GB/s | many (110 cross-plane) |
| **dual-plane, this fix** | **763.9 GB/s** | **0** |

~2× single-plane, exceeding the 2-node single-plane figure (both planes fully utilized).

## Note for k8s / Run:ai users debugging this

Many stacks (NGC/HPC-X images) load an **external** net plugin
(`NET/Plugin: Loaded net plugin NCCL RDMA Plugin v11` from
`/opt/hpcx/nccl_rdma_sharp_plugin/lib/libnccl-net.so`) which carries its **own copy** of this
net_ib code — so an in-tree rebuild has no effect unless you set `NCCL_NET_PLUGIN=none` or apply the
same fix to the plugin. Check `NCCL_DEBUG=INFO ... | grep "NET/Plugin\|Initialized NET plugin"` to
see which backend is on the data path. The same deterministic-ordering fix is needed wherever the
net_ib code actually executes.

Refs #2035.

### DIFF
--- a/src/transport/net_ib/init.cc
+++ b/src/transport/net_ib/init.cc
@@ -218,6 +218,20 @@ ncclResult_t ncclIbMakeVDeviceInternal(int* d, ncclNetVDeviceProps_t* props) {
   // Set the virtual bit on to avoid collision with physical planes when multiple planes are merged.
   mDev->planeId = (props->ndevs > 1) ? NCCL_IB_PLANE_VIRT_BIT : ncclIbDevs[props->devs[0]].planeId;
 
+  // Order the sub-devices of a fused vNIC deterministically (by device name) so that every node
+  // builds the same sub-device ordering. QP-to-sub-device pairing inside a fused vNIC is positional
+  // (qpIndex % ndevs), so a stable, node-independent order is required to pair sub-devices
+  // consistently across endpoints. Without this, the sub-device order follows PCI/BDF enumeration,
+  // which is not stable across nodes for SR-IOV VFs of a multi-port/multi-plane NIC and leads to
+  // cross-plane QP pairing (mismatched GID subnets -> ibv_modify_qp 110 / degraded bandwidth).
+  for (int a = 0; a < props->ndevs - 1; a++) {
+    for (int b = 0; b < props->ndevs - 1 - a; b++) {
+      if (strcmp(ncclIbDevs[props->devs[b]].devName, ncclIbDevs[props->devs[b + 1]].devName) > 0) {
+        int tmp = props->devs[b]; props->devs[b] = props->devs[b + 1]; props->devs[b + 1] = tmp;
+      }
+    }
+  }
+
   for (int i = 0; i < props->ndevs; i++) {
     ncclIbDev* dev = ncclIbDevs + props->devs[i];
     if (mDev->vProps.ndevs == NCCL_IB_MAX_DEVS_PER_NIC) return ncclInvalidUsage;


thomasgillis:
@koiker net_ib sorts the devices by PCIe order (including the VF), which is reliable across ranks.
The relevant line is: https://github.com/NVIDIA/nccl/blob/5067397c2676d5aed50042fc39e5c8ee96eb0027/src/transport/net_ib/init.cc#L484
The order of the device is then re-used for the fusion on the vNICs. Can you explain why this is not sufficient in your use-case? Traditionally, topology injection has been used to solve the VF inconsistency, is this not working for you?

Note: AFAIK the IB device naming is not a reliable across hosts.

koiker:
Thanks — I traced the path you referenced and you're right that it's stable on bare metal; the issue is specific to SR-IOV VFs in containers.

Confirming the mechanism: `IB_DEVICE_PCI_ORDER` defaults to 1, so `ncclIbDevs` is `qsort`'d by `ncclIbCompareDevs`, which compares `fullPciPath` via `pciPathToInt64()` (numeric BDF). `ncclIbMakeVDeviceInternal` then builds the vNIC from `props->devs[]` (indices into that PCI-sorted array), and QP pairing is positional (`devIndex = qpIndex % vProps.ndevs`, remote advertises `devIndex`). So both ends must agree on the intra-vNIC sub-device order, and that order is the BDF order.

That's reliable when BDF("rail r, plane p") is identical on every host. Under Kubernetes + SR-IOV it isn't:

- The pod sees **VFs** injected by the SR-IOV device plugin + Multus, not the physical PFs. The `fullPciPath` NCCL reads is the *assigned VF's* BDF.
- plane-0 and plane-1 are two different PFs (the two CX-8 ports), each with its own VF pool. The VF drawn from each pool is an independent allocation from a free list (scheduling/restart-order dependent). There's no invariant that `BDF(p0 VF) < BDF(p1 VF)`, and it's not coordinated across nodes.
- So the BDF sort orders the two fused sub-devices `p0,p1` on some nodes and `p1,p0` on others. Positional pairing then connects local plane-0 ↔ remote plane-1 → different RoCE GID subnets → `ibv_modify_qp` 110 (unroutable), or cross-plane traffic and collapsed bandwidth. We observed both (110 timeouts, ~89 GB/s), recovered to 763.9 GB/s once the intra-vNIC order was made plane-consistent.

On naming: agreed that `devName` isn't reliable across hosts in the general case. But it's worth noting it's not an arbitrary local convention here — the `roce_p{plane}_r{rail}` names are what NVIDIA's own documented Kubernetes deployment produces. The NIC Configuration Operator (`NicInterfaceNameTemplate`) renames RDMA/netdevs via udev rules anchored to per-rail PCI addresses, and the Kubernetes Launch Kit ships the Spectrum-X default `rdmaPrefix: "roce_p%plane%_r%rail%"`:

- https://docs.nvidia.com/networking/display/kubernetes2640/spectrum-x/spectrum-x-configuration.html
- https://docs.nvidia.com/networking/display/kubernetes2640/k8s-launch-kit/reference/config.html

So for any cluster following the documented NVIDIA Spectrum-X / Launch Kit setup, the device name is deterministic, identical across hosts, and encodes the plane+rail — i.e. it carries exactly the invariant the VF BDF order lacks. That's why sorting by name fixes it in practice for this (large) population of deployments. It's still environment-specific, so I'd rather not standardize on names upstream — `(railId, planeIdx)` with a GID/subnet fallback is the robust key — but the name sort is a reliable fix for anyone on the documented path.

thomasgillis:
@koiker for virtualized environments, the existing policy is to perform a topology injection, see https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-topo-file. 
NCCL will read the XML topology file and use that information instead of the net properties returned by the plugin.
Is that existing solution not working for you?

koiker:
Closing this MR — and documenting how we got here, because the path itself is the useful part.

When I opened it I had a working dual-plane config (`NCCL_NET_PLUGIN=none` + a deterministic fused sub-device sort) and attributed the recovery to the sort. That conclusion was reached **without a proper A/B control**. Working through the review with @TomGillis — and with an AI coding agent (Cursor) driving the build/test/log-analysis loop on the cluster — the question *"could disabling the plugin alone have fixed it?"* turned out to be exactly the right one, so I went back and ran the control.

To remove any doubt about which library was actually executing (plain `LD_PRELOAD` was unreliable on this stack — the external plugin carries its own copy of `net_ib`), I built **stock 2.30.7-1** and **patched 2.30.7-1** from identical source, each with a distinct load-time banner, and deployed by overwriting the on-disk lib. Every rank's banner then proves which one ran.

Results (banner-verified; B300 + Run:ai, SR-IOV dual-plane RoCE; 2 and 4 nodes; force- and auto-merge; plus heterogeneous interface-order and fractional 2-GPU/node allocations):

- **In-tree 2.30.7-1 already orders the fused VF sub-devices correctly and host-consistently.** Stock and patched are behaviorally identical — uniform `p0,p1` order, 0 `ibv_modify_qp` timeouts, ~746 GB/s @ 4 nodes. The patch is a **no-op** on 2.30.7-1 here.
- The failures I originally saw came from **(a)** the external HPC-X net plugin (`NCCL RDMA Plugin v11`), which doesn't honor the merge order and cross-pairs, and **(b)** in-tree **2.29.2**.
- The important bit: the **external NET plugin masked the fact that the latest in-tree version already fixes this.** When the plugin is loaded it overrides the in-tree IB transport (it's selected before `ncclNetIb` and the init loop `break`s), so every in-tree rebuild I did earlier — patched or not — had zero effect on the data path. That's what made the patch *look* necessary.

So the honest conclusion: **2.30.7-1 fixes the issue, with the operational caveat that the external plugin must be disabled (`NCCL_NET_PLUGIN=none`) so the in-tree IB transport runs.** No upstream change is needed for our case, so I'm withdrawing this.

Two notes I'll leave behind:
1. The real fix is whatever changed between **2.29 and 2.30.7** to make the in-tree ordering host-consistent for SR-IOV VFs (2.29.2 still cross-paired) — worth confirming that's intentional/robust.
2. I'll raise the plugin's merge-order behavior separately with the HPC-X (`nccl_rdma_sharp_plugin`) owners. The deterministic-sort idea remains available as an opt-in/env-gated hardening if ever wanted (port-based key preferred, name/GID-subnet fallback).

Thanks @TomGillis for pushing for the control — it changed the conclusion.

PR #2254, opened 2026-06-25
Adding FI_MORE to the Gin Proxy Path

hershys-aws:
## Description

Plugins in the RMA/GIN Host Proxy path would benefit from being able to batch sends at the network level. This change bumps the RMA version and adds in logic for passing `optFlags` from the proxy through NCCL into the plugin, so the plugin can hold off on ringing the NIC doorbell when it knows more ops are coming. i.e. In Libfabric, an `FI_MORE` flag can be set to let the NIC know that more data is coming and to hold off on ringing the doorbell meaning a batch of ops rings one doorbell instead of one per op.

The proxy sets the hint by looking checking if the next slot on the GFD queue is already filled. If so then, more ops are in flight. It falls back to an aggregate bit the kernel can stamp in the GFD, which covers the terminal op of a batch where look-ahead can't see the next op yet.

Note the `FI_MORE` handling itself lives in the plugin (e.g. aws-ofi-nccl), not in this change — NCCL just plumbs the optFlags hint down to it.

## Changes & Impact

- Bumping up the RMA version from v14 -> v15 to add support for passing optFlags to the plugin. Older v14 plugins don't get the param and behave as before.
- Carved out an aggregation bit in the GFD header for aggregating writes via FI_MORE (guarded by a static_assert on the GFD version).
- Added look-ahead logic in the proxy to set the flag when it sees more data coming on the GFD queue.

### DIFF
--- a/src/gin/gin_host_proxy.cc
+++ b/src/gin/gin_host_proxy.cc
@@ -246,7 +246,7 @@ static ncclResult_t proxyGinProcessGfd(struct ginProxyCtx* ctx, struct ginProxyH
     signalOp = mapGfdOpToSignalOp(gfd);
     NCCLCHECK(rmaBackend->iputSignal(ctx->rmaCtx, hostGpuCtx->contextId, 0, nullptr, 0, 0, nullptr, targetRank,
                                      signalOff, signalHandle, signalVal, signalOp, extractIsStrongSignal(gfd),
-                                     &state->request));
+                                     /*optFlags*/ 0, &state->request));
     return ncclSuccess;
   }
 
@@ -261,7 +261,7 @@ static ncclResult_t proxyGinProcessGfd(struct ginProxyCtx* ctx, struct ginProxyH
       return ncclInvalidUsage;
     }
     NCCLCHECK(rmaBackend->iget(ctx->rmaCtx, hostGpuCtx->contextId, srcOff, srcHandle, size, dstOff, dstHandle,
-                               targetRank, &state->request));
+                               targetRank, /*optFlags*/ 0, &state->request));
     return ncclSuccess;
   }
 
@@ -303,7 +303,7 @@ static ncclResult_t proxyGinProcessGfd(struct ginProxyCtx* ctx, struct ginProxyH
     if (signalOp == -1) {
       // First cast from 63 bits to 64 bits and then to void * to avoid warnings
       NCCLCHECK(rmaBackend->iput(ctx->rmaCtx, hostGpuCtx->contextId, srcOff, srcHandle, size, dstOff, dstHandle,
-                                 targetRank, &state->request));
+                                 targetRank, /*optFlags*/ 0, &state->request));
     } else {
       // Reconstruct the signal value
       signalVal = extractSignalVal(gfd);
@@ -312,7 +312,7 @@ static ncclResult_t proxyGinProcessGfd(struct ginProxyCtx* ctx, struct ginProxyH
         sizeof(uint64_t);
       NCCLCHECK(rmaBackend->iputSignal(ctx->rmaCtx, hostGpuCtx->contextId, srcOff, srcHandle, size, dstOff, dstHandle,
                                        targetRank, signalOff, ctx->signalsGinHandle, signalVal, signalOp,
-                                       extractIsStrongSignal(gfd), &state->request));
+                                       extractIsStrongSignal(gfd), /*optFlags*/ 0, &state->request));
     }
     break;
   default:
--- a/src/gin/gin_host_proxy.cc
+++ b/src/gin/gin_host_proxy.cc
@@ -13,6 +13,7 @@
 #include "checks.h"
 #include "gdrwrap.h"
 #include "nccl_device/gin/proxy/gin_proxy_device_host_common.h"
+#include "nccl_device/gin/gin_device_common.h"
 #include "compiler.h"
 #include "comm.h"
 #include "proxy_gpucontext/gpucontext.h"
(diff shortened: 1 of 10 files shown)


src/gin/gin_host_proxy.cc
```diff
@@ -171,15 +172,28 @@ static bool extractIsStrongSignal(ncclGinProxyGfd_t* gfd) {
   return gfd->qword[ncclGinProxyGfdSignalVal].signalVal.isStrongSignal != 0;
 }
 
+static bool extractAggregate(ncclGinProxyGfd_t* gfd) {
+  return gfd->qword[ncclGinProxyGfdHeaderExt].headerExt.aggregate != 0;
```
kgioioso: If the device code is old, this will be an unset field, and could return any random bit.

Please bump the GFD version, and follow the same pattern as `extractIsStrongSignal`, which checks the version

src/include/nccl_device/gin/proxy/gin_proxy.h
```diff
@@ -110,14 +114,16 @@ __device__ __forceinline__ void buildGfd(ncclGinProxyGfd_t* gfd, ncclGinProxyOp_
...
-                                         size_t signalOff, bool isStrongSignal = false) {
+                                         size_t signalOff, bool isStrongSignal = false, bool aggregate = false) {
   for (int i = 0; i < ncclGinProxyGfdQwords; i++) {
     gfd->qword[i].flag.v = 1;
   }
 
   gfd->qword[ncclGinProxyGfdHeader].header.version = (uint64_t)NCCL_GIN_PROXY_GFD_VERSION;
   gfd->qword[ncclGinProxyGfdHeader].header.size = (uint64_t)size;
   gfd->qword[ncclGinProxyGfdHeaderExt].headerExt.op = (uint16_t)op;
+  // Set before the flush early-return so a reused GFD buffer never carries a stale aggregate bit.
```
kgioioso: nit: I think it's clear that most of the header needs to be populated first. You can remove this comment

src/include/nccl_device/gin/proxy/gin_proxy_device_host_common.h
```diff
@@ -105,14 +105,16 @@ typedef union {
...
+    uint8_t aggregate:1;
+    uint8_t resv:6;
     uint16_t op;
     uint8_t resv2;
     uint32_t resv3;
   } __attribute__((packed)) headerExt;
 } ncclGinProxyQword_t;
 static_assert(sizeof(ncclGinProxyQword_t) == sizeof(uint64_t), "sizeof(ncclGinProxyQword_t) != sizeof(uint64_t)");
 static_assert(NCCL_GIN_PROXY_GFD_VERSION < (1 << 4), "NCCL_GIN_PROXY_GFD_VERSION must be less than 2^4");
+static_assert(NCCL_GIN_PROXY_GFD_VERSION >= 2, "GFD headerExt.aggregate bit requires GFD version >= 2");
```
kgioioso: I'm not sure this will do anything. We need a versioned GFD to guard against previously-compiled device code. This code will always have `NCCL_GIN_PROXY_GFD_VERSION >= 2`

src/plugin/rma/rma_v14.cc
```diff
@@ -6,16 +6,80 @@
...
+}
+
+static ncclResult_t ncclRma_v14_createContext(void* collComm, ncclRmaConfig_v15_t* config, void** rmaCtx) {
+  ncclRmaConfig_v14_t config_v14;
+  memset(&config_v14, 0, sizeof(config_v14));
+  config_v14.nContexts = config->nContexts;
+  config_v14.trafficClass = config->trafficClass;
+  config_v14.rankStride = config->rankStride;
+  return ncclRma_v14->createContext(collComm, &config_v14, rmaCtx);
+}
```
kgioioso: do we need this? I thought the signatures of these functions didnt change in v15

src/plugin/rma/rma_v14.cc
```diff
@@ -6,16 +6,80 @@
...
+  (void)optFlags;
+  return ncclRma_v14->iput(rmaCtx, context, srcOff, srcMhandle, size, dstOff, dstMhandle, rank, request);
+}
+
+// Drop optFlags and isStrongSignal (v14 signals are always strong).
+static ncclResult_t ncclRma_v14_iputSignal(void* rmaCtx, int context, uint64_t srcOff, void* srcMhandle, size_t size,
+                                           uint64_t dstOff, void* dstMhandle, uint32_t rank, uint64_t signalOff,
+                                           void* signalMhandle, uint64_t signalValue, uint32_t signalOp,
+                                           bool isStrongSignal, uint32_t optFlags, void** request) {
+  (void)isStrongSignal;
```
kgioioso: this should not be voided. It is used by v14.

kgioioso (commented):
Thanks for the PR. I like the overall idea. 

I'm a bit concerned with the interaction between aggregate requests and proxy progress loop. In particular, if the proxy progress loop moves on to another iteration (`ncclGinProxyProgress` returns), I believe we should set the aggregate flag to false, possible overriding the user's request. 

Also, I think we should consider tweaking the batch logic to only process >1 GFD if aggregate is set.

So, I'm thinking `ncclGinProxyProgress` should look something like this:
```
if (int p = 0; p < pollBatch; p+) {
  bool isAggregate;
  if (!proxyGinPollGfd) break;
  bool isLastInBatch = (p == pollBatch - 1);
  // overrides the user's aggregate flag if isLastInBatch is true
  // Updates isAggregate with the user's aggregate flag
  proxyGinProcessGfd(..., &isAggregate, isLastInBatch);
  
 if (!isAggregate) break;
}
```
Basically, aggregate decides whether we should process more GFDs. The pollBatch is used to enforce fair-sharing across devComms/connections.

What do you think?

kgioioso:
Thinking a bit more, I see what you're trying to do with the lookahead. So I think we need:
aggregate = (userSetAggregateFlag || nextGfdAvailable) && !isLastInBatch

hershys-aws:
> Thinking a bit more, I see what you're trying to do with the lookahead. So I think we need: aggregate = (userSetAggregateFlag || nextGfdAvailable) && !isLastInBatch

Thanks for the prompt review. I have changed the logic so that aggregate will be set as long as there are ops coming in and it's not the last op in the batch. I bumped the GFD version and placed a runtime check. I kept the `static_assert` there just as a sanity check for the next person who may change the header. Let me know if anything else is needed. Thanks for your time!

kgioioso:
I've mirror this to our internal repo. 

@hershys-aws , for completeness, can you post perf numbers? How much improvement is EFA seeing from this change?

You or @bhasunit can also send it to me over slack if preferred.

kgioioso:
@hershys-aws I discovered another issue. If the next GFD is not ready, but aggregate is set, we will early-return, and the ops will be left hanging until we re-enter Progress. This could make setting aggregate _worse_ than not setting it, which is an issue. 

At this point, I think this design needs a bit more thought. I'm tentatively thinking we should not use the aggregate flag from the device code, and instead only send aggregate to the plugin if the next GFD is already ready (and it's not the last in the batch). But I need to think a bit more. Let me know if you have ideas or feedback.

hershys-aws:
> @hershys-aws I discovered another issue. If the next GFD is not ready, but aggregate is set, we will early-return, and the ops will be left hanging until we re-enter Progress. This could make setting aggregate _worse_ than not setting it, which is an issue.
> 
> At this point, I think this design needs a bit more thought. I'm tentatively thinking we should not use the aggregate flag from the device code, and instead only send aggregate to the plugin if the next GFD is already ready (and it's not the last in the batch). But I need to think a bit more. Let me know if you have ideas or feedback.

Sorry about that, I didn't see this in the testing when running on our hardware. I talked to @bhasunit  and we decided the simplest thing at this point would be to go with the approach you have mentioned. My next push will only pass the flag to the plugin if we see the next GFD ready. Going to do this in `proxyGinProcessGfd`:

```
    // BEFORE
    bool aggregate = (extractAggregate(gfd) || isGfdAvailable(hostGpuCtx, targetRank)) && !isLastInBatch;
    // AFTER
    bool aggregate = isGfdAvailable(hostGpuCtx, targetRank) && !isLastInBatch;
```

and stripping out the plumbing which was placed in `src/include/nccl_device/gin/proxy/gin_proxy.h` and `src/include/nccl_device/gin/proxy/gin_proxy_device_host_common.h`.

kgioioso:
> Sorry about that, I didn't see this in the testing when running on our hardware. 

Just to clarify, this was a hypothetical regression, not a measured regression. But I think it was important enough to warrant a change.

> I talked to @bhasunit and we decided the simplest thing at this point would be to go with the approach you have mentioned. 

This looks good. Is there a perf difference between the current approach and actually looking at the device-side aggregate requests? If so, we can re-visit in the next release. But I'm hoping this approach is just as good

kgioioso:
@bhasunit @hershys-aws This was merged. I changed the name of the opt flag in this [commit](https://github.com/NVIDIA/nccl/commit/f6be6cfa070074baed121b76db7e2bb188633a25).

PR #2265, opened 2026-06-29
net_ib: add missing break in printIbWcStatusHint switch

EylonKrause:
## Description

`printIbWcStatusHint` is a `switch` on the IB work-completion status whose `IBV_WC_LOC_PROT_ERR`, `IBV_WC_WR_FLUSH_ERR` and `IBV_WC_RETRY_EXC_ERR` cases lack a `break`. Because they fall through, a single completion error prints its own hint plus the hints of every case below it — e.g. a local protection error (`IBV_WC_LOC_PROT_ERR`) also advises that "NCCL_IB_TIMEOUT is set too short", which is unrelated and misleading when debugging.

## Related Issues

None.

## Changes & Impact

- `src/transport/net_ib/common.h` (`printIbWcStatusHint`): add `break;` to each of the three cases so only the hint matching the actual completion status is printed. Diagnostic output only; no behavior change beyond the log text. Non-breaking.

## Performance Impact

None.

### Testing

- The fall-through is evident from the missing `break;`; the fix limits each completion status to its own hint.

### DIFF
--- a/src/transport/net_ib/common.h
+++ b/src/transport/net_ib/common.h
@@ -616,14 +616,17 @@ static inline void printIbWcStatusHint(int status) {
          "HINT: In many cases this error occurs when ACS is enabled which would break GPU Direct RDMA protocol.");
     INFO(NCCL_NET, "HINT: To confirm and fix the problem, you can set NCCL_NET_GDR_LEVEL=0 to disable ACS following "
                    "vendor documentation.");
+    break;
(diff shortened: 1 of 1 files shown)


EylonKrause:

EylonKrause:
Closing: superseded by the v2.31 rewrite of `printIbWcStatusHint` (each case now ends with an explicit `return;`, so the fall-through this PR fixed no longer exists on `master`).

Issue #374, opened 2020-08-18
Is NCCL creating blocking threads?


frankseide:
We are seeing a PyTorch process at CPU load of 400%. We use no multi-threading in the code, so CPU load should be at best 100%. Does NCCL create threads that busy-wait? Our code creates multiple communication groups.

sjeaugey:
Yes, NCCL creates one thread per communicator per rank for network communication. Threads should only be active when there is ongoing NCCL communication, meaning outside of NCCL regions they should be sleeping.

frankseide:
Thanks. Is it possible that they busy-wait, depending on an option? E.g. if CUDA is configured to busy-wait when synchronizing with the GPU, might NCCL obey that setting as well? I am just fishing, we don't have a clear handle where the CPU usage comes from. Unfortunately, we don't have sudo access on the lab machines where we observe it...

sjeaugey:
Yes, it is expected that they busy wait, so seeing 100% CPU for each NCCL thread is expected. Note however that when they have nothing to do, they will call sched_yield() inside the loop, so if anything else needs to use the CPU, their CPU usage should shrink.

frankseide:
We are also seeing n x 100% CPU when there are other processes trying to run (we do a lot of text pre-processing on the fly during model training, which runs in parallel processes). We are seeing slowdowns.

Is there a way to control the scheduling of these threads? E.g. is there a way to place the threads all on the same core, adjust their priority, or switch from sched_yield() (which is a form of self-rolled scheduler) to waiting for a wake-up event?

sjeaugey:
Since those threads use 100% CPU, the kernel might wrongly assume that they should keep a dedicated core. So if your other threads do not use 100% CPU, they might end up sharing a core and NCCL threads might keep a dedicated one.

So you could try to manually pin threads to specific cpus to avoid that (forcing them to share a given CPU). You could also move all NCCL threads to a single core. But in all cases, keep in mind that it could also affect the NCCL performance, especially if NCCL threads share the CPU with other processes which take the CPU for a full 100ms timeslice. If you're using RDMA, having a single core for all NCCL communication might be ok, but if you're using TCP/IP, you might want to use more than one core.

[...]

PR #2277, opened 2026-07-08
Add profiler proxy poll backoff

pawelchcki:
## Description

Add an explicit proxy poll backoff for profiler kernel-channel timing waits.

This targets a profiler-specific manifestation of the proxy busy-wait concern discussed in #374. Profiler kernel-channel proxy ops wait for GPU-written start/completion counters; when those counters have not advanced, the proxy can repeatedly poll only those memory locations while staying runnable, which may show up as high CPU usage for the proxy progress thread.

To reproduce, run a two-rank single-node NCCL collective workload with a profiler plugin loaded and kernel-channel profiling enabled:

```bash
NCCL_PROFILER_PLUGIN=/path/to/libnccl-profiler-example.so \
NCCL_PROFILE_EVENT_MASK=64 \
NCCL_PROFILE_DUMP_FILE=/tmp/nccl-prof \
mpirun -np 2 ./build/all_reduce_perf -b 8 -e 1G -f 2 -g 1
```

`NCCL_PROFILE_EVENT_MASK=4095` also enables this path.

## Related Issues

Possibly related to #374 as the symptoms are identical

## Changes & Impact

- Add `pollDelayUsec` to `ncclProxyArgs` so progress callbacks can request a bounded poll delay separately from the existing `idle` signal.
- Have the main proxy progress loop sleep for the smallest requested positive delay only when all active work is idle and no new proxy ops were appended.
- Have the profiler transport request `NCCL_PROFILER_KERNEL_CH_POLL_SLEEP_USEC`, defaulting to 1 usec, while it is waiting for kernel-channel counters.
- Keep profiler kernel-channel start/stop tracking aligned with the existing `sub->nsteps` counter convention.

This does not change the public NCCL API. The new behavior is limited to idle profiler kernel-channel polling.

## Performance Impact

Expected impact is lower proxy progress thread CPU usage while profiler kernel-channel polling is idle, with a small potential profiling-latency tradeoff controlled by `NCCL_PROFILER_KERNEL_CH_POLL_SLEEP_USEC`.

### DIFF
--- a/src/include/proxy.h
+++ b/src/include/proxy.h
@@ -210,6 +210,7 @@ struct ncclProxyArgs {
   int nPeers;
 
   int idle;
+  int64_t pollDelayUsec;
 
   // Element linking
   struct ncclProxyArgs* next;
--- a/src/proxy.cc
+++ b/src/proxy.cc
@@ -1014,7 +1014,11 @@ void* ncclProxyProgress(void* proxyState_) {
       }
       if (added == 0) {
         if (idle && pollDelayUsec > 0) {
-          std::this_thread::sleep_for(std::chrono::microseconds(pollDelayUsec));
+          struct ncclProxyOpsPool* pool = state->opsPool;
+          std::unique_lock<std::mutex> lock(pool->mutex);
+          pool->cond.wait_for(lock, std::chrono::microseconds(pollDelayUsec), [&]() {
+            return pool->nextOps != -1 || state->stop != 0;
+          });
         } else {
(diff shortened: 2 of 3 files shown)


pawelchcki:
Simple benchmark based on all_reduce_perf

image

armratner:
https://github.com/NVIDIA/nccl/commit/7a4bde51418b98721cd438fc3748625251f04e52

The above commit should've fixed this - are you still seeing this with latest NCCL?

PR #2279, opened 2026-07-09
GIN: multi-thread GIN proxy with per-thread endpoint assignment

mozarhua:
## Description
                           
This PR introduces **Multiple GIN progress threads** (`GIN_PROXY_NTHREADS`) to distribute GIN connections across N progress threads.
                                                                                                                      
## Changes & Impact

Spawn `GIN_PROXY_NTHREADS` (default 1) progress threads, distributing the GIN connections across them via round-robin (thread `t` owns connections `t`, `t+T`, `t+2T`, ...).

Synchronization uses a `std::shared_timed_mutex` with an atomic state variable (not started, running, write pending, stop). Proxy threads acquire a shared lock to iterate the `devComms` list; the main thread sets write-pending (causing readers to back off), acquires the exclusive lock to mutate the list, then resumes readers.

Thread and connection count resolution:
- If ginCommCount < nthreads, increase ginCommCount to nthreads.
- `NCCL_GIN_MAX_CONNECTIONS` (4) remains the hard cap for both thread and connection counts.

`ncclGinDevCommFree` holds the exclusive lock only across the list unlink (not the slow `destroyContext()` calls): once released, contexts are destroyed with the workers still progressing the rest of the list.

GIN_PROXY_NTHREADS=1 (default) preserves original single-thread behavior.

Device code unchanged — context→connection mapping remains unchanged.

### DIFF
--- a/src/gin/gin_host.cc
+++ b/src/gin/gin_host.cc
@@ -40,35 +40,55 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
   return ncclSuccess;
 }
 
-void* ncclGinProgress(struct ncclGinState* ginState_) {
-  struct ncclGinState* ginState = (struct ncclGinState*)ginState_;
+// Writer helpers: signal readers to back off, acquire exclusive lock, mutate, release, resume.
+static void ginProgressWriteLock(struct ncclGinState* ginState) {
+  ginState->progressState.store(ncclGinProgressWritePending, std::memory_order_release);
+  ginState->devCommRwMutex.lock();
+}
+
+static void ginProgressWriteUnlock(struct ncclGinState* ginState) {
+  ginState->devCommRwMutex.unlock();
+  ginState->progressState.store(ncclGinProgressRunning, std::memory_order_release);
+}
+
+// Per-thread progress worker. Thread t owns GIN connections t, t+proxyNthreads, t+2*proxyNthreads, ...
+// across all devComms.
+void* ncclGinProgress(struct ncclGinState* ginState, int threadIdx) {
   if (ncclOsCpuCount(ginState->cpuAffinity)) {
     ncclOsSetAffinity(ginState->cpuAffinity);
   }
   while (1) {
-    std::unique_lock<std::mutex> lock(ginState->mutex);
-    if (ginState->proxyThreadStopSignal) return NULL;
-    struct ncclGinStateDevComm* dc = ginState->devComms;
-    while (dc) {
-      struct ncclGinBackendState* backend = &ginState->backends[dc->backendIndex];
-      for (int commIdx = 0; commIdx < backend->ginCommCount; commIdx++) {
-        if (dc->devHandles[commIdx]->needsProxyProgress) {
-          ncclResult_t ret = backend->ncclGin->ginProgress(dc->ginCtx[commIdx]);
-          if (ret != ncclSuccess) {
-            COMPILER_ATOMIC_STORE(&ginState->asyncResult, ret, std::memory_order_release);
-            INFO_LOC(NCCL_ALL, "-> %d [GIN Progress Thread]", ret);
-            return NULL;
+    int state = ginState->progressState.load(std::memory_order_acquire);
+    if (state == ncclGinProgressStop) return NULL;
+    // Back off while the main thread is mutating the devComms list.
+    if (state == ncclGinProgressWritePending) {
+      std::this_thread::yield();
+      continue;
+    }
+    {
+      std::shared_lock<std::shared_timed_mutex> rlock(ginState->devCommRwMutex);
+      struct ncclGinStateDevComm* dc = ginState->devComms;
+      while (dc) {
+        struct ncclGinBackendState* backend = &ginState->backends[dc->backendIndex];
+        for (int commIdx = threadIdx; commIdx < backend->ginCommCount; commIdx += ginState->proxyNthreads) {
+          if (dc->devHandles[commIdx]->needsProxyProgress) {
+            ncclResult_t ret = backend->ncclGin->ginProgress(dc->ginCtx[commIdx]);
+            if (ret != ncclSuccess) {
+              COMPILER_ATOMIC_STORE(&ginState->asyncResult, ret, std::memory_order_release);
+              INFO_LOC(NCCL_ALL, "-> %d [GIN Progress Thread %d]", ret, threadIdx);
+              return NULL;
+            }
           }
         }
+        dc = dc->next;
       }
-      dc = dc->next;
     }
-    lock.unlock();
     std::this_thread::yield();
   }
 }
 
 NCCL_PARAM(GinNconnections, "GIN_NCONNECTIONS", -2);
+NCCL_PARAM(GinProxyNthreads, "GIN_PROXY_NTHREADS", 1);
 
 ncclResult_t ncclGinConnectOnce(struct ncclComm* comm) {
   ncclTeam_t ginTeam;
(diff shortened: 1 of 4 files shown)


kgioioso:
Thanks for the submission.

My guess is you forked off some version that did not have my proxy-thread-state cleanup. Apologies if you had to resolve merge conflicts. This MR re-introduces a lot of the complexity I tried to remove, particularly around the condition variable and the state machine (which are both removed on `dev`, see https://github.com/NVIDIA/nccl/commit/9e6d3748f6775547b54546991fc8b1aa8a290bb0)

Can we simplify? My first instinct was to just use one reader-writer mutex. The proxy threads all use shared locks when processing the devComm list, and then the main thread uses a exclusive lock when it updates the list. But, the proxy threads spend most of their time holding the lock, so this may lead to writer starvation.

I think we can get away with 1 mutex, no condition variable, and a single state variable tracking the following states (stored in an atomic):
* not started
* running
* write pending
* stop signal

To update the devCommList, the writer:
* atomically sets state to write pending
* acquires the (exclusive) mutex, updates the devCommList, frees the mutex
* atomically sets state to running

Each proxy thread does while 1:
* if write pending, continue
* acquire (shared) mutex
* iterate over dev comms
* free mutex

What do you think?

mozarhua:
Hi @kgioioso. Thanks for your review! Your proposal sounds good — that's essentially a reader-writer lock with writer priority (the write_pending flag prevents reader starvation of the writer).

I'm assuming you're referring to `std::shared_mutex` (C++17) semantics. Since NCCL targets C++14 ([ref: makefiles/common.mk](https://github.com/NVIDIA/nccl/blob/5067397c2676d5aed50042fc39e5c8ee96eb0027/makefiles/common.mk#L89-L94)) for CUDA versions < 13, I could use `std::shared_timed_mutex` which is available in C++14 and provides the same shared/exclusive locking. Alternatively, `pthread_rwlock_t` or an atomic-
based SWMR spinlock would also work. Any preference?

kgioioso:
> I'm assuming you're referring to std::shared_mutex (C++17) semantics

Yes, that's what I had in mind. And you're right that it's too new. Unfortunate.

> Since NCCL targets C++14 ([ref: makefiles/common.mk](https://github.com/NVIDIA/nccl/blob/5067397c2676d5aed50042fc39e5c8ee96eb0027/makefiles/common.mk#L89-L94)) for CUDA versions < 13, I could use std::shared_timed_mutex which is available in C++14 and provides the same shared/exclusive locking. Alternatively, pthread_rwlock_t or an atomic-
based SWMR spinlock would also work. Any preference?

We are trying to move away from `pthread_*` because it is not supported on windows. Let's use `std::shared_timed_mutex` with a comment explaining why we aren't using shared_mutex.

mozarhua:
> Let's use std::shared_timed_mutex with a comment explaining why we aren't using shared_mutex.

Hi @kgioioso . I've updated the PR to use `std::shared_timed_mutex` with the writer-priority pattern you described (atomic state + rwlock). Added a comment in the header explaining the C++14/C++17 choice. Let me know if the approach looks good or if you'd like any adjustments. Thanks!

src/gin/gin_host.cc
```diff
@@ -40,35 +40,64 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
...
+}
+
+static void ginProgressWriteUnlock(struct ncclGinState* ginState) {
+  ginState->devCommRwMutex.unlock();
+  ginState->progressState.store(ncclGinProgressRunning, std::memory_order_release);
+}
+
+// Per-thread progress worker. Thread t owns GIN connections t, t+proxyNthreads, t+2*proxyNthreads, ...
+// across all devComms. proxyNthreads=1 (default) reproduces the original single-thread behavior.
+// needsProxyProgress gates the call per-connection; GDAKI/GPI return false.
```
kgioioso: nit, comments should not refer to how things used to be, or other pieces of the code (GDAKI/GPI)


```suggestion
// Per-thread progress worker. Thread t owns GIN connections t, t+proxyNthreads, t+2*proxyNthreads, ...
// across all devComms.
```

src/gin/gin_host.cc
```diff
@@ -40,35 +40,64 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
...
+//
+// Synchronization with the writer (main thread):
+//   - Writer sets progressState to WritePending; readers skip the iteration.
+//   - Writer acquires exclusive lock on devCommRwMutex, mutates the list, releases.
+//   - Writer sets progressState back to Running.
+//   - Each proxy thread acquires shared lock, iterates, releases.
```
kgioioso: I think most of this is evident from the code. If you feel more comments are necessary, you can add a comment above the `WritePending` if condition

src/gin/gin_host.cc
```diff
@@ -40,35 +40,64 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
...
+    if (state == ncclGinProgressWritePending) {
+      std::this_thread::yield();
+      continue;
+    }
+    {
+      std::shared_lock<std::shared_timed_mutex> rlock(ginState->devCommRwMutex);
+      struct ncclGinStateDevComm* dc = ginState->devComms;
+      while (dc) {
+        struct ncclGinBackendState* backend = &ginState->backends[dc->backendIndex];
+        for (int n = t; n < backend->ginCommCount; n += ginState->proxyNthreads) {
```
kgioioso: nit: revert the loop variable name (`commIdx`)

src/gin/gin_host.cc
```diff
@@ -40,35 +40,64 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
...
+// across all devComms. proxyNthreads=1 (default) reproduces the original single-thread behavior.
+// needsProxyProgress gates the call per-connection; GDAKI/GPI return false.
+//
+// Synchronization with the writer (main thread):
+//   - Writer sets progressState to WritePending; readers skip the iteration.
+//   - Writer acquires exclusive lock on devCommRwMutex, mutates the list, releases.
+//   - Writer sets progressState back to Running.
+//   - Each proxy thread acquires shared lock, iterates, releases.
+//
+void* ncclGinProgress(struct ncclGinState* ginState, int t) {
```
kgioioso: nit

```suggestion
void* ncclGinProgress(struct ncclGinState* ginState, int threadIdx) {
```

src/gin/gin_host.cc
```diff
@@ -40,35 +40,64 @@ ncclResult_t ncclGetRailedGinType(struct ncclComm* comm, ncclGinType_t* ginType)
...
+      struct ncclGinStateDevComm* dc = ginState->devComms;
+      while (dc) {
+        struct ncclGinBackendState* backend = &ginState->backends[dc->backendIndex];
+        for (int n = t; n < backend->ginCommCount; n += ginState->proxyNthreads) {
+          if (dc->devHandles[n]->needsProxyProgress) {
+            ncclResult_t ret = backend->ncclGin->ginProgress(dc->ginCtx[n]);
+            if (ret != ncclSuccess) {
+              COMPILER_ATOMIC_STORE(&ginState->asyncResult, ret, std::memory_order_release);
+              // Stop all progress threads on first error; reported via ncclCommGetAsyncError.
+              ginState->progressState.store(ncclGinProgressStop, std::memory_order_release);
```
kgioioso: Before, the main thread was the only thread changing the state. Now, multiple threads can change the state, and we need to worry about different threads overwriting other values. Seems like we can go from Stop -> WritePending -> Running, which is very confusing.

Is there a harm in keeping the other threads running?

src/gin/gin_host.cc
```diff
@@ -355,26 +398,34 @@ ncclResult_t ncclGinDevCommSetup(struct ncclComm* comm, struct ncclDevCommRequir
...
+
+    if (!needsStart) ginProgressWriteLock(ginState);
+
     struct ncclGinStateDevComm* last = ginState->devComms;
     if (last) {
       while (last->next) last = last->next;
       last->next = ginStateDevComm;
     } else {
       ginState->devComms = ginStateDevComm;
     }
```
kgioioso, line 407: nit, this seems a bit error prone. Let's move the Unlock closer to the Lock
```suggestion
    if (!needsStart) ginProgressWriteLock(ginState);

struct ncclGinStateDevComm* last = ginState->devComms;
if (last) {
while (last->next) last = last->next;
last->next = ginStateDevComm;
} else {
ginState->devComms = ginStateDevComm;
}

if (!needsStart) ginProgressWriteUnlockLock(ginState);
```

src/gin/gin_host.cc
```diff
@@ -388,28 +439,32 @@ ncclResult_t ncclGinDevCommSetup(struct ncclComm* comm, struct ncclDevCommRequir
...
     if (dc->devHandles[0]->handle == devComm->ginHandles[0]) break;
     prevDc = dc;
     dc = dc->next;
   }
-
-  std::unique_lock<std::mutex> lock(ginState->mutex);
-  // Remove from linked list
   if (prevDc) prevDc->next = dc->next;
   else ginState->devComms = dc->next;
```
kgioioso, line 450: again, we should move the lock closer to the unlock. The lock is not needed to read the list, so you can move the lock/unlock to this critical section. This has the added bonus that there are no early-exits where we unlock.

kgioioso (commented):
Overall looks good, the only thing remaining is style nits.

kgioioso:
Please also post or slack perf numbers. With whatever benchmark you are using, what perf do you see for 1,2, and 4 threads.

src/gin/gin_host.cc
mozarhua: Removed unnecessary comments.

src/gin/gin_host.cc
mozarhua: Removed unnecessary comments and added a one-liner above the `WritePending` if condition.

src/gin/gin_host.cc
mozarhua: Good point — It's not harmful to keep other threads running. (An application hang or timeout will trigger and check the error.) Removed the Stop from the error path.

src/gin/gin_host.cc
mozarhua: Moved the Unlock / Lock closer to each other, only for the list modification.

src/gin/gin_host.cc
mozarhua: Moved the Unlock / Lock closer to each other, only for the list modification.

mozarhua:
> Please also post or slack perf numbers. With whatever benchmark you are using, what perf do you see for 1,2, and 4 threads.

Will slack perf numbers offline. Let me know if any further comments. Thanks!

kgioioso:
This was merged and will be mirrored to GitHub shortly. 

@mozarhua I changed the implementation from the state machine to separate state variables (stopSignal, threadsCreated, writePending). The state machine as it was written had an issue that was causing our internal CI to fail: A devComm that does not need proxy progress is created and then destroyed. Then, a devComm that does need proxy progress is created, but all GIN operations hang. The issue was the destroying the first DevComm called WriterUnlock, which moved the state to Running, even though no proxy threads were running. Then, the proxy threads were not created for the second DevComm.

Thanks for your contribution. The perf numbers are quite encouraging.

mozarhua:
> I changed the implementation from the state machine to separate state variables (stopSignal, threadsCreated, writePending). 

@kgioioso Thank you for catching that edge case and fixing it! Appreciate the thorough review and the improvement to separate state variables. Happy to help with any follow-up work.

PR #2284, opened 2026-07-13
Mixed Precision Reduce Scatter

summerinsect:
Implement a specific reduce scatter kernel for bf16 input and fp32 accumulation. 

Implement correctness and performance tests.


summerinsect:
Oh I don't want to merge it into the official repository. It seems to be a mistake.

PR #2290, opened 2026-07-15
Fix CE synchronization for two-rank LSA teams

ngoyal2707:
## Why is this change needed?

Copy-engine collectives can hang or fail with `invalid resource handle` on two-rank LSA (sub)communicators. 

Main reason is: `ncclMemOpSync` selects multicast synchronization whenever NVLS is generally available (`src/ce_coll.cc`), while symmetric-kernel initialization only enables LSA multicast for teams larger than two ranks (`src/sym_kernels.cc`).

For a two-rank communicator, the first CE collective therefore requests a multicast team that initialization deliberately did not establish. The team is created lazily in `symTeamObtain`, whose contract requires a team barrier afterward (`src/dev_runtime.cc`) that the CE path does not perform: the creating rank's 4-byte multicast flag write can execute before the peer rank has imported the multicast handle and bound its memory, so the write is lost and the peer's `CU_STREAM_MEM_OP_WAIT_VALUE_32` waits forever. On multi-node NVLink (MNNVL) systems the fabric-handle import latency makes this a deterministic hang rather than a rare race.

## Summary

- Use a shared `ncclCeUseMCSync` predicate matching the symmetric-kernel LSA multicast capability gate.
- Keep CE payload copies enabled for two-rank teams, but synchronize their ready/completion flags with unicast peer writes.
- Report the effective multicast/unicast strategy consistently in tuning logs and profiler events.

## Repro

nccl-tests on two MNNVL-connected nodes with 4 GPUs each. 

```
mpirun -H <nodeA>:4,<nodeB>:4 -np 8 -x NCCL_TESTS_SPLIT=MOD4 \
-x NCCL_CTA_POLICY=ZERO \
./build/all_gather_perf -b 1M -e 32M -f 2 -g 1 -R 2
```

With `NCCL 2.30.7` this hangs in the first warmup iteration (no result rows are ever printed). Dropping `-x NCCL_CTA_POLICY=ZERO` (kernel path) completes normally. With this fix, both commands complete.

### DIFF
--- a/src/ce_coll.cc
+++ b/src/ce_coll.cc
@@ -130,6 +130,10 @@ bool ncclCeAvailable(struct ncclComm* comm, ncclFunc_t coll, int /*ncclDevRedOp_
   return true;
 }
 
+bool ncclCeUseMCSync(struct ncclComm* comm) {
+  return comm->nvlsSupport && ncclTeamLsa(comm).nRanks > 2 && !comm->p2pCrossClique;
+}
+
 ncclResult_t ncclPrepMCSync(struct ncclComm* comm, bool isComplete, CUstreamBatchMemOpParams* batchParams,
                             size_t* opIdx, cudaStream_t stream) {
   ncclResult_t ret = ncclSuccess;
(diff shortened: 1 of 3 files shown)


ngoyal2707:
nvm looks like its fixed here on dev
https://github.com/NVIDIA/nccl/commit/430afbcef0d0dbe4bac8de317cf4cedd50ffefa2

PR #2291, opened 2026-07-15
Fix plugin-specific topology device indices

kzlxd:
## Summary

Store device indices independently for each network plugin API instead of sharing a single topology `dev` attribute.

The topology XML now uses:

| Plugin API | Device index attribute |
|------------|------------------------|
| NET        | `dev`                  |
| GIN        | `gdev`                 |
| RMA        | `rdev`                 |
| CollNet    | `cdev`                 |

Legacy topology XML files containing only `dev` remain supported through fallback logic.

## Problem

NET, GIN, RMA, and CollNet share the same physical `<net>` XML nodes, matched by device name. However, their device indices belong to separate plugin-local namespaces and are not guaranteed to use the same ordering.

This is visible on B300 systems using bonded Mellanox devices:

- The external NET plugin enumerates devices in bond-name order.
- Internal GIN/RMA devices may be reordered by PCI topology.
- Each plugin previously overwrote the same XML `dev` attribute.
- NET is imported last, so GIN/RMA topology nodes received NET plugin indices.
- GIN then interpreted the NET index in its own namespace and selected a different bond device.

The physical topology node was correct, but the device index passed to the plugin was incorrect.

## Changes

- Store NET, GIN, RMA, and CollNet indices in separate XML attributes.
- Make GIN and RMA topology nodes read `gdev` and `rdev`.
- Preserve compatibility with legacy XML by falling back to `dev`.
- Store a separate CollNet device index in NET topology nodes.
- Return the CollNet index when resolving devices for CollNet graphs.
- Query flush properties through the corresponding NET or CollNet plugin.
- Include RMA when updating and pruning virtual NIC topology entries.
- Increase the XML attribute capacity to accommodate the additional indices.
- Log all plugin-specific indices for topology diagnostics.

## Validation

- Validated on a B300 bond configuration where NET and GIN device enumeration orders differ.
- Confirmed topology-selected devices resolve to the expected bond interfaces.
- Confirmed performance is restored after eliminating the cross-plugin index overwrite.
- Patch application check passed.

## Compatibility

Existing topology XML files containing only `dev` continue to work. The new attributes are used when available and otherwise fall back to the legacy NET device index.

### DIFF
--- a/src/graph/paths.cc
+++ b/src/graph/paths.cc
@@ -572,11 +572,12 @@ ncclResult_t ncclTopoIsGdrAvail(struct ncclTopoSystem* system, int rank, bool* a
 NCCL_PARAM(NetForceFlush, "NET_FORCE_FLUSH", 0);
 
 // Based on the system topology, determine whether an explicit iflush is needed on the GDR recv path.
-ncclResult_t ncclTopoNeedFlush(struct ncclComm* comm, int64_t netId, int netDev, int rank,
+ncclResult_t ncclTopoNeedFlush(struct ncclComm* comm, int64_t netId, int netDev, int rank, bool collNet,
                                enum ncclTopoFlushType* flush) {
   *flush = ncclTopoFlushAlways;
   ncclNetProperties_t props;
-  NCCLCHECK(comm->ncclNet->getProperties(netDev, &props));
+  if (collNet) NCCLCHECK(comm->ncclCollNet->getProperties(netDev, &props));
+  else NCCLCHECK(comm->ncclNet->getProperties(netDev, &props));
   if (props.forceFlush == 1 || ncclParamNetForceFlush()) return ncclSuccess;
   int g;
   struct ncclTopoSystem* system = comm->topo;
--- a/src/graph/search.cc
+++ b/src/graph/search.cc
@@ -1411,7 +1411,7 @@ ncclResult_t ncclTopoGetNetDev(struct ncclComm* comm, int rank, struct ncclTopoG
     } else {
       NCCLCHECK(getNvlsNetDev(comm, graph, channelId, &netId));
     }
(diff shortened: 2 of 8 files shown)


kgioioso:
Good find. One issue with this MR is that in 2.31 (`dev` branch on GitHub) GIN can have multiple backends, each with its own set of indices. This will lead to _many_ indices in the XML, which is a bit confusing.

One idea we have is to keep the XML as is, and then re-discover the per-plugin indices once the XML is converted to a system. I'm working on this now to determine feasibility, will keep you updated. 

Regardless, thanks for reporting. We plan to fix this for the next release.

kzlxd:
> Good find. One issue with this MR is that in 2.31 (`dev` branch on GitHub) GIN can have multiple backends, each with its own set of indices. This will lead to _many_ indices in the XML, which is a bit confusing.
> 
> One idea we have is to keep the XML as is, and then re-discover the per-plugin indices once the XML is converted to a system. I'm working on this now to determine feasibility, will keep you updated.
> 
> Regardless, thanks for reporting. We plan to fix this for the next release.

@kgioioso  Thanks, that makes sense.

I checked the current dev branch and confirmed that localGinDevs is computed
once and then reused for every entry in ginState.backends[]. Therefore a
single gdev attribute would only solve the index mapping for one backend.

A backend-local resolver after XML-to-system conversion looks cleaner. The
mapping should preferably use a stable physical identity such as PCI BDF +
port, with GUID/name as fallback, and resolve the device index inside the
backend loop before listen/getProperties/connect.

One case to consider is that different backends may expose different device
subsets. In that case topology selection should be filtered per backend, or
fail explicitly when a selected physical NIC cannot be mapped.

I am happy to validate the resulting patch on the B300 bond setup where the
NET and GIN enumeration orders differ.

PR #2296, opened 2026-07-16
EFA GDA: Implement timeout in flushImplMode

bhasunit:
## Description
Implement the timeout logic in the Flush polling loop following the GDAKI pattern:
- Start clock64() timer inside the thread_rank()==0 block
- Check elapsed cycles against timeoutCycles in each polling iteration
- Return ncclTimeout on expiry
- On abort, return ncclInProgress (internal sentinel) to trigger early exit via goto, then map back to ncclSuccess before returning to caller
- Use goto to exit the if-block on early return (timeout or abort), then broadcast thread 0's result to all coop participants via shared memory (safe for all coop types including ncclCoopAny)
- Change flushImplMode/flushImpl return type from void to ncclResult_t

### DIFF
--- a/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
+++ b/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
@@ -605,14 +605,18 @@ NCCL_DEVICE_INLINE static void putValueImpl(ncclGinCtx ctx, Coop coop, int peer,
 /* ── flushImplMode: mode-templated Flush implementation ───────────── */
 
 template <bool HasTimeout, ncclGinResourceSharingMode mode, typename Coop>
-NCCL_DEVICE_INLINE static void flushImplMode(ncclGinCtx ctx, Coop coop, cuda::memory_order ord, uint32_t* abortFlag,
-                                             uint64_t timeoutCycles) {
+NCCL_DEVICE_INLINE static ncclResult_t flushImplMode(ncclGinCtx ctx, Coop coop, cuda::memory_order ord,
+                                                     uint32_t* abortFlag, uint64_t timeoutCycles) {
   (void)ord;
-  (void)timeoutCycles;
-  // TODO: Implement timeout
+  if NCCL_IF_CONSTEXPR (!HasTimeout) (void)timeoutCycles;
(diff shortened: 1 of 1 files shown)


src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
```diff
@@ -647,24 +657,32 @@ NCCL_DEVICE_INLINE static void flushImplMode(ncclGinCtx ctx, Coop coop, cuda::me
...
   }
-  coop.sync();
+done:
+  // Broadcast result from thread_rank()==0 to all participants.
+  // Use shared memory — safe for all coop types including ncclCoopAny.
+  {
+    __shared__ int sharedResult;
+    if (coop.thread_rank() == 0) sharedResult = (int)result;
+    coop.sync();
+    result = (ncclResult_t)sharedResult;
```
kgioioso, line 671: FYI, we opted for no broadcast (threads may return different values). Reasoning: the broadcast is reasonably expensive, so if users want consistent values they should do it themselves.

src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
```diff
@@ -647,24 +657,32 @@ NCCL_DEVICE_INLINE static void flushImplMode(ncclGinCtx ctx, Coop coop, cuda::me
...
+      if (result != ncclSuccess) goto done;
     }
   }
-  coop.sync();
+done:
+  // Broadcast result from thread_rank()==0 to all participants.
+  // Use shared memory — safe for all coop types including ncclCoopAny.
+  {
+    __shared__ int sharedResult;
+    if (coop.thread_rank() == 0) sharedResult = (int)result;
```
pakmarkthub, line 669: What will happen if two warps of the same CTA enter this function concurrently? Now, there are two threads with `coop.thread_rank() == 0`. Don't we have a data race in this case?

xiaofanl-nvidia:
https://github.com/NVIDIA/nccl/commit/8bc8b18e2b6eac367ad4dd2ab44faf4772c431f5 merged. Closing.

PR #2301, opened 2026-07-21
gin: return rounded-up context count to application via devComm

rauteric:
## Description

devComm->ginContextCount (returned to application) was set to the requested context count before being rounded up, so the rounded-up value of contexts actually allocated (a multiple of ginCommCount) was never reported back to the application. Move the assignment after ROUNDUP so devComm->ginContextCount reflects the rounded-up allocation, allowing the application to use the additional contexts


kgioioso:
Merged. Will be mirrored shortly. Thanks for your contribution.

Issue #2303, opened 2026-07-23
[Issue]: OW_LATENCY dispatch/combine receives time out on GB200/GB300 NVL72


Oseltamivir:
## NCCL EP LOW_LATENCY dispatch times out on GB200/GB300 NVL72 (HIGH_THROUGHPUT and classic-NVLink LL both pass)

**Setup:** GB200/GB300 NVL72, 4 GPUs/tray (also reproduces world=8 across two trays), one rank/GPU. DeepSeek-V3-like workload: 256 routed experts, top-k 8, hidden 7168, BF16.

```python
import nccl.core, nccl.ep
comm = nccl.core.Communicator.init(nranks, rank, unique_id)
cfg = nccl.ep.GroupConfig(algorithm=nccl.ep.Algorithm.LOW_LATENCY, num_experts=256,
    max_dispatch_tokens_per_rank=256, max_recv_tokens_per_rank=256 * nranks,
    max_token_bytes=7168 * 2, alloc=alloc_cfg)
grp = nccl.ep.Group.create(comm, cfg)     # succeeds; windows register fine
h = grp.create_handle(layout=nccl.ep.Layout.EXPERT_MAJOR, topk_idx=idx, stream=s)
h.dispatch(DispatchInputs(x), DispatchOutputs(...), stream=s)  # receives time out
s.synchronize()                                                # raises 719
```

`NCCL_CUMEM_ENABLE=1`; fails with `NCCL_MNNVL_ENABLE=1` and `=0` alike (verified value reaches ranks via `os.environ`, since baked container env vars can silently override job-level exports).

**Control experiments:** Swapping to `algorithm=HIGH_THROUGHPUT` (+ `Layout.FLAT`, unweighted combine) passes on identical world/fabric/payload (world=8 and 16 over MNNVL). The identical LL sequence passes on H100/H200/B200/B300 (world=8, classic intra-node NVLink, x86_64).

| Setup | P2P transport | HT | LL |
|---|---|---|---|
| H100/H200/B200/B300, world=8, 1 node, x86_64 | classic intra-node NVLink | ✅ | ✅ |
| GB200/GB300, world=8, 2 trays, `MNNVL_ENABLE=1` | P2P/MNNVL | ✅ | ❌ timeout |
| GB200, world=4, 1 tray, `MNNVL_ENABLE=1` | P2P/MNNVL | ✅ | ❌ timeout |
[...]

Issue #12, opened 2016-02-26
What does out-of-place and in-place mean in the test routines?


linnanwang:
Hello,

What does out-of-place and in-place mean in the test routines? Thank you.

cliffwoolley:
An in-place operation uses the same buffer for its output as was used to
provide its input.  An out-of-place operation has distinct input and output
buffers.

Hope this helps,
Cliff

linnanwang:
Hello Cliff,

Thanks for the reply. This is very helpful. However, what if I can to see the difference between tree-based broadcast and ring-based broadcast? Is there such benchmarks so that I can compare with? Thanks.

cliffwoolley:
NCCL's design allows it to choose among many algorithms for each
collective, but this early version of NCCL only thus far implements
ring-based (not tree-based or other) algorithms.

Thanks,
Cliff

Issue #11, opened 2016-02-25
occasional crashes when using more than one comm per GPU


amithr1:
Hi All,

I have noticed crashes when I overload a device with more than one nccl comm. For example,
below I want to use 6 instances of the comm with only two devices 0, 1.  I see crashes even with smaller instances..for eg. two instances of comm with each device. Does nccl assume that only one comm is created per device? This is restrictive if this is the case,

./build/test/single/broadcast_test 10000000 6 0  0 0 0 0 1
INFO NCCL debug level set to INFO
INFO rank 0 using buffSize = 2097152
INFO rank 0 using device 0 (0000:03:00.0)
INFO rank 1 using buffSize = 2097152
INFO rank 1 using device 0 (0000:03:00.0)
INFO rank 2 using buffSize = 2097152
INFO rank 2 using device 0 (0000:03:00.0)
INFO rank 3 using buffSize = 2097152
INFO rank 3 using device 0 (0000:03:00.0)
INFO rank 4 using buffSize = 2097152
INFO rank 4 using device 0 (0000:03:00.0)
Segmentation fault

Amith

nluehr:
NCCL should support multiple ranks per GPU. After a few dozen runs on my workstation, I have not been able to reproduce the bug. How frequently do you see it? Does it happen if all ranks use just a single GPU? What GPU/CPU and OS are you running on?

Thanks,
Nathan

amithr1:
Hi Nathan,

I still see the seg faults..Looks like the crashes are seen in ncclCommInitAll routine as I exit immediately after. I am also pasting the nvidia-smi cmd result below


nvidia-smi 
Mon Mar 21 11:40:42 2016       
+------------------------------------------------------+                       
| NVIDIA-SMI 352.59     Driver Version: 352.59         |                       
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla K80           On   | 0000:03:00.0     Off |                    0 |
| N/A   27C    P8    27W / 149W |     55MiB / 11519MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  Tesla K80           On   | 0000:04:00.0     Off |                    0 |
[...]

PR #2306, opened 2026-07-23
nccl_ep: generation-tag LOW_LATENCY P2P signals to prevent stale-signal consumption

Oseltamivir:
Fixes issue #2303: the `LOW_LATENCY` receive-timeout failure reported in #2303 (LL dispatch/combine receives time out on GB200/GB300 NVL72 while HT works).

## Root cause

The LL count/flag protocol double-buffers signal slots and polls for a **nonzero value**, but signal values carry no generation: a signal left over from the previous use of the same parity buffer (two calls earlier) is bit-identical to a fresh one whenever the workload repeats — per-pair token counts are workload-determined, and finish flags are the constant `1`. If any rank falls one full parity cycle behind (any scheduling hiccup suffices), its peers consume the stale signals, complete their calls without it, and lap it arbitrarily. The pipeline desynchronizes and wedges with the familiar systematic `Warning: NCCL EP timeout for dispatch/combine receive` → `trap()` → error 719.

We root-caused this with a device-side event ring (globaltimer-stamped records of DISP/COMB start, clean-done, count/flag sent/got, dumped at first timeout). At the wedge the four ranks were executing *different call sequences* — two ranks in dispatch #12, one in #11, one still in **combine #9** — while receivers had recorded `CNT_GOT` "from" the lagging rank for dispatch calls it **never executed**: stale counts from two calls earlier, value-indistinguishable at a fixed workload.

This explains the platform asymmetry: on GB200/GB300 the failure is reproducible within hundreds of iterations at any token count (fast iterations make the 2-call lap easy), while on x86 single-node platforms the same defect surfaces only rarely as top-load combine corruption. Any µs-scale perturbation of the kernels (e.g. printf instrumentation) masks it — which made it look like a memory-ordering issue. It is not one: the CUDA-level primitives check out (we validated `st.release.sys` / `ld.acquire.sys` posted-store visibility across cuMem FABRIC and POSIX-FD peer mappings on GB200 in isolation), and neither system-scope atomics for the signal stores nor `__threadfence_system()` after the buffer cleans changes the outcome.

## Fix

Tag every P2P signal value with a per-op generation and poll for the exact generation, so stale signals are inert:

```
count: (gen << 16) | (numTokensSent + 1)     poll accepts iff value >> 16 == gen
flag:  (gen << 16) | 1                        poll accepts iff value >> 16 == gen
```

The generation is stamped per launch (stream-ordered `cudaMemcpyToSymbolAsync` from pinned staging; a SEND-phase launch advances it, a RECV-only launch reuses it). The GIN path is untouched — its accumulating signals have different semantics. This also removes the protocol's sensitivity to clean-vs-arrival ordering on the recycled parity buffer.

Notes for review:

- Generation state is process-global in this change; moving it into per-group workspace (or plumbing it as a kernel argument) is a straightforward follow-up we're happy to do if you prefer that shape.
- `numTokensSent + 1` fits 16 bits up to 65534 dispatch tokens per rank; an `EP_HOST_ASSERT` guard could be added if desired.

## Validation

GB200 NVL72 (aarch64, CUDA 13.0 driver, nvcc 13.3, world within one NVLink domain):

- Before: LL wedges within hundreds of iterations (both world=4 single tray and world=8 across two trays), 100% reproducible.
- After: full decode ladder T=1..256, 2048 samples per point, dispatch+combine correctness checks passing; world=4 and world=8 validated.
- HT behavior unchanged.

One pre-existing rarity remains out of scope: an occasional combine-correctness flake at exactly `max_dispatch_tokens_per_rank` (top rung), which we also observe on x86 with the unmodified wheel.

### DIFF
--- a/contrib/nccl_ep/device/low_latency.cu
+++ b/contrib/nccl_ep/device/low_latency.cu
@@ -9,6 +9,7 @@
  * See LICENSE.txt for more license information.
  */
 
+#include <atomic>
 #include <cooperative_groups.h>
 #include "nccl_device.h"
 #include "device_primitives.cuh"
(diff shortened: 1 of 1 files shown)


kwen2501:
Thanks for the PR. NCCL-EP is moved to the nccl-extensions repo. Would you like to migrate the PR over? We can also help making a copy.

Oseltamivir:
Migrated to NVIDIA/nccl-extensions#6 — thanks @kwen2501.

The port is not a straight copy: it is rebased onto `ll_ep.cuh` / `ll_ep_adapter.cu`, and the generation now travels as a kernel argument (threaded like `signalsBase`) rather than a `__device__` symbol, since the LL kernels over there are JIT-compiled and a symbol in the AOT module is not the one the kernel would read.

The root-cause description has also been corrected. This PR attributed the failure to a rank lapping a parity cycle and said LL worked on x86; both were wrong. The real trigger is two LL handles on one group aliasing the same parity signal slots — `handle->ll.buffer_idx` is per-handle while its layout offsets point into the per-group `rdma_buffer` — and it reproduces on x86 too. Full details in NVIDIA/nccl#2303 and on the new PR.

Closing in favour of the migrated one.

PR #2309, opened 2026-07-27
Add IB active-standby failover

### LINKED ISSUE #2310: [RFE]: Add active-standby paths for fused IB NICs
### Goal

Add an opt-in active-standby data path for fused Net IB devices. Normal payload and CTS traffic should use the topology-nearest primary NIC, while a second connected NIC remains ready for failover.

### Problem statement and test evidence

The current fused-device data path treats participating QPs as equivalent and splits each request approximately evenly across them. That assumption does not hold when one GPU reaches one NIC through PIX and another through PHB.

H100 measurements show the following per-NIC bandwidth and congestion behavior:

| Configuration | Per-NIC bandwidth | PFC/CNP |
| --- | ---: | --- |
| One PIX NIC | 400 Gbps | None |
| One PHB NIC | 234 Gbps | None |
| Two PHB NICs | 117 Gbps | A small amount of PFC |
| One PIX and one PHB NIC | 70 Gbps | Heavy PFC/CNP on the PIX NIC |

Two PHB NICs at approximately 117 Gbps each provide roughly the same aggregate bandwidth as one PHB NIC at 234 Gbps, indicating that the PHB paths likely share upstream bandwidth. More importantly, combining PIX and PHB reduces each NIC to approximately 70 Gbps and produces heavy PFC/CNP on the otherwise faster PIX NIC.

The result indicates that equal striping across heterogeneous DMA and PCIe paths creates resource contention and backpressure rather than aggregating their nominal bandwidth. The slower path delays request completion, while concurrent access to the same GPU can contend for PCIe, host-bridge, GPU DMA, and network buffering resources.

### Who benefits

Users running NCCL over fused IB devices on heterogeneous GPU-to-NIC PCIe topologies, especially systems where one NIC is connected through PIX and another through PHB.

### Target infrastructure

This request targets NCCL Net IB P2P transport with NIC Fusion, IB resiliency port failover, and port recovery. It does not introduce a public NCCL API.

### Requested behavior

- Configure unordered redundant HCA pairs.
- Expose two topology-directed virtual NIC directions so NCCL selects the nearest primary for each GPU.
- Keep primary and standby data QPs connected, but send normal payload and CTS traffic only on the primary.
- Probe standby data QPs in both directions without affecting user request completion accounting.
- Map failed primary QPs to the same lane on the standby and reuse completion probing for selective retransmission.
- Automatically return to the topology-selected primary after successful port recovery.
- Preserve existing Active-Active behavior when the feature is not configured.

The initial scope handles one failed device at a time. Standby health probing is driven by primary payload or CTS progress.

### Workflow impact

This avoids concurrent traffic over heterogeneous PIX and PHB paths in the healthy state while preserving transparent failover. It also prevents standby resources from inflating the topology bandwidth reported to NCCL graph search.

### Priority

Medium. The feature targets deployments where fused NIC resiliency is required but Active-Active traffic across asymmetric PCIe paths causes a measured performance regression and congestion backpressure.

### Validation

The implementation has been validated with clean builds on two NCCL nodes, non-destructive active-standby functional cases, dual-GPU PIX/PHB topology direction selection, pair-order equivalence, and real IB port down/up failover and recovery.

Implementation: #2309

tingmingzhong:
## Description

Add an opt-in active-standby data path for fused Net IB devices. An unordered HCA pair now produces two topology-directed virtual NICs so NCCL can select the PCIe-nearest primary for each GPU while keeping the other device connected as a standby.

Normal payload and CTS traffic use only the selected primary. If that device fails, logical QPs are mapped to the same lane on the standby, existing completion probing drives selective retransmission, and successful port recovery automatically restores the primary path.

## Related Issues

Closes #2310

## Changes & Impact

- Parse and validate `NCCL_IB_RESILIENCY_HCA_PAIRS` with matching `NCCL_NET_FORCE_MERGE` groups.
- Separate topology properties from runtime properties and create both primary directions for each HCA pair.
- Exchange and validate active-standby policy, logical device roles, and QP lane counts during connection setup.
- Keep standby data QPs ready while selecting only primary QPs for payload and CTS traffic.
- Probe standby data QPs in both directions with rate-limited, signaled RDMA writes.
- Isolate health-probe completions from user request completion accounting.
- Fail over to the same QP lane, recover through the existing port recovery protocol, and automatically fail back to the topology-selected primary.
- Re-arm recovery QPs after successful recovery so repeated cycles do not accumulate stale work requests.

There are no public API changes. Existing Active-Active behavior is unchanged unless HCA pairs are explicitly configured with port failover enabled. The current implementation handles one failed device at a time, and standby health probes are driven by primary payload or CTS progress.

## Performance Impact

Normal active-standby traffic uses only the topology-selected primary NIC, avoiding concurrent PIX and PHB data paths for the same GPU. The standby carries only rate-limited health probes until failover. Configurations without `NCCL_IB_RESILIENCY_HCA_PAIRS` retain the existing behavior.

Validation performed:

- Clean builds on two NCCL nodes.
- Core non-destructive active-standby functional cases.
- Dual-GPU PIX/PHB topology direction selection and pair-order equivalence.
- Real IB port down/up failover, recovery, and automatic failback.

### DIFF
--- a/src/transport/net_ib/common.cc
+++ b/src/transport/net_ib/common.cc
@@ -57,6 +57,9 @@ ncclResult_t ncclIbBaseCommInit(struct ncclIbNetCommBase* baseComm, bool isSend)
     memset(&baseComm->qps[i].rtsAttr, 0, sizeof(baseComm->qps[i].rtsAttr));
   }
   baseComm->nqps = -1;
+  baseComm->dataPathPolicy = ncclIbDataPathActiveActive;
+  baseComm->primaryDevIndex = -1;
+  baseComm->standbyDevIndex = -1;
   baseComm->splitDataOnQps = ncclParamIbSplitDataOnQps();
   baseComm->nDataQps = -1;
   baseComm->isSend = isSend;
--- a/src/transport/net_ib/common.h
+++ b/src/transport/net_ib/common.h
@@ -57,8 +57,31 @@ struct ncclIbMrCache {
 extern int ncclNMergedIbDevs;
 #define NCCL_IB_MAX_DEVS_PER_NIC NCCL_NET_MAX_DEVS_PER_NIC
 #define MAX_MERGED_DEV_NAME (MAXNAMESIZE * NCCL_IB_MAX_DEVS_PER_NIC) + NCCL_IB_MAX_DEVS_PER_NIC
+
+enum ncclIbDataPathPolicy {
+  ncclIbDataPathActiveActive = 0,
+  ncclIbDataPathActiveStandby = 1,
+};
+
+#define NCCL_IB_DATA_PATH_INFO_VERSION 1
+struct ncclIbDataPathInfo {
+  uint8_t version;
+  uint8_t policy;
+  uint8_t primaryDevIndex;
+  uint8_t standbyDevIndex;
+  uint16_t qpsPerDevice;
+};
+
 struct alignas(64) ncclIbMergedDev {
+  // vProps is the runtime view used for QP/CQ/MR creation. In active-standby
+  // mode it is normalized as [primary, standby]. topoVProps is reported to
+  // NCCL core and contains only the primary device.
   ncclNetVDeviceProps_t vProps;
+  ncclNetVDeviceProps_t topoVProps;
+  enum ncclIbDataPathPolicy dataPathPolicy;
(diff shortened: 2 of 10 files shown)


xiaofanl-nvidia:
++ @thomasgillis @raminudelman

PR #2311, opened 2026-07-27
EFA GDA: drop shared-mem result broadcast in flushImplMode

anshumang:
[EFA GDA] Removes the within CTA return value sharing in `flush` when timeout is set.


kgioioso:
Merged. Thanks for your contribution.

PR #2314, opened 2026-07-30
EFA GDA: re-read submitted_count while draining in flush

bhasunit:
flushImplMode snapshotted submitted_count and spun until the NIC FI_WRITE counter matched it exactly, so concurrent posting on the same context could push completions past the stale target and hang the flush. Re-read it with a scoped relaxed atomic load on each iteration instead.


kgioioso:
This was merged. Closing.

PR #2319, opened 2026-08-04
Initial PACE open source commit

Gaojiaqi:
PACE is a parallelism-aware collective engine for LLM training and inference. It fuses the memory operations that frameworks otherwise insert around communication — layout conversion, dtype casting, scatter/gather — directly into the collective kernels, and exposes a parallelism interface rather than the traditional collective interface: a compute kernel's output feeds straight into PACE, and PACE's output is consumed directly by the next kernel.

### DIFF
--- a/contrib/pace/README.md
+++ b/contrib/pace/README.md
@@ -6,7 +6,7 @@ communication — layout conversion, dtype casting, scatter/gather — directly
 into the collective kernels, and exposes a *parallelism* interface rather than
 the traditional collective interface: a compute kernel's output feeds straight
 into PACE, and PACE's output is consumed directly by the next kernel. PACE is
-built on top of NCCL's **GIN (Graph Interface for Networking)** device API.
+built on top of NCCL's **GIN (GPU Interface for Networking)** device API.
 
 This `contrib/` entry contains the **All-Gather (AG)**, **Reduce-Scatter
 (RS)** and **Scatter-Gather (SG / Ulysses-CP All-to-All)** collectives of
@@ -70,10 +70,10 @@ tensor between them:
 * **Prologue** prepares the compute kernel's output for communication:
   multi-tensor (list) input, on-the-wire dtype conversion, strided input read
   in place, and scaling factors.
-* **Communication** moves the data. We provide SM-resident consume kernels
-  similar to NCCL's, plus resource-lean **0-SM (intranode)** and **1-SM
-  (internode)** variants to minimize the SM footprint. Both variants use
-  pipelined staging, keeping the shared buffer bounded — independent of the
+* **Communication** moves the data. We provide SM-resident consume kernels,
+  plus resource-lean **0-SM (intranode)** and **1-SM (internode)** variants to
+  minimize the SM footprint (Scatter-Gather and All-Gather only). Both variants
+  use pipelined staging, keeping the shared buffer bounded — independent of the
   tensor size — without sacrificing bandwidth. For All-Gather collective, CUDA
   Graph is supported to reduce CPU launching overhead.
 * **Epilogue** is the reverse of the prologue: it lays the received data out so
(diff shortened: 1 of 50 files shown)


contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
+# PACE — Parallelism-Aware Collective Engine
+
+PACE is a parallelism-aware collective engine for LLM training and inference.
+It fuses the memory operations that frameworks otherwise insert around
+communication — layout conversion, dtype casting, scatter/gather — directly
+into the collective kernels, and exposes a *parallelism* interface rather than
+the traditional collective interface: a compute kernel's output feeds straight
+into PACE, and PACE's output is consumed directly by the next kernel. PACE is
+built on top of NCCL's **GIN (Graph Interface for Networking)** device API.
```
nv-lschneider: Please change the abbreviation to: GIN (GPU initiated Networking). Thank you

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+Every PACE collective is structured as three fused stages — **prologue**,
+**communication**, and **epilogue** — so the framework never has to touch the
+tensor between them:
+
+* **Prologue** prepares the compute kernel's output for communication:
+  multi-tensor (list) input, on-the-wire dtype conversion, strided input read
+  in place, and scaling factors.
+* **Communication** moves the data. We provide SM-resident consume kernels
+  similar to NCCL's, plus resource-lean **0-SM (intranode)** and **1-SM
```
nv-lschneider: This not incorrect, but it leaves the impression that NCCL does not support Zero-SM collectives.
That is not correct, please reconsider the statement to clarify that NCCL supports both zero-SM and SM utilizing collectives as well.
Maybe you do not need an explicit comparison to NCCL here in the first place.

nv-lschneider: Also, for zero-SM only certain collective can be supported.
reducing collectives, absolutely need SMs.

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+|---|---|---|---|
+| Multi-tensor (list) input | ✅ | ✅ | ✅ |
+| Dtype conversion | ✅ `fp32→bf16` | ✅ `fp32`/`bf16` | — |
+| Strided input | — | — | ✅ QKV-split |
+| Scaling factors | — | ✅ `pre`/`postdivide` | — |
+
+## Hardware prerequisites
+
+* **GPU:** SM 9.0 or newer. NVLink between GPUs and
+  (for multi-node) a GIN-capable fabric (e.g. GPUDirect Async / GDAKI).
```
nv-lschneider: I think this statement can be read incorrectly. You need NVLIN and/or GIN networking capabilities, correct?

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+* Multi-node runs require GIN networking that NCCL exposes through the
```
nv-lschneider: This statement can also be read incorrectly. Please clarify, that if multi nodes are within an nvlink domain (MNNVL), the GIN support is not required.
I assume you can do all transports via NVLINK in that case.
And it might be more helpful to point to documentation, instead of a header file.
For example this one:
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/device_gin.html

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+  `nccl_device.h` GIN device API.
+
+## Software prerequisites
+
+* PyTorch ≥ 2.1 (Python ≥ 3.12)
+* CUDA Toolkit ≥ 12.0
```
nv-lschneider: Please ensure and double check that this CUDA version is supported with the NCCL version you quote here.
Thank you.

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+* **NCCL ≥ 2.30** with GIN support — the host-side GIN API
+  (`ncclDevCommCreate`, `ncclMemAlloc`, `ncclCommWindowRegister`,
+  `ncclGetLsaDevicePointer`, …) and the `nccl_device.h` device API are
+  required. The version bundled with most PyTorch distributions is too old;
```
nv-lschneider: Please clarify this statement.
I don't think NCCL is generally too old in pytorch.
Instead explain how to check the NCCL version and if it is actually too old, how to update it for pytorch.

Thank you.

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+  git clone https://github.com/NVIDIA/nccl
+  cd nccl
+  make -j install BUILDDIR=/path/to/nccl-install
+  ```
+  then point `NCCL_DIR` at `/path/to/nccl-install` (see below).
+
+## Build
+
+PACE is a PyTorch C++ extension. The build links the GIN-capable NCCL
+**statically** (`libnccl_static.a`) into `pace_cpp` so that calls resolve to
```
nv-lschneider, line 100: We need to be a little bit careful here.
I understand what you are accomplishing, but maybe recommending users to use 2 different NCCL versions at the same time is not ideal.
It is not supported to have 2 different runtime versions of NCCL at the same time.

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+the NCCL version it was built against, independent of the (often older)
+`libnccl.so.2` that PyTorch loads into the process.
+
+```bash
+NCCL_DIR=/path/to/nccl-install \
+TORCH_CUDA_ARCH_LIST="9.0" \
```
nv-lschneider: Maybe it is worth pointing out, that higher arch SM are supported?

contrib/pace/README.md
```diff
@@ -0,0 +1,181 @@
...
+given collective are ignored by it — `ag_zero_sm` is AG-only, `use_wg` is
+RS-only, and SG always uses the data-ring.
+
+| Field | Default | Applies to | Effect |
+|---|---|---|---|
+| `slot_unroll` | `16` | AG / RS / SG | Copy slot length (unrolls of 1024 float4). |
+| `nvl_ring_size` | `4` | AG / RS / SG | NVLink ring depth of the staging ring buffer. |
+| `rdma_ring_size` | `4` | AG / RS / SG | RDMA ring depth of the staging ring buffer. |
+| `num_sms` | `32` | AG / RS / SG | SM-resident kernel grid size. `0` selects the 0-SM (stream) path, `1` the 1-SM lean variant. |
+| `ultra_node_scope` | `0` | AG / RS / SG | NVL fabric scope override (debug). `0` accepts NCCL's natural LSA-team scope (hypernode auto-detect, e.g. NVL72 across boxes is one NVL group). `>0` narrows the scope to force gdaki/inter-team traffic within one NCCL LSA team (e.g. `4` on a GB200 4×2 box simulates 2 nodes). |
```
nv-lschneider: Is this necessary? Wouldn't it be better to use 
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-mnnvl-clique-id
for this.
I.e. each clique set like this, for example `NCCL_MNNVL_CLIQUE_ID=$SLURM_NODEID`, assigns each LSA team to a single slurm node. and inter-node traffic is automatically seen as networking/GIN.

nv-lschneider: Also please call inter-team traffic GIN instead of: `gdaki/inter-team`.
GDAKI is just a GIN backend and transparent to the user.

nv-lschneider (changes requested):
Thank you so much for you effort to integrate PACE into NCCL as a contrib package.

I did not review the full source code just now, but I took a look at the README.
Overall it looks good, but could you refine it with my suggestions?

Thank you!

contrib/pace/README.md
nv-lschneider: Thank you for addressing my comments.

I think it still says interface for GIN.
The correct abbreviation is: GIN (GPU initiated networking).
Thank you.

xiaofanl-nvidia:
Merged to staging branch here: https://github.com/NVIDIA/nccl/commit/a764de587071a67a305fd303447a769b4b224227

Will be part of 2.31 release soon. Closing.

PR #2320, opened 2026-08-04
EFA GDA : make store to mmio explicit and spin with relaxed loads

anshumang:
Use MMIO stores for EFA GDA WQE and doorbell writes, and relaxed loads for maximum inflight, SQ capacity and doorbell-order spin-loop polling to reduce post-path ordering overhead.


xiaofanl-nvidia:
merged here https://github.com/NVIDIA/nccl/commit/3236684d68c77d53b279cf00458fb207ad2f2d36

PR #2321, opened 2026-08-04
EFA GDA: a Put always posts at least one WQE

akkart-aws:
## Description

putImplMode skipped the post entirely for a put with no payload and no signal/counter (hasPayload || needsSignalEp gate): with nothing to deliver and nothing to tick, the operation looked like a no-op and was silently dropped.

That was an as-if optimization from before request aggregation existed, and aggregation broke its premise. With
ncclGinOptFlagsAggregateRequests, a put carries a LOCAL side effect: its non-aggregated doorbell rendezvous in postRdmaWrite is what publishes earlier deferred WQEs on the QP. Dropping "empty" puts therefore made a deferred stream impossible to terminate when its last real put happened to be aggregated -- the tail WQEs (and their signals) were never handed to the NIC and the peer waited forever.

Change the contract so a Put always posts at least one WQE. The empty case degenerates to a 0-byte write to the peer's per-context scratch via the peer DATA endpoint (target slot 0), which binds no FI_REMOTE_WRITE -- remotely unobservable; locally the doorbell rendezvous runs and submitted_count/FI_WRITE tick in lockstep, so Flush accounting stays exact.

### DIFF
--- a/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
+++ b/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
@@ -338,7 +338,6 @@ NCCL_DEVICE_INLINE static void putImplMode(ncclGinCtx ctx, Coop coop, int peer, 
     nccl_ofi_gin_gdaki_dev_handle* dev = getDevHandle(ctx);
 
     bool hasPayload = hasWins && bytes > 0;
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
merged here https://github.com/NVIDIA/nccl/commit/c55dfa93d8624ab817977ddff70549c1d6efc3fe

PR #2333, opened 2026-08-07
fix(efa-gda): Chunk device puts larger than 1 GiB

akkart-aws:
## Description

A GIN device put passed its full size to a single RDMA write WQE. EFA caps a write at efadv max_rdma_size (1 GiB); an oversized WQE fails on the NIC as a CQ error that this CQ-less path never observes, so the put hung silently. The u32 SGE length field additionally truncates sizes >= 4 GiB into short or zero-byte writes.

## Related Issues

Split oversized puts into full-cap leading chunks plus a tail. The leading chunks target the peer's data endpoint, so no remote signal fires and the caller's counter does not tick. Because EFA SRD is unordered, a put that announces completion (signal or counter) must not post its tail until the leading chunks are delivered: spin until the data endpoint's local FI_WRITE counter catches up with submitted_count (an SRD local completion implies remote placement). The tail then carries the signal/counter through the existing path, so the announcement implies full-payload visibility and safe source reuse. Plain puts announce nothing and skip the wait; their ordering is provided by the caller's later flush, signaled put, or barrier.

### DIFF
--- a/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
+++ b/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
@@ -54,6 +54,13 @@ static constexpr cuda::thread_scope ncclGinScope =
  * has wrapped. Never compare absolute counter values. */
 static constexpr uint32_t EFA_CNTR_MASK = 0x7fffffffu;
 
+/* Hardware cap on a single RDMA write: efadv_device_attr.max_rdma_size,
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
https://github.com/NVIDIA/nccl/commit/5cf0cac6ae48a3ede66feb455a4c5be5481bd1c9 merged.

Issue #2335, opened 2026-08-10
VFIO P2P on GeForce maps to a private mirror; NCCL hangs or corrupts


noonghunna:
## Summary

On a 2-GPU VFIO passthrough VM, QEMU's `x-nv-gpudirect-clique` makes the NVIDIA driver grant bidirectional peer access, but the resulting mapping is a **private mirror**, not a path to the peer:

- GPU A writes a pattern through GPU B's peer window.
- GPU A reads the complete pattern back through that window.
- GPU B reads its allocation locally and still sees zeros.
- The result is symmetric in both directions.

This causes every normal NCCL P2P protocol to hang; the one tested copy-engine workaround completes with silently wrong results. The complete observable address chain is correct through QEMU/VFIO, the host IOMMU, Linux DMA APIs, and NVIDIA's open Kernel RM. The mirror forms **after the open KRM boundary**, inside closed GSP-RM/GPU peer-aperture or page-table state.

## Environment

- 2× GeForce RTX 3090 (GA102), PCIe passthrough, no NVLink
- Guest: Ubuntu 24.04, kernel 6.8, NVIDIA 610.57.04 open kernel module
- Host: Proxmox VE 9.1, QEMU 10.1, q35
- `x-nv-gpudirect-clique=1` on both GPUs
- `NVreg_RegistryDwords="RMForceStaticBar1=1;PeerMappingOverride=1"`
- `cudaDeviceCanAccessPeer=1` and `nvidia-smi topo -p2p {r,w}=OK` both ways
- NCCL failure reproduced on 2.25.1, 2.28.9, and 2.29.7

## Minimal discriminator

Observed with a 1 KiB target allocation:

```text
dir 0->1  writer read-back: c0ffee00 5ec897b9  (PAT=c0ffee00 5ec897b9)
dir 0->1  owner local read:  00000000 00000000
dir 1->0  writer read-back: c0ffee00 5ec897b9  (PAT=c0ffee00 5ec897b9)
dir 1->0  owner local read:  00000000 00000000
```

On a healthy mapping, both reads return the pattern. Round-trip bandwidth tests falsely pass because write-to-mirror followed by read-from-mirror is an identity operation.

<details>
<summary>Complete CUDA reproducer</summary>

Build with `nvcc -arch=sm_86 -O2 -o mirror-proof mirror-proof.cu`.

```cuda
#include <cstdio>
#include <cuda_runtime.h>
#define N 256
[...]

PR #2338, opened 2026-08-10
Add NCCL_P2P_VALIDATE to detect false peer-access mirrors

aarav1109s:
## Summary
- Adds optional `NCCL_P2P_VALIDATE=1` in `p2pCanConnect`: **enable peer access**, perform a **device-side peer store** (SM kernel with `__threadfence_system`, matching the #2335 discriminator), then verify with an **owner-local** read. Failed pairs disable P2P and fall back.
- Results are cached per CUDA device pair under a **mutex** (safe for concurrent communicator init).
- When validation is requested but a peer is **invisible** (`CUDA_VISIBLE_DEVICES` / `busIdToCudaDev == -1`) so the SM probe cannot run, **fail closed** and disable P2P for that pair instead of allowing the unvalidated IPC/VMM P2P path. Out-of-range device ids likewise fail closed.
- Documents the false peer-access / private-mirror failure class and warns that `NCCL_P2P_USE_CUDA_MEMCPY` can complete with wrong data on this class.
- Addresses https://github.com/NVIDIA/nccl/issues/2335

This does not fix driver/GSP aperture programming; it fails closed when the direct peer visibility contract is broken, and also when validation was requested but cannot be performed. The probe intentionally avoids `cudaMemcpyPeer` without peer access, which CUDA may stage through host memory and pass even when NCCL’s SM peer aperture is a private mirror.

Default remains off (`NCCL_P2P_VALIDATE=0`) per CONTRIBUTING guidance on workarounds for external components.

## Test plan
- [ ] Build NCCL with this branch (Makefile + CMake device `p2p_validate.cu`)
- [ ] Healthy multi-GPU host: `NCCL_P2P_VALIDATE=1` leaves P2P enabled; collectives match baseline
- [ ] Against the #2335 mirror reproducer (or equivalent VFIO/clique false grant): probe fails, P2P disabled for the pair, collectives succeed via fallback; without the flag, hang / CE wrong-data behavior unchanged
- [ ] With `CUDA_VISIBLE_DEVICES` hiding a peer used via IPC: `NCCL_P2P_VALIDATE=1` disables P2P for that pair (fail closed)
- [ ] Concurrent multi-comm init with `NCCL_P2P_VALIDATE=1` does not race (TSan / stress optional)
- [ ] Docs build for the new env / troubleshooting sections


aarav1109s:
Addressed the review feedback on `39f4bc10`:

1. **Probe path** — replaced `cudaMemcpyPeer` with `cudaDeviceEnablePeerAccess` + device-side peer store (`p2p_validate.cu`, same pattern family as the #2335 reproducer) and owner-local verify, so we exercise the direct aperture NCCL SM protocols use rather than a possible host-staged copy.
2. **Thread safety** — process-global cache is guarded with `std::mutex`; concurrent probes prefer a failure if any observer sees one.
3. **DCO** — commit rewritten with `Signed-off-by: aarav1109s <193567815+aarav1109s@users.noreply.github.com>`.

I don’t have a VFIO/clique mirror rig in this environment; the probe is structured to match the published discriminator, but confirmation on that hardware would still be valuable.

aarav1109s:
Addressed the remaining invisible-device blocker on `7cfb5757`:

**Fail closed when the probe cannot run.** On CUDA 10.1+, NCCL still allows P2P with devices hidden by `CUDA_VISIBLE_DEVICES` via IPC/VMM. The SM-store/owner-read probe needs visible `cudaDev` indices, so with `NCCL_P2P_VALIDATE=1` that path now sets `*ret = 0` instead of returning success unvalidated. Out-of-range device ids in the validate cache likewise fail closed. Docs updated accordingly.

**Builds / functional tests.** This contributor environment has no CUDA toolkit / GPUs (`nvcc` / `nvidia-smi` unavailable), so Makefile/CMake builds and the healthy multi-GPU / #2335 verification items in the test plan remain unchecked and need a GPU host (or maintainer CI) before merge. Happy to iterate further if you prefer moving validation onto the IPC/VMM import path instead of fail-closed at `p2pCanConnect`.

xiaofanl-nvidia:
As per the discussion on https://github.com/NVIDIA/nccl/issues/2335, closing this PR since we believe it's better to address the driver issue directly. 

BTW from NCCL side, we have released some new tools such as NCCL diagnostics which could be helpful to find this kind of lower level issues in the future. We would like to introduce these systems in a way that is isolated to the RAS subsystem so that we don't impact performance for users with healthy systems, and keep the NCCL code modular and maintainable.

PR #2341, opened 2026-08-11
gin: Make signal from C bindings respect ginStrongLegacySignals

dboyan:
## Description

As of now, the GIN signals issued from C-style bindings are legacy signals without strength semantics. However, it does not follow the new ginStrongLegacySignals flags like C++ API. So the users of C bindings can only issue weak signal, completely breaking applications that assume the legacy API semantics and creating unnecessary hurdles for programmers.

This commit introduces ginStrongLegacySignals to C bindings so that it is aligned with the legacy semantics of CUDA C++ APIs. In theory, the C bindings should allow users to issue strong or weak signals but it is beyond scope of this commit.

## Related Issues

N/A

## Changes & Impact

The GIN signals from C bindings now follows ginStrongLegacySignals flags in the associated devcomm like their CUDA C++ counterparts.

## Performance Impact

N/A

### DIFF
--- a/src/include/nccl_device/impl/gin__funcs.h
+++ b/src/include/nccl_device/impl/gin__funcs.h
@@ -517,6 +517,7 @@ NCCL_DEVICE_INLINE void ncclGinPut_v2(ncclGin_C* net, ncclTeam team, int peer, n
   if (isSignal) {
     signal.type = NCCL_GIN_SIGNAL_TYPE_INDEXED;
     signal.indexedSignal.signalId = signalId;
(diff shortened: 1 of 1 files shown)


kgioioso:
Thanks @dboyan . This is a good find. Can you also update `ncclGinSignal_v2`

dboyan:
> Can you also update `ncclGinSignal_v2`

All C APIs that involve signals (put, putValue, signal) are updated in the current version.

kgioioso:
Missed that - thanks.

Issue #1781, opened 2025-07-14
Problems with using NVLink across Kubernetes Pods on a same node


Syspretor:
When deploying a LLM inference system on a Kubernetes cluster using the PD disaggregation framework (e.g., Dynamo). The efficiency of this transport between Prefill Pod and Decode Pod impacts the throughput of the inference system. Therefore, **I want that NVlink can be utilized in data transport between different pods located on the same node.** 

However, I found that **even when scheduling both Pods onto the same node, NVLink couldn't be utilized for data transport between Pods, and only RDMA or TCP was available**. Consequently, I conducted some experiments with NCCL:

- Environment
   - Nccl version: 2.27.5
   - Driver Version: 535.161.07
   - CUDA Version: 12.8
   - Node info: 8 * H20 with nvlink full connectivity
  
I conducted NCCL tests using MPIJob, where Rank0 and Rank1 are run in two separate Pods on a a same node, each allocated with one GPU resource.
```
mpirun \
  --allow-run-as-root -x CUDA_VISIBLE_DEVICES=0,1 -x NCCL_DEBUG=TRACE -x PATH -np 1 --host 10.131.x.x /opt/nccl-tests/build/all_reduce_perf -b 8 -e 2G -f 2 -g 1 -t 1 -a 2 -n 50 : \
  --allow-run-as-root -x CUDA_VISIBLE_DEVICES=2,3 -x NCCL_DEBUG=TRACE -x PATH -np 1 --host 10.131.x.x /opt/nccl-tests/build/all_reduce_perf -b 8 -e 2G -f 2 -g 1 -t 1 -a 2 -n 50
```

Get the following output：
```
nccl-tests-gdr-2-worker-0:298:306 [0] NCCL INFO ncclCommInitRank comm 0x5626a6a1a1a0 rank 0 nranks 2 cudaDev 0 nvmlDev 0 busId 8000 commId 0xe4b8e65544d3f326 - Init START
nccl-tests-gdr-2-worker-1:225:232 [0] NCCL INFO ncclCommInitRank comm 0x5588c7079580 rank 1 nranks 2 cudaDev 0 nvmlDev 2 busId a2000 commId 0xe4b8e65544d3f326 - Init START
nccl-tests-gdr-2-worker-1:225:232 [0] NCCL INFO RAS client listening socket at ::1<28028>
nccl-tests-gdr-2-worker-0:298:306 [0] NCCL INFO RAS client listening socket at ::1<28028>
[...]

PR #2342, opened 2026-08-15
Improve P2P diagnostics for shared-memory isolation

namtran1812:
## Summary

Improve NCCL P2P diagnostics for peers on the same host that cannot use P2P because they see different shared-memory devices.

This update keeps the change local to `src/diagnostics/p2p.cc`:

- keeps same-`hostHash` peers in the diagnostics candidate set, including peers separated by `/dev/shm` isolation
- preserves ownership tracking for dynamically constructed rank sets
- wraps the existing `ncclTopoCheckP2p` call in a diagnostics-local helper
- identifies `same hostHash + different shmDev` locally as `shmIsolation`
- logs these topology-skipped edges through `INFO(NCCL_INIT, ...)`
- does not treat shared-memory isolation as a diagnostic test failure
- does not modify the shared topology API or transport-selection behavior

Related to #1781.

## Motivation

Peers on the same physical host can see different `shmDev` values when running in isolated `/dev/shm` namespaces.

`ncclTopoCheckP2p` already rejects P2P for this case. However, these peers can be separated from the topology-local rank group before P2P diagnostics inspect them, so the reason for the skipped edge is not visible.

The diagnostics candidate set therefore includes peers sharing the local rank's `hostHash`. A helper local to P2P diagnostics calls the existing `ncclTopoCheckP2p` and, when P2P is rejected for a same-`hostHash` pair with different `shmDev` values, records the skipped edge as:

    reason=shmIsolation

This only improves diagnostic visibility; it does not make the pair eligible for P2P.

## Expected diagnostic output

For same-host peers with different shared-memory devices:

    Diagnostics P2P skipped srcRank=0 dstRank=1 reason=shmIsolation srcShmDev=87 dstShmDev=88

The edge is skipped rather than counted as a failed P2P diagnostic.

## Validation

The PR now targets the latest `dev` branch and contains one change:

    src/diagnostics/p2p.cc

`git diff --check` passes.

A local macOS `src.build` cannot complete because NCCL's generated device-header build requires the supported CUDA/Linux build environment; the failure occurs while generating `build/include/nccl_device/*`, before compiling this P2P diagnostics change.

DCO passes.

## Notes

The candidate-set expansion remains restricted to peers with the same `hostHash`. MNNVL and cross-clique behavior are preserved.

No changes are made to `ncclTopoCheckP2p` or the shared topology API.

### DIFF
--- a/src/diagnostics/p2p.cc
+++ b/src/diagnostics/p2p.cc
@@ -94,18 +94,47 @@ struct ncclDiagP2pMapping {
   int peerAccessDev;
 };
 
-static ncclResult_t ncclDiagP2pBuildRankSet(struct ncclComm* comm, int** ranks, int* rank, int* nRanks) {
+static ncclResult_t ncclDiagP2pBuildRankSet(struct ncclComm* comm, int** ranks, int* rank, int* nRanks,
+                                             bool* ranksOwned) {
+  *ranksOwned = false;
+
   // The topology is fused over the local ranks normally and over the MNNVL clique when MNNVL is active. Cross-clique
   // P2P expands that set to every rank in the same NVLink fabric domain.
   if (!comm->p2pCrossClique) {
-    *ranks = comm->MNNVL ? comm->clique.ranks : comm->localRankToRank;
-    *rank = comm->MNNVL ? comm->cliqueRank : comm->localRank;
-    *nRanks = comm->MNNVL ? comm->clique.size : comm->localRanks;
-    return ncclSuccess;
+    if (comm->MNNVL) {
+      *ranks = comm->clique.ranks;
+      *rank = comm->cliqueRank;
+      *nRanks = comm->clique.size;
+      return ncclSuccess;
+    }
+
+    // Keep same-host peers visible to diagnostics even when isolated shared-memory
+    // namespaces caused topology discovery to place them in different local nodes.
+    uint64_t hostHash = comm->peerInfo[comm->rank].hostHash;
+    int count = 0;
+    for (int peer = 0; peer < comm->nRanks; peer++) {
+      if (comm->peerInfo[peer].hostHash == hostHash) count++;
+    }
+
+    NCCLCHECK(ncclCalloc(ranks, count));
+    *ranksOwned = true;
+    *rank = -1;
+    *nRanks = count;
+
+    int slot = 0;
+    for (int peer = 0; peer < comm->nRanks; peer++) {
+      if (comm->peerInfo[peer].hostHash != hostHash) continue;
+      (*ranks)[slot] = peer;
+      if (peer == comm->rank) *rank = slot;
+      slot++;
+    }
+
+    return *rank >= 0 && slot == *nRanks ? ncclSuccess : ncclInternalError;
   }
 
   if (comm->nvlDomainSize <= 0) return ncclInternalError;
   NCCLCHECK(ncclCalloc(ranks, comm->nvlDomainSize));
+  *ranksOwned = true;
   *rank = -1;
   *nRanks = comm->nvlDomainSize;
   int slot = 0;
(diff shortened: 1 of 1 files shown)


stephenmsachs:
Hi @namtran1812. Thanks a lot for opening this PR and adding this functionality. Really appreciated!  We are looking at the changes and will provide feedback on possible updates to get this merged.

teojgo:
@namtran1812 Thank you for adding this, it addresses a diagnostics blind spot for peers separated by `/dev/shm` isolation.

I think it would be better to keep these changes local to P2P diagnostics, since the rejection reason is only needed there. This would avoid changing `ncclTopoCheckP2p`.

Peers with different `hostHash` values are already rejected by the existing topology logic, so a separate `hostIdentity` diagnostic reason may not be necessary. The missing case is same `hostHash` peers with different `shmDev` values, because they can be excluded from the topology-local diagnostic group before diagnostics inspects them.

A possible approach would be:
- Keep the same-`hostHash` diagnostics candidate list and its ownership tracking.
- Add a helper inside `src/diagnostics/p2p.cc` that calls the existing `ncclTopoCheckP2p`.
- If P2P is rejected for peers with the same `hostHash` but different `shmDev`, classify it locally as `shmIsolation`.
- Record the skipped edge through `INFO(NCCL_INIT, ...)`, without treating it as a diagnostic failure.

This keeps the shared topology API and behavior unchanged while still making the `/dev/shm` isolation visible within P2P diagnostics.

Could you also target the latest `dev` branch so it is aligned with the current diagnostics implementation? Let us know if you need help with the implementation.

namtran1812:
Thanks @teojgo — updated based on your feedback.

The PR now targets `dev`, and the implementation is entirely local to `src/diagnostics/p2p.cc`. I kept the same-`hostHash` candidate-set expansion and ownership tracking, removed the topology API/rejection-reason changes, and added a local wrapper around the existing `ncclTopoCheckP2p`.

When that check rejects a same-`hostHash` pair with different `shmDev` values, diagnostics now logs the skipped edge with `reason=shmIsolation` through `INFO(NCCL_INIT, ...)` without treating it as a diagnostic failure.

The PR is now a single commit touching only `src/diagnostics/p2p.cc`.

teojgo:
@namtran1812 thank you for the update. Could you please apply the two formatting changes? Furthermore, I have one suggestion for keeping the P2P diagnostic logs consistent.

`clang-format` reports the following two continuation-indentation changes:

```diff
 static ncclResult_t ncclDiagP2pBuildRankSet(struct ncclComm* comm, int** ranks, int* rank, int* nRanks,
-                                             bool* ranksOwned) {
+                                            bool* ranksOwned) {
```

```diff
 static ncclResult_t ncclDiagP2pCheckTopo(struct ncclComm* comm, int srcRank, int dstRank, int* p2p, int* read,
-                                             int* intermediateRank, bool* shmIsolation) {
+                                         int* intermediateRank, bool* shmIsolation) {
```

Also, since this is another skipped P2P diagnostic edge, would it make sense to use the existing `ncclDiagP2pLogEdge` helper?

For example:

```cpp
if (!edge->p2p && shmIsolation) {
  char extra[128];
  snprintf(extra, sizeof(extra), "reason=shmIsolation srcShmDev=%ld dstShmDev=%ld",
           (long)comm->peerInfo[comm->rank].shmDev, (long)comm->peerInfo[dst].shmDev);
  ncclDiagP2pLogEdge(comm, "skip", comm->rank, dst, edge, extra);
  continue;
}
```

This preserves the shared-memory device information while using the standard P2P edge-log format. Since `ncclDiagP2pLogEdge` uses `INFO(NCCL_INIT)`, the message remains debug-only and is not emitted as an unconditional diagnostic message.

teojgo:
@namtran1812 After a second review pass, I think it may be cleaner to keep this additional check local to the P2P diagnostics without expanding the normal diagnostic rank set.

`comm->peerInfo` already contains the hostHash and `shmDev` for every rank. Each rank could scan this information directly and identify peers with the same hostHash but different `shmDev` values. Peers with different hostHash values are already handled by the existing topology and MNNVL logic.

The isolated pair can still be logged through `ncclDiagP2pLogEdge()` to follow the existing P2P debug-record convention. A zero-initialized diagnostic record with pathType set to `PATH_DIS` would report `path=DIS`, `handle=NONE`, and `topoRead=0`, correctly indicating that no handle or read mode was selected before the pair was rejected. The shared-memory device information can be included as:

`reason=shmIsolation srcShmDev=<srcDev> dstShmDev=<dstDev>`

The resulting message would remain an `INFO(NCCL_INIT)` debug record:

`NCCL INFO Diagnostics P2P skip ...`

rather than being reported as an active NCCL DIAG failure.

This approach would leave `localRankToRank`, MNNVL grouping, allocation ownership, synchronization, and `ncclTopoCheckP2p()` unchanged.

PR #2358, opened 2026-08-18
Fix OOB read of caller's ncclConfig_t in shim

### LINKED ISSUE #2347: [Issue]: NCCL Checkpoint - allocInitParams() reads past the end of a caller-supplied ncclConfig_t (with version skew)

Application crash

### Version

- NCCL Checkpoint Shim version: NCCL v2.30.7-1, [contrib/nccl_checkpoint/shim.cc](https://github.com/NVIDIA/nccl/blob/7b83616df3ae082a1f32bb74c27458bfe8153a13/contrib/nccl_checkpoint/shim.cc)
- vllm built with NCCL version 2.27.5

### Description

allocInitParams() copies a caller-supplied ncclConfig_t with a whole-struct assignment and then stamps the current size/magic/version onto the result.

ncclConfig_t is size-prefixed so that a caller compiled against an older nccl.h may pass a shorter struct; parseCommConfig() honours that by reading size. The shim does not. Two consequences:

1. **Out-of-bounds read.** *configPtr copies sizeof(ncclConfig_t) bytes from an object the caller allocated smaller. This is a read past the end of a live object, reachable from any application built against a pre-2.30 header.
2. **The garbage is then validated.** Because size is rewritten to the current release's value, parseCommConfig() validates fields that the caller never initialized.

### Reproduction
Any PyTorch- or vLLM-based application; both ship against nvidia-nccl-cu12 2.27.5, which predates the six config fields added in 2.30 (nChannelsPerNetPeer, nvlinkCentricSched, graphUsageMode, numRmaCtx, maxP2pPeers, graphStreamOrdering). Take a checkpoint and restore:

```
init.cc:2371 (parseCommConfig) NCCL WARN Invalid config nChannelsPerNetPeer attribute value 0
NCCL Checkpoint: shim_checkpoint.cc:140 (restoreCommViaInit) -> 4

ncclCheckpointRestore() returns ncclInvalidArgument on every rank. Observed on 8 × L4, TP=8, vLLM, NCCL v2.30.7-1.
```

### Proposed Fix
Initialize from NCCL_CONFIG_INITIALIZER and memcpy only min(configPtr->size, sizeof(ncclConfig_t)) bytes over it, so fields the caller does not own retain their NCCL_CONFIG_UNDEF_INT sentinel and parseCommConfig() skips them. Read netName/commName from the local copy rather than configPtr to avoid the same overread on that path.

### Your platform details

- 8 x L4 GPU, testing multi GPU, single host checkpoint / restore.

nicolexin:
## Description

Problem: allocInitParams() in contrib/nccl_checkpoint/shim.cc copies the caller's config with a whole-struct assignment:

    params->config = *configPtr;

ncclConfig_t is versioned by its leading size field precisely so a caller built against an older nccl.h can pass a shorter struct. The assignment reads sizeof(ncclConfig_t) bytes regardless, so for any such caller it reads past the end of the caller's object. 

## Related Issues

Fixes #2347

## Changes & Impact

Solution: start from NCCL_CONFIG_INITIALIZER and memcpy only min(configPtr->size, sizeof(ncclConfig_t)) bytes over it, so fields the caller never had keep their NCCL_CONFIG_UNDEF_INT sentinel and validation skips them. Read netName and commName out of the local copy rather than the caller's struct, which removes the same overread on that path.

## Performance Impact

Not tested, should be negligible.

### DIFF
--- a/contrib/nccl_checkpoint/shim.cc
+++ b/contrib/nccl_checkpoint/shim.cc
@@ -2,6 +2,7 @@
 #include "shim_core.h"
 #include <dlfcn.h>
 #include <cstdio>
+#include <cstring>
 #include <vector>
 
 using namespace nccl_checkpoint;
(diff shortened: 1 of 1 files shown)


lrbison:
Took this change as part of https://github.com/NVIDIA/nccl/commit/8c764e4667b0236d7625ddce45ce4bf29fc1ae6f

Issue #2239, opened 2026-06-17
[Question]: ProxyOp and ProxyStep in NCCL Inspector


ahmd-k:
### Question

Are there any plans to add NCCL Inspector tracking for these events?

xiaofanl-nvidia:
++ @rishdas @armratner

ahmd-k:
Any updates?

rishdas:
Apologies for late reply yes will most likely try to do it 2.33/2.34 release time frame as an option.
We are also open to accepting pull request e.g. https://github.com/NVIDIA/nccl/pull/2304.
Thank you

LYAccc:
Hi @rishdas @ahmd-k , I’d like to take this on if no one is already working on it.
 
   I’m thinking of starting with opt-in ProxyOp/ProxyStep support in the Inspector’s verbose JSON output. Would that be a reasonable first scope?

Thanks!

rishdas:
yes that would and I will be happy to to review.
@xiaofanl-nvidia , @gcongiu, @shamisp

PR #2359, opened 2026-08-19
Add Inspector ProxyOp and ProxyStep tracking

LYAccc:
## Summary

Adds opt-in ProxyOp/ProxyStep tracking to the NCCL Inspector profiler plugin. Related to #2239.

Rebased the original eight implementation commits onto `dev` (`62d10969`), preserving that breakdown and adding review fixes and a documentation correction.

- Enable JSON-only tracking with `NCCL_INSPECTOR_ENABLE_PROXY=1`; Prometheus/OTLP do not arm Proxy callbacks or allocate Proxy pools.
- Use fixed-capacity active pools and a bounded per-communicator completed ring. Defaults: 1,024 active Ops, 4,096 active Steps, and **1,024 completed Proxy records**.
- Maintain the Op self-reference plus active-child references, snapshot parent correlation at Op start, and emit pointer-free completed records. Preserve first state callbacks and observed-byte aggregation.
- Add `NCCL_INSPECTOR_DUMP_PROXY_STEPS=0` for **Op-only output**. Step callbacks, pools, lifetime references, and Op byte/count aggregation remain active; only completed Step record construction, ring insertion, and output are suppressed.
- Emit **JSON v4.3**: header/metadata appear once in the communicator's `dump_stats` marker, followed by its counted Coll/P2P/Proxy payload records under one output lock. This affects JSON even with Proxy disabled. Both bundled analysis readers are adapted and retain older-format support. Current `dev` already uses v4.2.
- Reuse the ring's `dropped`/`droppedReported` accounting. Report `proxy_records`, `proxy_dropped_total`, and `proxy_dropped_since_last_dump` once per dump, including stats-only windows.
- Count failed Op allocations separately (`proxy_ops_dropped_*`) and reserve sequence numbers before allocation. Emit `n_steps_started` alongside completed/dropped Step counts. Pool warnings are one-shot and include allocation/capacity counts.
- Cache PID at initialization, remove redundant per-record `origin_pid`, and count foreign-PID PXN skips separately without dereferencing the foreign parent **or context**. `proxy_pxn_skipped_process_total` is process-scoped, repeated in communicator markers; aggregate by hostname/PID using max/latest, not a sum across markers.
- Address helper naming, function signature formatting, read-only parent locks, state-dispatch helpers, unreachable stopped checks, and documented/enforced backend → Proxy enablement → pool-sizing order. Preserve `NCCL_INSPECTOR_PROM_DUMP_STATS` parsing from `dev`.

`trans_size_bytes` means bytes observed by tracked Steps. Dropped Steps or missing transfer-size callbacks can undercount; `n_steps * chunk_size_bytes` cannot reconstruct missing bytes. Proxy timestamps are CPU callback timestamps, not GPU execution timing.

## Testing

The following validation was completed on a separate GPU server for the review-fix revision. GPU runs used a single node with four GPUs and `NCCL_INSPECTOR_REQUIRE_KERNEL_TIMING=0`; the default value of `1` has not yet been specifically validated. No macOS compatibility shims, CUDA stubs, or host-only test infrastructure are included in this PR.

### Completed validation

| Test | Result |
| --- | --- |
| Builds | Standalone Inspector, full NCCL, and nccl-tests all built successfully; confirmed that the newly built libraries were loaded. |
| Default-transport AllReduce | 192 collective records; correctness checks passed. |
| Socket AllReduce, full output | 768 Ops and 9,216 Steps; parent/child correlation, Step counts, and byte aggregation matched; zero drops. |
| Socket AllReduce, Op-only output | 768 Ops and no Step records; still aggregated 9,216 Steps and 1,409,286,144 bytes, matching full output. |
| Socket SendRecv, full output | 384 Ops and 7,168 Steps, correlated to 384 P2P records; zero drops. |
| Socket SendRecv, Op-only output | 384 Ops and no Step records; aggregated Step counts and bytes matched full output. |
| Op pool exhaustion, capacity 10 | 180 Ops emitted plus 588 reported lost = 768 total; exactly one exhaustion log message. |
| Step pool exhaustion, capacity 10 | 1,772 Steps completed plus 7,444 reported lost = 9,216 total; exactly one exhaustion log message. |
| Ring overwrite, capacity 16 | 4,024 records retained/emitted plus 5,960 reported overwritten = 9,984 total; cumulative and interval counters agreed. |
| Stats-only output | 11 real dump windows with no event records and only loss statistics; losses were reported correctly. |
| Prometheus | Ran with stats disabled and enabled for approximately 61 seconds each; captured metrics for all four GPUs in both modes. |
| OTLP | Aggregated and verbose modes produced 48 and 1,404 valid HTTP exports, respectively. |
| JSON and analysis tools | v4.3 format, context, and record-count validation passed; both analysis tools read the output successfully; legacy-format compatibility and malformed-input checks passed. |

### Performance comparison

Forced Socket transport, four GPUs, message sizes from 1 MiB to 16 MiB. Each mode was run three times; values below are medians. All correctness checks passed, with zero dropped records.

| Mode | Bandwidth (GB/s) | Total elapsed time (s) | JSON size (MB) |
| --- | ---: | ---: | ---: |
| Proxy disabled | 4.523 | 2.616 | 0.624 |
| Full Op + Step output | 4.467 | 2.666 | 28.245 |
| Op-only output | 4.569 | 2.717 | 2.985 |

Op-only output reduced JSON volume by approximately **89.4%** relative to full output. Bandwidth differences were small and subject to run-to-run variation; these measurements do **not** establish a consistent performance improvement or regression. This compares output modes on the review-fix revision, not before/after revisions.

### Documentation correction

Clarified that `NCCL_INSPECTOR_DUMP_THREAD_INTERVAL_MICROSECONDS=-1` (the default) disables the internal dump thread entirely and produces no output. There is no teardown/finalization dump path; the earlier README promise was incorrect, and the observed absence of JSON and `dump_stats` records is expected behavior, not a regression from this PR. The usage introduction now explicitly requires a non-negative interval to obtain output. The environment-variable help text and communication-pattern example were corrected as well. No runtime logic was changed.

### Not yet validated / limitations

- Default kernel-timing filtering (`NCCL_INSPECTOR_REQUIRE_KERNEL_TIMING=1`): not specifically validated; these runs used `0`.


LYAccc:
Hi @rishdas, I’ve opened this PR for the issue(https://github.com/NVIDIA/nccl/issues/2239). When you get a chance, could you take a look? Thanks!

LYAccc:
Hi @rishdas just a friendly ping—when you have a chance, could you please take a look at this PR addressing #2239?

rishdas:
Hi @LYAccc 
Apologies I missed the message before. I will try to go over this next week.

LYAccc:
Thanks @rishdas, no worries! Appreciate you taking a look :)

LYAccc:
Hi @rishdas, just a friendly follow-up on this PR. I’ve split the two original large commits into eight smaller to make life easier for you!

rishdas:
Thank you apologies got busy with few things will circle early next week on this.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -72,6 +72,15 @@ static uint64_t inspectorRecordP2pEventTrace(struct inspectorEventTraceInfo* evt
   return evtTrace[eventIndex].sn;
 }
 
+static uint64_t inspectorRecordProxyEventTrace(
```
rishdas: Formatting suggestion — please match the two sibling functions directly above (`inspectorRecordEventTrace`, `inspectorRecordP2pEventTrace`), which keep the return type and first parameter on the same line and align the rest under it.

```suggestion
static uint64_t inspectorRecordProxyEventTrace(struct inspectorEventTraceInfo* evtTrace,
                                               int eventIndex,
                                               struct inspectorProxyOpInfo* proxyOpInfo) {
```

(First line is 88 columns, well inside the 120 limit.)

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
   inspectorEventPoolReleaseP2p(p2pInfo);
 }
 
+static inspectorResult_t inspectorPluginProxyOpInfoRef(
```
rishdas: Please fix the function signature format to be in sync with other parts of the code

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+    if (inspectorLockWr(&collInfo->guard) != inspectorSuccess) return false;
+    if (collInfo->type != ncclProfileColl || collInfo->commInfo == nullptr) {
+      inspectorUnlockRWLock(&collInfo->guard);
+      return false;
+    }
+    *commInfo = collInfo->commInfo;
+    *parentSn = collInfo->sn;
+    inspectorUnlockRWLock(&collInfo->guard);
+    return true;
+  }
```
rishdas: Push to different function
suggestion on name inspectorPluginProxyOpGetParentInfo
You can club commInfo and parentSn into struct also and include Type in it and have this function fill it up and return.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+    if (inspectorLockWr(&p2pInfo->guard) != inspectorSuccess) return false;
+    if (p2pInfo->type != ncclProfileP2p || p2pInfo->commInfo == nullptr) {
+      inspectorUnlockRWLock(&p2pInfo->guard);
+      return false;
+    }
+    *commInfo = p2pInfo->commInfo;
+    *parentSn = p2pInfo->sn;
+    inspectorUnlockRWLock(&p2pInfo->guard);
+    return true;
+  }
```
rishdas: Same as above push it to function.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+
+  return false;
+}
+
+static void inspectorUpdateCommProxyRecord(
```
rishdas: name should be `inspectorPluginUpdateCommProxyRecord`

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+  uint32_t bufferSize = ring->size + 1;
+  if ((ring->tail + 1) % bufferSize == ring->head) {
+    commInfo->proxyRecordsDropped += 1;
+  }
+  inspectorRingEnqueue(ring, completedProxy);
+  commInfo->dump_proxy = inspectorRingNonEmpty(ring);
+  inspectorUnlockRWLock(&commInfo->guard);
+}
+
+static void inspectorBuildCompletedProxyOpRecord(
```
rishdas: change name to inspectorPluginUpdateProxyOpRecord

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+    const struct inspectorProxyOpInfo* proxyOpInfo) {
+  memset(completedProxy, 0, sizeof(*completedProxy));
+  completedProxy->recordType = NCCL_INSP_PROXY_RECORD_STEP;
+  completedProxy->metadata = proxyOpInfo->metadata;
+  completedProxy->proxyStep.proxyStepSn = proxyStepInfo->proxyStepSn;
+  completedProxy->proxyStep.step = proxyStepInfo->step;
+  completedProxy->proxyStep.transSizeBytes = proxyStepInfo->transSizeBytes;
+  memcpy(completedProxy->proxyStep.evntTrace, proxyStepInfo->evntTrace,
+         sizeof(completedProxy->proxyStep.evntTrace));
+}
```
rishdas: A lot of the timing data is present in Event trace but I would urge to revise this bit by including per step and per op exec time and bw calculation.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+  proxyOpInfoPtr->metadata.peer = eDescr->proxyOp.peer;
+  proxyOpInfoPtr->metadata.channelId = eDescr->proxyOp.channelId;
+  proxyOpInfoPtr->metadata.isSend = eDescr->proxyOp.isSend != 0;
+  proxyOpInfoPtr->nSteps = eDescr->proxyOp.nSteps;
+  proxyOpInfoPtr->chunkSize = eDescr->proxyOp.chunkSize;
+  proxyOpInfoPtr->stopped = false;
+  proxyOpInfoPtr->eventSeqNum = 0;
+  inspectorRecordProxyEventTrace(
+    proxyOpInfoPtr->evntTrace, NCCL_INSP_EVT_TRK_PROXY_OP_START,
+    proxyOpInfoPtr);
```
rishdas, line 400: Push this to different static inline function for readibility

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+
+  uint64_t parentType = *(uint64_t*)eDescr->parentObj;
+  if (parentType != ncclProfileColl && parentType != ncclProfileP2p) return;
+
+  struct inspectorCommInfo* commInfo = nullptr;
+  uint64_t parentSn = 0;
+  if (!inspectorPluginProxyOpParentSnapshot(eDescr->parentObj, parentType,
+                                            &commInfo, &parentSn)) {
+    return;
+  }
```
rishdas: Can be pushed into function like get parent info.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+}
+
+static void inspectorPluginProxyStepInfoInit(
+    struct inspectorProxyStepInfo** proxyStepInfo,
+    ncclProfilerEventDescr_t* eDescr) {
+  *proxyStepInfo = nullptr;
+  if (eDescr->parentObj == nullptr) return;
+  struct inspectorProxyOpInfo* proxyOpInfo
+    = (struct inspectorProxyOpInfo*)eDescr->parentObj;
+  if (proxyOpInfo->type != ncclProfileProxyOp) return;
```
rishdas: Same, push to get parent info

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+  if (inspectorLockWr(&proxyOpInfo->guard) != inspectorSuccess) {
+    inspectorEventPoolReleaseProxyStep(proxyStepInfoPtr);
+    return;
+  }
+
+  if (proxyOpInfo->type != ncclProfileProxyOp || proxyOpInfo->stopped) {
+    inspectorUnlockRWLock(&proxyOpInfo->guard);
+    inspectorEventPoolReleaseProxyStep(proxyStepInfoPtr);
+    return;
+  }
```
rishdas: This can be moved up in func where we get ProxyOpInfo

rishdas:
> Thank you apologies got busy with few things will circle early next week on this.

I have started the review, been juggling with things, will finish all comments and publish

LYAccc:
Thanks for the update, @rishdas. Appreciate your time to review this. I’ll address the comments once they’re posted.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -718,6 +1017,79 @@ static ncclResult_t inspectorPluginRecordEventStateKernelCh(struct inspectorKern
...
+    switch (eState) {
+      case ncclProfilerProxyStepSendGPUWait:
+        return NCCL_INSP_EVT_TRK_PROXY_STEP_SEND_GPU_WAIT;
+      case ncclProfilerProxyStepSendPeerWait_v4:
+        return NCCL_INSP_EVT_TRK_PROXY_STEP_SEND_PEER_WAIT;
+      case ncclProfilerProxyStepSendWait:
+        return NCCL_INSP_EVT_TRK_PROXY_STEP_SEND_WAIT;
+      default:
+        return -1;
+    }
```
rishdas: You can further push it to function

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -718,6 +1017,79 @@ static ncclResult_t inspectorPluginRecordEventStateKernelCh(struct inspectorKern
...
+
+  switch (eState) {
+    case ncclProfilerProxyStepRecvWait:
+      return NCCL_INSP_EVT_TRK_PROXY_STEP_RECV_WAIT;
+    case ncclProfilerProxyStepRecvFlushWait:
+      return NCCL_INSP_EVT_TRK_PROXY_STEP_RECV_FLUSH_WAIT;
+    case ncclProfilerProxyStepRecvGPUWait:
+      return NCCL_INSP_EVT_TRK_PROXY_STEP_RECV_GPU_WAIT;
+    default:
+      return -1;
```
rishdas, line 1081: Same.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -791,7 +1171,17 @@ __hidden ncclResult_t inspectorPluginRecordEventState(void* eHandle,
...
-  if (type == ncclProfileKernelCh && eState == ncclProfilerKernelChStop) {
+  if (type == ncclProfileProxyOp) {
+    struct inspectorProxyOpInfo* proxyOpInfo
+      = (struct inspectorProxyOpInfo*)eHandle;
+    return inspectorPluginRecordEventStateProxyOp(proxyOpInfo, eState);
+  } else if (type == ncclProfileProxyStep) {
+    struct inspectorProxyStepInfo* proxyStepInfo
+      = (struct inspectorProxyStepInfo*)eHandle;
+    return inspectorPluginRecordEventStateProxyStep(
+      proxyStepInfo, eState, eStateArgs);
```
rishdas: Nit formating guidline.
```
inspectorPluginRecordEventStateProxyStep(proxyStepInfo,
                                         eState,
                                         eStateArgs);
```

plugins/profiler/inspector/inspector.cc
```diff
@@ -376,8 +386,7 @@ inspectorResult_t inspectorCommInfoListFinalize(struct inspectorCommInfoList* co
     TRACE_INSPECTOR("NCCL Inspector: comm %lu still in tracker",
                     commList->comms->commHash);
     nextComm = commList->comms->next;
-    inspectorRingFinalize(&commList->comms->completedCollRing);
-    inspectorRingFinalize(&commList->comms->completedP2pRing);
+    inspectorCommInfoRingsFinalize(commList->comms);
```
rishdas, line 398: Nice.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -58,7 +59,7 @@ static inspectorResult_t inspectorCommInfoHeader(jsonFileOutput* jfo,
 static inspectorResult_t inspectorCommInfoMetaHeader(jsonFileOutput* jfo) {
   JSON_CHK(jsonStartObject(jfo));
   {
-    JSON_CHK(jsonKey(jfo, "inspector_output_format_version")); JSON_CHK(jsonStr(jfo, "v4.0"));
+    JSON_CHK(jsonKey(jfo, "inspector_output_format_version")); JSON_CHK(jsonStr(jfo, "v4.1"));
```
rishdas: **Blocking: format-version collision with `dev`.**

`dev` already emits `v4.1` — bumped by #2304, merged as `ae9733378 profiler/inspector: report ring buffer drops`, where `v4.1` means "a `dump_stats` record exists". This PR independently bumps `v4.0` -> `v4.1` to mean "proxy records exist". We'd ship two mutually incompatible `v4.1` formats with no way for a consumer to tell them apart.

Please make this `v4.2` after rebasing onto `dev`.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+  completedProxy->recordSn = ++commInfo->nextProxyRecordSn;
+
+  if (ring->size == 0 || ring->entries == nullptr) {
+    commInfo->proxyRecordsDropped += 1;
+    inspectorUnlockRWLock(&commInfo->guard);
+    return;
+  }
+
+  uint32_t bufferSize = ring->size + 1;
+  if ((ring->tail + 1) % bufferSize == ring->head) {
```
rishdas: **Blocking: this duplicates drop accounting that now exists on `dev`.**

After the #2304 merge, `inspectorRingEnqueue()` already detects the overwrite and maintains `ring->dropped` / `ring->droppedReported`. Here we re-derive the same condition by hand, immediately before calling the enqueue that detects it, and record it in a second counter (`commInfo->proxyRecordsDropped`). Two counters for one event.

Please drop the manual peek and `proxyRecordsDropped`, and read `ring->dropped` like the coll and P2P paths do.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -257,6 +258,163 @@ static inline inspectorResult_t inspectorCompletedP2p(jsonFileOutput* jfo,
...
+  }
+
+  JSON_CHK(jsonStartObject(jfo));
+  {
+    JSON_CHK(jsonKey(jfo, "record_type"));
+    JSON_CHK(jsonStr(jfo, proxy->recordType == NCCL_INSP_PROXY_RECORD_OP
+                           ? "proxy_op" : "proxy_step"));
+    JSON_CHK(jsonKey(jfo, "record_sn"));
+    JSON_CHK(jsonUint64(jfo, proxy->recordSn));
+    JSON_CHK(jsonKey(jfo, "proxy_records_dropped"));
```
rishdas: **Blocking: report drops once per dump, not on every record.**

This stamps a cumulative counter onto every single proxy record. That's the same per-record-vs-per-dump point I raised on #2304, and `dev` now has the `dump_stats` marker record specifically to hold this kind of field.

Please move it there as `proxy_records` / `proxy_dropped_total` / `proxy_dropped_since_last_dump`, alongside the existing `coll_*` and `p2p_*` fields, and drop the per-record key. Two secondary problems disappear with it: as a cumulative value stamped on records a consumer can't tell *when* the drops happened, and see my note on the `recordsDropped` assignment above for the empty-ring case.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -369,13 +527,62 @@ static inspectorResult_t inspectorCommInfoDumpP2p(jsonFileOutput* jfo,
...
+  inspectorLockWr(&commInfo->guard);
+  if (commInfo->dump_proxy) {
+    if (commInfo->completedProxyRing.size > 0
+        && drainedProxy.capacity() < commInfo->completedProxyRing.size) {
+      drainedProxy.reserve(commInfo->completedProxyRing.size);
+    }
+    INS_CHK(inspectorRingDrain<inspectorCompletedProxyRecord>(
+      &commInfo->completedProxyRing, drainedProxy));
+    commInfo->dump_proxy = inspectorRingNonEmpty(&commInfo->completedProxyRing);
+    recordsDropped = commInfo->proxyRecordsDropped;
```
rishdas: `recordsDropped` is only assigned inside `if (commInfo->dump_proxy)`, so a dump window where the ring is empty but drops *did* occur reports `0` instead of the real count. Folding this into the `dump_stats` record (see my other comment) fixes it naturally, since that path snapshots unconditionally under the same guard.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+  struct inspectorCommInfo* commInfo = nullptr;
+  uint64_t parentSn = 0;
+  if (!inspectorPluginProxyOpParentSnapshot(eDescr->parentObj, parentType,
+                                            &commInfo, &parentSn)) {
+    return;
+  }
+
+  struct inspectorProxyOpInfo* proxyOpInfoPtr
+    = inspectorEventPoolAllocProxyOp();
+  if (proxyOpInfoPtr == nullptr) return;
```
rishdas: **Blocking: ProxyOp-pool exhaustion is silent data loss.**

You count step-pool exhaustion (`n_steps_dropped`) and ring overwrite (`proxy_records_dropped`), but when `inspectorEventPoolAllocProxyOp()` returns `nullptr` the Op just disappears — no counter anywhere. It doesn't even leave a sequence-number hole, because `metadata.proxyOpSn` is assigned from `commInfo->nextProxyOpSn` only *after* a successful allocation, so gap analysis can't recover it either.

Measured on 4x GB200 with `NCCL_INSPECTOR_PROXY_OP_POOL_SIZE=16`: **858 failed Op allocations, 200 Op records emitted, and nothing in the JSON indicating any loss.**

Given #2304 just landed specifically to stop the Inspector losing records silently, please add a per-comm counter and surface it in `dump_stats`.

Related caveat worth documenting either way: because dropped steps contribute no bytes, a ProxyOp's `trans_size_bytes` is bytes *observed*, not bytes *transferred*. In the run above, Ops with dropped steps averaged 332 KB against 2.18 MB for Ops without, at similar `n_steps` (7.5 vs 9.0) — a 6.5x under-report. And it can't be reconstructed from `n_steps * chunk_size_bytes`, since `chunkSize` is the max slice size, not the realized one. Please state in the README that `trans_size_bytes` is only trustworthy when `n_steps_dropped == 0`.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+}
+
+static void inspectorPluginProxyOpInfoInit(
+    struct inspectorProxyOpInfo** proxyOpInfo,
+    ncclProfilerEventDescr_t* eDescr) {
+  *proxyOpInfo = nullptr;
+
+  // PXN may carry a parent pointer from a different process. Check the PID
+  // before reading parentObj or any context derived from it.
+  if (eDescr->proxyOp.pid != getpid() || eDescr->parentObj == nullptr) return;
```
rishdas: Two things on this line.

**`getpid()` is a syscall on the ProxyOp hot path.** glibc removed its `getpid()` cache in 2.25, so this is a real syscall per ProxyOp start. Please cache the pid once at `inspectorGlobalInit()` and compare against the cached value.

**The PXN skip is silent.** Cross-process detached-PXN Ops return here with no counter and no log, so on a real PXN multi-node run proxy records vanish with no indication — same class of problem as the Op-pool exhaustion above. Please count these separately (they're a legitimate, expected skip rather than an error, so they deserve their own counter) and surface it.

Minor follow-on: because mismatches are filtered out here, `metadata.originPid` is always `getpid()` by construction, which makes the emitted `origin_pid` JSON field a constant.

plugins/profiler/inspector/inspector_event_pool.cc
```diff
@@ -546,6 +695,88 @@ struct inspectorP2pInfo* inspectorEventPoolAllocP2p() {
...
+  if (g_eventPool.proxyStepChunkList == nullptr) {
+    return nullptr;
+  }
+
+  pthread_mutex_lock(&g_eventPool.proxyStepPoolLock);
+
+  if (g_eventPool.proxyStepFreeList == nullptr) {
+    pthread_mutex_unlock(&g_eventPool.proxyStepPoolLock);
+    WARN_INSPECTOR(
+      "NCCL Inspector: ProxyStep pool exhausted (capacity: %u) - allocation failed!",
```
rishdas: **Blocking: this warning is not rate-limited, and it sits on the proxy progress hot path.**

It fires on every failed allocation. Measured on an 8-second 4x GB200 run with small pools: **2,389 warning lines** (1,531 from here, 858 from the ProxyOp variant at line 721).

#2304 established the one-shot pattern for exactly this — see `ringDropWarned` with `__atomic_exchange_n` in `inspector_plugin.cc`. Please reuse it here and in `inspectorEventPoolAllocProxyOp()`.

While you're here: `proxyStepAllocCount` / `proxyOpAllocCount` are maintained but never read anywhere. They'd be genuinely useful in this message.

plugins/profiler/inspector/inspector_event_pool.cc
```diff
@@ -546,6 +695,88 @@ struct inspectorP2pInfo* inspectorEventPoolAllocP2p() {
...
+  if (g_eventPool.proxyOpChunkList == nullptr) {
+    return nullptr;
+  }
+
+  pthread_mutex_lock(&g_eventPool.proxyOpPoolLock);
+
+  if (g_eventPool.proxyOpFreeList == nullptr) {
+    pthread_mutex_unlock(&g_eventPool.proxyOpPoolLock);
+    WARN_INSPECTOR(
+      "NCCL Inspector: ProxyOp pool exhausted (capacity: %u) - allocation failed!",
```
rishdas: Same unthrottled-warning issue as the ProxyStep variant — please apply the one-shot guard here too. This one accounted for 858 of the 2,389 lines in my single node test run.
you can make it Info level also.

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+    }
+    return;
+  }
+
+  if (inspectorLockWr(&proxyOpInfo->guard) != inspectorSuccess) {
+    inspectorEventPoolReleaseProxyStep(proxyStepInfoPtr);
+    return;
+  }
+
+  if (proxyOpInfo->type != ncclProfileProxyOp || proxyOpInfo->stopped) {
```
rishdas: **This `stopped` check (and the one at line 415) is unreachable — worth either removing or documenting.**

`ncclProfilerStopProxyOpEvent()` sets `sub->opEventHandle = NULL` right after stopping, and `ncclProfilerStart{Send,Recv}ProxyStepEvent()` pass that same field as `eDescr.parentObj`. So a step starting after its Op stopped arrives with `parentObj == nullptr` and is already rejected by your null check above.

This matters because as written the code reads as though a step *can* arrive with a live-but-stopped parent — which would be a use-after-free question, since the Op is pool-released and its rwlock destroyed once refCount hits zero. It isn't reachable, but a future reader can't tell that from here. Same reasoning is what makes the unsynchronized `proxyOpInfo->type` read before the lock safe; that deserves a comment too.

(For the record, the refcount design itself is correct — I went looking for a UAF here and there isn't one.)

plugins/profiler/inspector/inspector_plugin.cc
```diff
@@ -219,6 +231,220 @@ static void inspectorPluginP2pInfoCleanup(struct inspectorP2pInfo *p2pInfo) {
...
+  inspectorEventPoolReleaseProxyOp(proxyOpInfo);
+}
+
+static bool inspectorPluginProxyOpParentSnapshot(
+    void* parentObj, uint64_t parentType,
+    struct inspectorCommInfo** commInfo, uint64_t* parentSn) {
+  if (parentType == ncclProfileColl) {
+    struct inspectorCollInfo* collInfo
+      = (struct inspectorCollInfo*)parentObj;
+    if (inspectorLockWr(&collInfo->guard) != inspectorSuccess) return false;
```
rishdas: This takes the parent's **write** lock to read two fields (`commInfo` and `sn`). `inspectorLockRd()` would do, and this runs once per ProxyOp against a lock the collective completion path also takes for every operation.

(Applies to the P2P branch below as well.)

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -369,13 +527,62 @@ static inspectorResult_t inspectorCommInfoDumpP2p(jsonFileOutput* jfo,
...
+  }
+  inspectorUnlockRWLock(&commInfo->guard);
+
+  if (!drainedProxy.empty()) {
+    *needs_writing = true;
+    JSON_CHK(jsonLockOutput(jfo));
+    for (size_t i = 0; i < drainedProxy.size(); i++) {
+      JSON_CHK(jsonStartObject(jfo));
+      {
+        JSON_CHK(jsonKey(jfo, "header"));
```
rishdas: **Output volume — my main non-correctness concern.**

Every proxy record repeats the full `header` and `metadata` blocks. Measured on 4x GB200, `all_reduce_perf -b 1M -e 16M -f 2 -g 4 -w 2 -n 20` over a forced Socket NET path:

| | JSON size | busbw |
|---|---|---|
| proxy off | 882 KB | 0.621 GB/s |
| proxy on | 35.7 MB | 0.524 GB/s |

That's **40x** output growth and a **15.6%** bandwidth loss, with 98% of the file being proxy records and **31.4% of the proxy bytes being repeated header/metadata**.

Three things would help, in rough order of value: emit `header`/`metadata` once per dump rather than per record; let step records be enabled separately from op records (op-level alone is useful and roughly 10x cheaper here); and reconsider `NCCL_INSPECTOR_DUMP_PROXY_RING_SIZE` defaulting to 4096, which is 4x the collective ring.

This is opt-in so it isn't a blocker, but anyone who enables it at scale will notice immediately.

plugins/profiler/inspector/inspector.h
```diff
@@ -236,6 +302,38 @@ struct inspectorKernelChInfo {
...
+// completed records copy the pointer-free metadata instead.
+struct inspectorProxyOpInfo {
+  uint64_t type;
+  int refCount;
+  struct inspectorCommInfo* commInfo;
+  struct inspectorProxyOpMetadata metadata;
+  int nSteps;
+  int chunkSize;
+  bool stopped;
+  uint32_t nStepsStarted;
```
rishdas, line 318: `nStepsStarted` is incremented in `inspectorPluginProxyStepInfoInit()` and never read anywhere. Either emit it (it would pair naturally with `n_steps_completed` / `n_steps_dropped` and would make Op-level step accounting self-checking) or drop it.

plugins/profiler/inspector/inspector.h
```diff
@@ -157,6 +182,42 @@ struct inspectorCompletedOpInfo {
   struct inspectorEventTrkOpInfo evtTrk;
 };
 
+// Pointer-free metadata shared by active and completed Proxy records.
+struct inspectorProxyOpMetadata {
+  uint64_t parentType;
+  uint64_t parentSn;
+  uint64_t proxyOpSn;
+  pid_t originPid;
```
rishdas: See my note on the PID check: since foreign-PID Ops are filtered out before a record is ever built, this field can only ever hold `getpid()`, so the emitted `origin_pid` is a constant. I would suggest to either drop it, or make it meaningful by actually recording the skipped cross-process Ops.

plugins/profiler/inspector/inspector_json.cc
rishdas: Just a note by time it gets merged it might require another revision.

plugins/profiler/inspector/inspector.cc
```diff
@@ -1107,6 +1179,8 @@ inspectorResult_t inspectorGlobalInit(int rank) {
 
   INS_CHK(inspectorGlobalStateInit());
   initP2pTrackingFromEnv();
+  initOutputBackendFromEnv();
+  initProxyTrackingFromEnv();
```
rishdas, line 1211: **These three calls now have a load-bearing order, and nothing here says so.**

`initProxyTrackingFromEnv()` decides `enableNcclInspectorProxy` by reading `enableNcclInspectorPromDump` and `inspectorOtelIsEnabled()`, both of which are set by `initOutputBackendFromEnv()` on the line above. `inspectorEventPoolInitFromEnv()` two lines below then reads `enableNcclInspectorProxy` to decide whether to size the proxy pools at all.

So the chain is backend -> proxy-enable -> pool-sizing, and every failure mode from getting it wrong is silent. Move the proxy call above the backend call and the Prometheus/OTEL gate stops firing, so proxy callbacks stay armed under a backend that cannot consume them. Move the pool init above the proxy call and you either allocate both proxy pools for a disabled feature or skip them for an enabled one.

I verified the current order is correct (1182 -> 1183 -> 1185), so this is not a live bug. I'm raising it because this PR is what creates the dependency, and the rebase onto `dev` disturbs exactly this region: #2304 added `NCCL_INSPECTOR_PROM_DUMP_STATS` parsing at the old Prometheus site, and reconciling that with your relocation into `initOutputBackendFromEnv()` was one of only three real conflicts I hit. That is precisely the kind of edit that silently reorders these.

A two-line comment stating the required order would be enough. Better still, have `inspectorEventPoolInitFromEnv()` assert that backend init has already run, so a future reordering fails loudly instead of quietly disabling telemetry.

plugins/profiler/inspector/README.md
```diff
@@ -295,9 +305,9 @@ Each output file contains JSON objects with the following structure:
     "nnodes": 1
   },
   "metadata": {
-    "inspector_output_format_version": "v4.0",
+    "inspector_output_format_version": "v4.1",
```
rishdas: The version bump needs to reach the docs too — this example and the one at line 345 both hardcode `"v4.1"`, so whatever my format-version comment on `inspector_json.cc` settles on has to be swept through here as well.

Worth flagging that this gets worse after the rebase. #2304 added its own `dump_stats` example to the README on `dev`, also carrying `v4.1`. Once both are in the same file you'll have three version strings across two independently-versioned features, so the sweep has to cover the pre-existing `dev` example and not just the two you added.

rishdas (changes requested):
Thanks for this — the ProxyOp/ProxyStep correlation design is sound and the lifetime handling is careful. I built it standalone and ran it on 4x GB200 over a forced Socket NET path to exercise the proxy events, and the core mechanism holds up well:

- Correlation is exact across **3,680 ProxyOps / 35,328 ProxySteps**: zero byte-total mismatches and zero step-count mismatches between op and step records.
- All 10 trace timestamps are populated on 100% of records in both directions, with plausible durations.
- Ring drop accounting is exact on all 4 ranks, and step-pool exhaustion is attributed to the correct op.
- Clean build under `-Wall -Wextra`, and the default-off path is unchanged.
- The JSON-only gate for Prometheus/OTEL works and warns rather than silently misbehaving, and the fact that `NCCL_INSPECTOR_POOL_GROW` does not apply to the proxy pools is already documented. Both appreciated.

My comments fall into three groups.

**1. Please retarget to `dev`.** This is against `master`, but the Inspector has moved on `dev` — #2304 (`ae9733378 profiler/inspector: report ring buffer drops`) landed there and overlaps this PR directly. I rebased all 8 commits onto `dev` to review them and hit exactly three real conflicts, all tractable:

- `inspector_json.cc` — #2304 restructured `inspectorCommInfoDump` around the new `dump_stats` record. Watch the early-return path: as written, a comm with only proxy traffic and no coll/p2p records returns before proxy records are emitted.
- `inspector.cc` — #2304 added `NCCL_INSPECTOR_PROM_DUMP_STATS` parsing at the old Prometheus site, so that parse has to travel with your relocation into `initOutputBackendFromEnv()`.
- `README.md` — trivial.

`inspector.h`, `inspector_event_pool.{cc,h}` and `inspector_plugin.cc` all apply cleanly.

**2. Blocking, and mostly a consequence of that overlap.** The format version is the important one: `dev` already ships `v4.1` meaning "a `dump_stats` record exists", and this PR independently bumps to `v4.1` meaning "proxy records exist". Shipping two incompatible `v4.1` formats leaves consumers no way to tell them apart. Related: the manual drop detection here duplicates the `ring->dropped` accounting `dev` now maintains, and the per-record `proxy_records_dropped` field belongs in `dump_stats` alongside the existing `coll_*`/`p2p_*` counters.

The other blocking item is independent of the rebase: **ProxyOp-pool exhaustion is silent data loss.** When `inspectorEventPoolAllocProxyOp()` returns null the op vanishes with no counter, and because `proxyOpSn` is assigned only after a successful allocation it doesn't even leave a sequence-number gap. With `NCCL_INSPECTOR_PROXY_OP_POOL_SIZE=16` I measured 858 failed allocations against 200 emitted records, with nothing in the JSON indicating loss. Same shape as the cross-process PXN skip. Given #2304 landed specifically to stop the Inspector losing records quietly, these should be counted and surfaced.

**3. Overhead, which is worth a design conversation rather than a fix.** On `all_reduce_perf -b 1M -e 16M -f 2 -g 4 -w 2 -n 20`, enabling proxy tracking took output from 882 KB to 35.7 MB (**40x**) and busbw from 0.621 to 0.524 GB/s (**-15.6%**). 98% of the file is proxy records and 31.4% of those bytes are header/metadata repeated on every record. It's opt-in so it isn't a blocker, but anyone enabling it at scale will notice immediately. Emitting `header`/`metadata` once per dump, and letting op records be enabled without step records, would recover most of it cheaply.

That last point is also why I'd push back gently on my own earlier request for per-step exec time and bandwidth — those add fields to records that are already the dominant cost. If they go in, they should be labelled as CPU proxy-thread timing so nobody compares them against the GPU-derived `coll_exec_time_us`.

Nothing I found is a correctness bug in the proxy state machine itself; I went looking for a use-after-free around the refcount and pool release and there isn't one. The blocking items are all about the `dev` overlap and about loss being observable.

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Extracted the parent lookup into inspectorPluginProxyOpGetParentInfo() as suggested. It fills an inspectorProxyParentInfo struct containing commInfo, type, and sn, and returns whether the lookup succeeded.

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done as you suggested

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Following your later feedback on output overhead, kept the existing event timestamps without adding derived duration or bandwidth fields

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done, move all of them into inspectorPluginProxyOpFillInfo().

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done, moved the logic into inspectorPluginProxyOpGetParentInfo().

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done.

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done. I moved the parent validation into inspectorPluginProxyStepGetParentInfo() and removed the unreachable stopped check. Good catch, thanks!

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done, moved into inspectorProxyStepRecvStateEventIndex

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done, split into two functions inspectorProxyStepSendStateEventIndex and inspectorProxyStepRecvStateEventIndex

plugins/profiler/inspector/inspector_json.cc
LYAccc: Rebased onto dev and bumped the format to v4.3

plugins/profiler/inspector/inspector_plugin.cc
LYAccc: Done. Removed the manual overflow check and proxyRecordsDropped, and now use the ring’s existing drop counters.

plugins/profiler/inspector/inspector_json.cc
LYAccc: Done. I moved these fields into dump_stats and removed the per-record drop counter. Drop statistics are now sampled even when the Proxy ring is empty.

plugins/profiler/inspector/inspector.h
LYAccc: Done. I removed origin_pid. The local PID remains in the dump marker, and skipped cross-process Ops are now counted separately in dump_stats.

plugins/profiler/inspector/inspector.cc
LYAccc: Done, added two comments documenting the order, an assertion and a runtime check.

plugins/profiler/inspector/README.md
LYAccc: Done. Updated the README to match v4.3, including the dump_stats example inherited from dev.

LYAccc:
Hi @rishdas, thanks again for the detailed review! I’ve addressed your comments and added the test results and known limitations to the PR description. could you plz take another look when you have a chance :)

rishdas:
@LYAccc On the **Known issue — not yet fixed** item: that one is a documentation bug rather than a code bug, so please don't spend time chasing a runtime fix.

`-1` was only ever intended to mean "do not dump periodically". The clause promising output *at communicator teardown/finalization* was documented in error — that path has never existed. The code is behaving exactly as designed:

- `inspectorStartDumpThread()` returns early when `intervalUsecs < 0` and never constructs `dumper`:
  ```c
  static inspectorResult_t inspectorStartDumpThread(int64_t intervalUsecs) {
    if (intervalUsecs < 0) {
      INFO_INSPECTOR("NCCL Inspector: dump thread disabled "
                     "(interval is -1); not starting internal dump thread.");
      return inspectorSuccess;
    }
  ```
- the only two call sites of `inspectorCommInfoListDump()` are inside the dump thread's own dump routine (`inspector.cc:690-691`), so with no thread there is no writer;
- `inspectorGlobalFinalize()` only stops and deletes the dumper — it performs no final flush.

So "zero JSON records and zero `dump_stats` records" with `-1` is correct behavior, and it is not a regression from this PR. What needs fixing is the README sentence that describes the non-existent teardown dump. Something like:

```suggestion
- `NCCL_INSPECTOR_DUMP_THREAD_INTERVAL_MICROSECONDS=<interval>` (default: `-1`)
  Sets the interval (in microseconds) for the internal dump thread to write output. A value of `-1` (default) disables the internal dump thread entirely, so no output is written; set a non-negative value to obtain output. A value of `0` enables continuous dumping (dumps as fast as possible). Set to a positive value to enable periodic dumps at the specified interval (e.g., `500` for every 500 µs). When Prometheus mode is enabled (`NCCL_INSPECTOR_PROM_DUMP=1`), a minimum of `30000000` (30 seconds) is enforced to align with the node exporter polling interval.
```

Two follow-ups worth folding in while you are there:

1. Please drop the "Known issue — not yet fixed" section from the PR description and note the README correction instead, so the fix isn't blocked waiting on a phantom bug.
2. Since `-1` is the default and produces no output at all, the docs should say so plainly up front. Every worked example in the README sets an explicit interval, which is why this has gone unnoticed; a reader taking the default gets silence. Calling that out in the Quick Start would save the next person the same investigation.

Thanks for digging into it and reporting it clearly either way — the write-up is what made it quick to pin down.

LYAccc:
@rishdas, I’ve updated the README and PR description as suggested.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -261,6 +262,157 @@ static inline inspectorResult_t inspectorCompletedP2p(jsonFileOutput* jfo,
...
+      JSON_CHK(jsonUint64(jfo, trace[NCCL_INSP_EVT_TRK_PROXY_STEP_RECV_FLUSH_WAIT].sn));
+      JSON_CHK(jsonKey(jfo, "recv_gpu_wait_sn"));
+      JSON_CHK(jsonUint64(jfo, trace[NCCL_INSP_EVT_TRK_PROXY_STEP_RECV_GPU_WAIT].sn));
+    }
+    JSON_CHK(jsonKey(jfo, "proxy_step_stop_sn"));
+    JSON_CHK(jsonUint64(jfo, trace[NCCL_INSP_EVT_TRK_PROXY_STEP_STOP].sn));
+  }
+  JSON_CHK(jsonFinishObject(jfo));
+
+  JSON_CHK(jsonKey(jfo, "event_trace_ts"));
```
rishdas, line 330: **Please complement these absolute timestamps with elapsed time per phase.** This is my main ask on the revision.

Right now a consumer has to know the state machine to get anything out of a Step record. It has to know that the phase order is direction-specific -- `send_gpu_wait -> send_peer_wait -> send_wait -> stop` for sends, `recv_wait -> recv_flush_wait -> recv_gpu_wait -> stop` for receives -- and that each timestamp marks *entry* into a phase, so a phase's cost is the delta to the *next* one. None of that is recoverable from the record itself.

Decoding one real receive Step from my run, `proxy_op_sn 399 / step 10`:

| phase | derived µs |
|---|---:|
| start -> `recv_wait` | 1 |
| network receive | 6,433 |
| flush | 4,388 |
| GPU consume | 653 |
| total | 11,475 |

That is genuinely useful, and nobody will compute it by hand.

**There is also a correctness trap.** These keys are emitted unconditionally from a zero-initialised trace array, so a state that never fired is published as `"..._ts": 0` rather than omitted. Subtracting across it yields a nonsense ~1.8e15 µs phase, and the consumer cannot distinguish "took no time" from "never happened". The plugin can, because it has the matching `sn` -- so it is the right place to compute this.

Concretely, per direction:

```
send:  gpu_produce_us = send_peer_wait - send_gpu_wait
       peer_credit_us = send_wait      - send_peer_wait
       wire_us        = stop           - send_wait
recv:  wire_us        = recv_flush_wait - recv_wait
       flush_us       = recv_gpu_wait   - recv_flush_wait
       gpu_consume_us = stop            - recv_gpu_wait
both:  total_us       = stop - start
```

Emitting a start anchor plus these durations instead of five absolute timestamps is also *smaller*: 123 bytes against 184 on a representative record, about 11% off Step-record bytes, which matters given the volume numbers in my other comment.

If you would rather not break existing consumers, add the durations alongside and deprecate the absolute block later. I would also suggest making `event_trace_sn` opt-in while you are here -- it is 20.4% of Step bytes and is debug-only detail, where the durations are what an operator actually wants.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -261,6 +262,157 @@ static inline inspectorResult_t inspectorCompletedP2p(jsonFileOutput* jfo,
...
+    JSON_CHK(jsonKey(jfo, "proxy_op_start_sn"));
+    JSON_CHK(jsonUint64(jfo, trace[NCCL_INSP_EVT_TRK_PROXY_OP_START].sn));
+    JSON_CHK(jsonKey(jfo, "proxy_op_in_progress_sn"));
+    JSON_CHK(jsonUint64(jfo, trace[NCCL_INSP_EVT_TRK_PROXY_OP_IN_PROGRESS].sn));
+    JSON_CHK(jsonKey(jfo, "proxy_op_stop_sn"));
+    JSON_CHK(jsonUint64(jfo, trace[NCCL_INSP_EVT_TRK_PROXY_OP_STOP].sn));
+  }
+  JSON_CHK(jsonFinishObject(jfo));
+
+  JSON_CHK(jsonKey(jfo, "event_trace_ts"));
```
rishdas, line 287: Same request as on the Step trace below, applied to Ops: please emit `total_us` (`proxy_op_stop_ts - proxy_op_start_ts`) alongside the absolute stamps.

The Op is where an elapsed time is least ambiguous, since there is exactly one interval and no direction-specific ordering to know about. It is also the field that makes `NCCL_INSPECTOR_DUMP_PROXY_STEPS=0` genuinely self-sufficient: in Op-only mode the Step records that would let you derive timing are precisely what you have suppressed, so without a duration here the cheap mode loses all timing signal.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -261,6 +262,157 @@ static inline inspectorResult_t inspectorCompletedP2p(jsonFileOutput* jfo,
...
+      JSON_CHK(jsonInt(jfo, proxy->proxyOp.nSteps));
+      JSON_CHK(jsonKey(jfo, "chunk_size_bytes"));
+      JSON_CHK(jsonInt(jfo, proxy->proxyOp.chunkSize));
+      JSON_CHK(jsonKey(jfo, "n_steps_started"));
+      JSON_CHK(jsonUint32(jfo, proxy->proxyOp.nStepsStarted));
+      JSON_CHK(jsonKey(jfo, "n_steps_completed"));
+      JSON_CHK(jsonUint32(jfo, proxy->proxyOp.nStepsCompleted));
+      JSON_CHK(jsonKey(jfo, "n_steps_dropped"));
+      JSON_CHK(jsonUint32(jfo, proxy->proxyOp.nStepsDropped));
+      JSON_CHK(jsonKey(jfo, "trans_size_bytes"));
```
rishdas, line 399: **Please also derive bandwidth from `trans_size_bytes` over the elapsed time.** You already carry both halves; only the division is missing, and it is the number people actually want from a proxy trace.

For an Op, `trans_size_bytes / (proxy_op_stop_ts - proxy_op_start_ts)` gives an `op_bw_gbs`. For a Step, the meaningful denominator is the **wire phase only**, not the whole Step -- `stop - send_wait` for sends and `recv_flush_wait - recv_wait` for receives. Using total Step time instead folds in GPU waits and flow-control stalls and understates the link badly: on the receive Step I decoded, wire time was 6,433 µs of an 11,475 µs Step, so total-time bandwidth would read 1.8x low.

Two caveats worth baking into the field names and the docs rather than leaving to the reader:

1. These are CPU proxy-thread timings. They must not be compared against `coll_busbw_gbs`, which is GPU-derived. A distinct suffix or a `timing_source: "proxy_cpu"` marker would stop that mistake.
2. Steps are pipelined, so several are in flight concurrently. A per-Step figure is an instantaneous estimate for one slice, not achieved throughput, and summing across Steps will exceed line rate. The Op-level figure is the one that aggregates sensibly.

Given that, my suggestion would be to emit `op_bw_gbs` unconditionally and a per-Step `wire_us` from which a consumer can compute its own rate, rather than publishing a per-Step bandwidth that is easy to misread.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -261,6 +262,157 @@ static inline inspectorResult_t inspectorCompletedP2p(jsonFileOutput* jfo,
...
+      JSON_CHK(jsonSize_t(jfo, proxy->proxyOp.transSizeBytes));
+      INS_CHK(inspectorCompletedProxyOpTrace(jfo, proxy));
+    } else {
+      JSON_CHK(jsonKey(jfo, "proxy_step_sn"));
```
rishdas, line 403: **I would like to push for nesting Steps inside their Op after all.** I argued against this earlier on the grounds that a nested record would be unbounded; measuring the real output, that objection does not hold up.

From the 4xGB200 socket run on this revision (`all_reduce`, 1 MiB-8 MiB):

| | |
|---|---:|
| `proxy_step` records | 14,976 (8.27 MB, **89.4% of the file**) |
| `proxy_op` records | 1,664 (0.86 MB) |
| steps per Op | min 6, **median 12**, max 12 |

Twelve children per parent is not unbounded -- and `nSteps` is known at Op start, so the Op can size its own array up front. `nSteps` does scale with message size over chunk size, so a nested form still needs a cap with an overflow counter for very large messages, but that is a bounded, well-understood shape rather than an open-ended one.

What nesting buys, on top of the structural clarity you were after:

- The 16.1% of Step bytes spent re-stating parent fields disappears by construction (see my other comment).
- The per-record JSON envelope -- `{"proxy_trace":{"record_type":...,"record_sn":...}}` on all 14,976 records -- collapses to one per Op.
- Consumers stop having to join on `(rank, proxy_op_sn)` to answer "what happened in this transfer", which is the question they always ask. I wrote that join to validate this PR; it is fiddly precisely because `proxy_op_sn` is per-communicator *per rank*, so the rank must be part of the key or ranks silently collide.

The real costs, which are the ones to weigh rather than the record size:

- An Op only emits once it finalises, so a **hung or never-completing Op emits nothing** -- losing the trace in the case you most want it. Flat Steps stream out as they complete. If you nest, please keep a way to flush a partial Op, or this becomes a debuggability regression.
- Ring entries are deliberately fixed-size and pointer-free so they can be copied in and drained without ownership questions. A nested array either inflates every entry to the worst case or reintroduces pointers into ring records.

Given that tension, a hybrid may be the better answer: nest by default, since the median Op is 12 Steps and that is where the 89% saving is, and keep flat Step emission available for the streaming and hang-debugging cases. I would rather see that discussed explicitly in the PR than have the flat form be the default by inertia.

plugins/profiler/inspector/inspector_json.cc
```diff
@@ -261,6 +262,157 @@ static inline inspectorResult_t inspectorCompletedP2p(jsonFileOutput* jfo,
...
+    JSON_CHK(jsonUint64(jfo, proxy->metadata.parentSn));
+    JSON_CHK(jsonKey(jfo, "proxy_op_sn"));
+    JSON_CHK(jsonUint64(jfo, proxy->metadata.proxyOpSn));
+    JSON_CHK(jsonKey(jfo, "rank"));
+    JSON_CHK(jsonInt(jfo, proxy->metadata.rank));
+    JSON_CHK(jsonKey(jfo, "channel_id"));
+    JSON_CHK(jsonInt(jfo, proxy->metadata.channelId));
+    JSON_CHK(jsonKey(jfo, "peer"));
+    JSON_CHK(jsonInt(jfo, proxy->metadata.peer));
+    JSON_CHK(jsonKey(jfo, "direction"));
```
rishdas, line 385: Six fields here are re-emitted on every Step record despite being fully determined by the parent Op and derivable from `proxy_op_sn`: `parent_type`, `parent_sn`, `rank`, `channel_id`, `peer`, and `direction`.

Measured on the run in my other comment, that is **1,332,672 bytes, 16.1% of all Step-record bytes**, to restate 1,664 Ops' worth of information 14,976 times. It is the same duplication problem as the per-record `header`/`metadata` you just fixed, one level down.

Nesting removes this by construction. If nesting is rejected, dropping these from Step records is still worth doing on its own -- a consumer already has to hold Op state to interpret a Step, since `parent_sn` alone does not tell it the chunk size or the Step count. Should you keep any single field, keep `direction`, since it is what tells a reader which trace keys to expect.

plugins/profiler/inspector/README.md
```diff
@@ -430,7 +444,131 @@ This will include additional event trace information in the JSON output, showing
...
+
+`trans_size_bytes` is bytes **observed**, not necessarily all bytes transferred.
+It sums the first transfer-size callback for each tracked Step (`send_wait` for
+sends, `recv_flush_wait` for receives). Treat it as a complete byte total only
+when `n_steps_dropped == 0` and the transfer-size callbacks were received. A Step
+without a transfer-size callback still counts as completed, but contributes no
+bytes. Do not reconstruct missing bytes from `n_steps * chunk_size_bytes`:
+`chunk_size_bytes` is a maximum slice size, not the realized transfer size.
+
+For receive steps, the direction-specific trace fields are `recv_wait`, `recv_flush_wait`, and `recv_gpu_wait` instead of the send fields shown above. All `event_trace_ts` values are CPU Proxy callback timestamps in microseconds, not GPU execution timing. Any durations or bandwidth derived from them must not be compared directly with GPU-derived collective timing.
```
rishdas, line 547: While the trace semantics are being documented here, two additions would save every future reader the reverse-engineering I just did.

State the **phase model explicitly**: each `*_wait_ts` is the timestamp at which the Step *entered* that phase, the orders are `send_gpu_wait -> send_peer_wait -> send_wait -> stop` and `recv_wait -> recv_flush_wait -> recv_gpu_wait -> stop`, and a phase's cost is therefore the delta to the following entry. Right now the field list says which keys appear for which direction but never says the values are entry points rather than durations, which is the one thing a reader must know to use them.

Then say what each phase **means**, which is not guessable from the names. From `net.cc`: `send_gpu_wait` is waiting for the GPU to *produce* the data into the send buffer; `send_peer_wait` is waiting for receiver buffer credit; `send_wait` is waiting for `ncclNetIsend` to complete; `recv_wait` is waiting for data to arrive; `recv_flush_wait` is waiting for the flush that makes landed data visible to the GPU; `recv_gpu_wait` is waiting for the GPU to *consume* the buffer and free the slot.

Please also document that an unfired state is published as `0` rather than omitted, so consumers do not subtract across it -- though I would rather see that fixed than documented, per my comment on the emission code.

rishdas (commented):
Re-reviewed at `3254c695b`. Thanks for turning this around so thoroughly — every blocking item from the last pass is genuinely fixed, and I verified them in the code rather than taking the replies at face value: `v4.3` clears the version collision, the manual overflow check is gone in favour of the ring's own counters, drop reporting moved into `dump_stats`, `proxy_ops_dropped_*` now makes Op-pool loss visible (with sequence numbers reserved *before* allocation, which closes the gap-analysis hole), and both pool warnings are one-shot. The unreachable `stopped` check, the write-vs-read lock, `n_steps_started`, `origin_pid`, and the init-ordering comment are all addressed too.

I also re-ran the validation matrix against your revision on a 4xGB200 node, grafting this PR's `plugins/profiler/inspector/` onto our internal `dev` (the PR touches only that directory) and forcing the socket transport so Proxy events actually fire. 14 cells, all pass:

| check | result |
|---|---|
| default off | `v4.3`, zero proxy records |
| full Op+Step | 1,664 Ops / 14,976 Steps, correlation exact, no Op undercounting its observed Steps |
| Op-only (`DUMP_PROXY_STEPS=0`) | 1,664 Ops / 0 Steps, 988 KB vs 8.9 MB |
| Op-pool exhaustion | 259 emitted + 328 counted dropped, one warning line |
| Step-pool exhaustion | 12,327 Steps counted dropped, one warning line |
| ring overwrite | 3,984 counted dropped, accounting consistent |
| Prometheus gate | Proxy correctly refused, no JSON |
| marker contract | zero payload records carrying `header`/`metadata` |
| p2p path | 825 Ops / 13,145 Steps on `sendrecv` |

Correctness was clean throughout (`#wrong=0`, no hangs in graph-replay stress). Two pre-existing issues showed up that are **not** yours: `sendrecv`/`broadcast` emit no `coll_perf` records under `NCCL_INSPECTOR_REQUIRE_KERNEL_TIMING=1` (the task-relative `KernelCh.channelId` and allgatherv-scheduler cases). Notably `sendrecv` *proxy* records emit fine, since proxy timing is CPU-side and bypasses that gate.

The comments below are all enhancements rather than blockers. Three themes: give the wait timestamps meaning by emitting elapsed time, derive bandwidth from `trans_size_bytes` over that elapsed time, and reconsider nesting Steps inside their Op.

**One thing that needs a profiler-interface change, so it is a follow-up rather than a request here: Proxy records have no notion of which NIC they used.** `ncclProfilerEventDescr_v5_t.proxyOp` carries `pid`, `channelId`, `peer`, `nSteps`, `chunkSize`, `isSend`, and `proxyStep` carries only `step`, so the plugin cannot know or derive the device. `channel_id` correlates with a NIC on multi-rail nodes but does not identify one, and the channel-to-device mapping lives in core's connection resources, invisible to the profiler.

The awkward part is that core has it right at the call site: `resources->netDev` is set at connection setup (`net.cc:764`) and `net.cc` already casts `sub->connection->transportResources` a few lines from where it records the Step states. So exposing it is one field, not a redesign — adding `netDev` (ideally alongside the device name or PCI path from `ncclNetProperties`) to the proxyOp descriptor in the next profiler interface version. Without it, per-NIC attribution -- arguably the main reason to trace proxy activity at scale -- has to be reconstructed from `NCCL_DEBUG=INFO` logs. Worth filing separately; I am happy to raise it if you would rather not expand this PR's scope.

rishdas:
> m happy to raise it if you would rather not expand this PR's scope.

To clarify I mean I will take up this action item of working on getting device information to plugin

plugins/profiler/inspector/inspector_json.cc
LYAccc: Hi @rishdas thanks for suggesting nested output, makes more sense here.— I’d like to make it the default and keep flat mode as an option. That should reduce repeated parent fields and make the traces easier to read.

  I’d buffer a capped number of completed Step snapshots per Op, flushing when full or when the Op finishes. Each Step would occupy its own small ring slot, rather than making every slot large enough for a whole Step array.

  Each snapshot is 152 bytes, so caps of 4/8/16 would cost about 0.6/1.2/2.4 KiB per Op, excluding allocator overhead. I’m leaning toward 8, allocating lazily and reusing the buffer until the Op is recycled. Flat and Op-only modes wouldn’t allocate it.

  I’d enqueue each batch under one lock, Steps first and Op metadata last. With FIFO eviction, the Steps would be dropped before their parent, so any retained Step would still have its Op metadata. The dump would drain the ring as it does today, then group the records by Op into nested JSON.

  For hang debugging, I’d emit an Op-start record so an unfinished Op isn’t completely invisible. Partial and final segments would be distinguished by is_final. 

  Does that sound reasonable? If so, given your hands-on experience with the profiler, I’d love to hear what you think would be a good cap for the Step buffer :)

PR #2362, opened 2026-08-19
p2p: register whole record for IPC to fix multi-segment reuse

yshalabi:
## Description

An IPC registration for a VA record backed by multiple cuMem handles only imported the segment under the registered buffer, so reusing that import for a different segment of the record produced an out-of-bounds remote address.

## Changes & Impact

None

## Performance Impact

N/A


xiaofanl-nvidia:
merged here: https://github.com/NVIDIA/nccl/commit/da459565539f40fccd11292f08f07f248a6b2d5a

PR #2364, opened 2026-08-21
net/ib: release device list on all ncclIbInitDevices error paths

### LINKED ISSUE #2189: [Issue]: Potential memory leak in ncclIbInitDevices error paths after wrap_ibv_get_device_list()

Application crash

### Share Your Debug Logs

In `src/transport/net_ib/init.cc`, `ncclIbInitDevices()` obtains an IB device list through:

```cpp
wrap_ibv_get_device_list(&devices, &nIbDevs)

The returned list should be released with wrap_ibv_free_device_list(devices) before leaving ncclIbInitDevices().

Valgrind reports a definite leak from this allocation path:
8 bytes in 1 blocks are definitely lost
  at calloc
  by ibv_get_device_list
  by wrap_ibv_get_device_list
  by ncclIbInitDevices
  by ncclIbInit
  by ncclNetInit
  by commAlloc
  by ncclCommInitRankFunc
  by ncclAsyncJobMain

Static analysis also shows that some exit/error paths after wrap_ibv_get_device_list() can leave ncclIbInitDevices() without releasing devices. One visible example is the Data Direct error path:

WARN("NET/IB: Error in mlx5dv_get_data_direct_sysfs_path with device %s", devices[d]->name);
return res;

This issue was observed with Valgrind memcheck and also confirmed by static analysis of `src/transport/net_ib/init.cc`.


Minimal code path:

1. `ncclIbInitDevices()` calls `wrap_ibv_get_device_list(&devices, &nIbDevs)` successfully.
2. At least one IB device is opened successfully with `wrap_ibv_open_device()`.
3. Data Direct probing reaches `wrap_mlx5dv_get_data_direct_sysfs_path()`.
4. `wrap_mlx5dv_get_data_direct_sysfs_path()` returns an error other than `ncclSuccess` or `ncclInvalidArgument`.
5. The function returns directly from this branch, bypassing `wrap_ibv_free_device_list(devices)`.

Relevant path:

```cpp
if (ncclSuccess != wrap_ibv_get_device_list(&devices, &nIbDevs)) {
  ret = ncclInternalError;
  goto fail;
}
...
ncclResult_t res = wrap_mlx5dv_get_data_direct_sysfs_path(...);
...
} else {
  WARN("NET/IB: Error in mlx5dv_get_data_direct_sysfs_path with device %s", devices[d]->name);
  return res;
}


### NCCL Version

2.29.2

### Your platform details

Observed with Valgrind memcheck while running BCCL 2.29.2.




### Error Message & Behavior

<img width="945" height="265" alt="Image" src="https://github.com/user-attachments/assets/1e630ffe-48a0-42c2-8306-374c2821e7ea" />
Valgrind reports:

8 bytes in 1 blocks are definitely lost
  at calloc
  by ibv_get_device_list
  by wrap_ibv_get_device_list
  by ncclIbInitDevices
  by ncclIbInit
  by ncclNetInit
  by commAlloc
  by ncclCommInitRankFunc

This indicates that the device list returned by ibv_get_device_list() is not released on at least one ncclIbInitDevices() exit path.

### PR DESCRIPTION
**Pull request by rootkiller6788**
## Fix

Closes #2189. `ncclIbInitDevices()` calls `wrap_ibv_get_device_list()` (which allocates the device list) but only released it on the success path. Any error after the allocation leaked the list:

- the `wrap_mlx5dv_get_data_direct_sysfs_path()` failure branch did `return res;` directly, bypassing `wrap_ibv_free_device_list()`;
- every `NCCLCHECKGOTO(..., ret, fail)` after allocation (`ncclIbQueryOooRqSize`, `ncclIbGetPciPath`, `ncclIbGidInfoQuery`, and the `wrap_ibv_close_device()` failure) jumped to `fail:`, which had no cleanup;
- two plain `NCCLCHECK` calls after allocation (`ncclCalloc` for the data-direct `pciPath`, `ncclIbStatsInit`) returned directly on error, leaking the list too.

## Change

- Hoist `struct ibv_device** devices` to function scope (initialized to `NULL`) so the `fail:` label can reach it.
- Route the data-direct failure branch through `fail:`.
- Convert the two post-allocation plain `NCCLCHECK` calls to `NCCLCHECKGOTO(..., ret, fail)`.
- Release the device list in `fail:` (guarded by the `NULL` check), and set `devices = NULL` after the success-path free so later failure paths do not double-free.

Behavior on the success path is unchanged. All error paths now release the list exactly once.

### DIFF
--- a/src/transport/net_ib/init.cc
+++ b/src/transport/net_ib/init.cc
@@ -286,6 +286,7 @@ ncclResult_t ncclIbFinalizeDevices(void) {
 extern int64_t ncclIbArThreshold;
 ncclResult_t ncclIbInitDevices(ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction) {
   ncclResult_t ret = ncclSuccess;
+  struct ibv_device** devices = NULL;
   if (netRefCount++) return ret;
   ncclProfilerFunction = profFunction;
   if (ncclParamIbDisable()) return ncclInternalError;
@@ -311,7 +312,6 @@ ncclResult_t ncclIbInitDevices(ncclDebugLogger_t logFunction, ncclProfilerCallba
 
       // Detect IB cards
       int nIbDevs;
-      struct ibv_device** devices;
 
       // Check if user defined which IB device:port to use
       const char* userIbEnv = ncclGetEnv("NCCL_IB_HCA");
@@ -390,7 +390,8 @@ ncclResult_t ncclIbInitDevices(ncclDebugLogger_t logFunction, ncclProfilerCallba
                 TRACE(NCCL_NET, "NET/IB: Device %s does not support Data Direct DMA.", devices[d]->name);
               } else {
                 WARN("NET/IB: Error in mlx5dv_get_data_direct_sysfs_path with device %s", devices[d]->name);
-                return res;
+                ret = res;
+                goto fail;
               }
             }
           }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by xiaofanl-nvidia** [comment]
++ @raminudelman @jynv to review and mirror if this is good.

File: src/transport/net_ib/init.cc [hunk 1]
```diff
@@ -466,9 +467,13 @@ ncclResult_t ncclIbInitDevices(ncclDebugLogger_t logFunction, ncclProfilerCallba
...
-        ret = ncclInternalError;
-        goto fail;
+      if (devices) {
+        ncclResult_t freeRet = wrap_ibv_free_device_list(devices);
+        devices = NULL;
+        if (freeRet != ncclSuccess) {
+          ret = ncclInternalError;
+          goto fail;
+        }
       }
```
raminudelman, line 477: I don't see how this change is required and fixes something compared to the shorter 3 lines we have.

jynv:
Thanks for the fix. The memory is allocated in provider but never freed in the error path. Please revise the change based on the comment above and simplify the error handling path.

PR #2365, opened 2026-08-21
Fix task-slot publication race in socket transport (NULL data send on aarch64)

jsolman:
## What

`ncclNetSocketGetTask()` publishes task slots to `persistentSocketThread` helper threads with plain, unfenced stores: the slot fields (`op`/`data`/`size`/`sock`/`offset`/`result`) are written, then `used = 1` is stored with no ordering guarantee. On weakly-ordered architectures (aarch64) the helper can observe `used == 1` before the preceding stores are visible, read `data` as NULL (task queues are `ncclCalloc`'d), and execute `send(fd, NULL, size)` — a deterministic EFAULT:

```
NCCL WARN ncclOsSocketProgressOpt: Call to send to <ip><port> failed : Bad address
           (op=0 fd=N buf=(nil) size=65536 off=0)
```

Depending on timing this either tears down the connection (`ncclRemoteError` cascade at init) or silently wedges it: in the symmetric reap race, `ncclNetSocketTest()` resets a slot's `used`/`result` fields while the helper thread is in-flight inside `ncclSocketProgress()`, so the error is overwritten by the fresh `ncclSuccess` initialization before the helper's error check observes it — the helper believes the task succeeded, `Test()` never sees completion, and the peer's GPU kernels spin in `waitPeer` forever (a permanent 98%-SM-at-idle-power hang of the whole collective).

## Fix

Make the `used` flag the publication/reap synchronization point with acquire/release atomics (all in `src/transport/net_socket.cc`), using NCCL's existing `COMPILER_ATOMIC_STORE`/`COMPILER_ATOMIC_LOAD` macros:

- `ncclNetSocketGetTask()`: release-store `used = 1` when publishing a slot
- `persistentSocketThread()`: acquire-load (`COMPILER_ATOMIC_LOAD(...) == 1`) in the helper scan
- `ncclNetSocketTest()`: release-store `used = 0` on both reap paths (task-with-subs and main-thread-progress)

x86 is unaffected (TSO forbids the store reordering), which is why the race only reproduces on weakly-ordered systems.

## Evidence

- Live capture on a 4-node Jetson Thor (SM110, aarch64) cluster, NCCL 2.31.2, socket transport: `buf=(nil)` EFAULT with `size=65536`, `off=0`, clustered during communicator init, ~1-in-2 to 1-in-3 launches.
- Instrumented builds with glibc backtrace captured the exact stack: `ncclOsSocketProgressOpt <- socketProgress <- ncclSocketProgress <- persistentSocketThread` — the send executes from the helper's task loop, from a slot whose `data` is NULL.
- An entry probe in `ncclSocketProgress` fired with `buf=NULL` immediately before the EFAULT, while the error-check code directly after the call in the same iteration never observed the error — only explainable by concurrent slot mutation (reap + reuse) between syscall return and check.
- The task queue memory is calloc'd, explaining why the pointer is exactly NULL (initialization value) rather than garbage.

## Validation

- 2-rank allreduce over real inter-node sockets: correct results, no spurious warnings on the clean path (patch is behavior-neutral when the race doesn't fire).
- Post-fix 4-node vLLM launch cycles: no `Bad address` / NULL-ptr events observed across repeated communicator-heavy launches (tripswitch instrumentation retained in the validation builds); long-running multi-node serving continues clean.

CC @sjeaugey @bbpt


jsolman:
## Field validation update

This fix has now been running in production for many hours on the 4-node Jetson Thor (SM110, aarch64) cluster where the bug was originally observed — the exact deployment that was experiencing ~1-in-2 launch failures and permanent serving hangs.

- **Pre-fix**: intermittent `Call to send ... failed : Bad address` (buf=(nil)) launches and silent multi-node wedges (peer ranks spinning in `waitPeer`, 98% SM at idle power)
- **Post-fix** (this change, deployed as a swapped-in NCCL build): **many continuous hours of serving with zero recurrence** — no NULL-send EFAULTs, no wedges, no stalls

The hang problems that preceded this fix have not recurred since deployment.

jsolinsky-fb:
@nv-udeodhar @kiskra-nvidia Will you be able to review this soon? Thanks!

jsolman:
Can we get this reviewed and merged? @kiskra-nvidia @nv-udeodhar  @rgioiosa78

sjeaugey:
Looks good to me.

I'm not sure I would have added such a large comment to explain memory ordering; this is just standard acquire / release semantics which weren't implemented so far.

Same on the commit message, I would have been fine with a concise message like 

```
net/Socket: use proper acquire/release memory semantics for thread synchronization

When dispatching work to progress threads, make sure the "used" field is set/read 
with proper acquire/release semantics to avoid race conditions on instruction-reordering
architectures, e.g. aarch64.
```
I think it says is all, given the code is literally just replacing a couple of loads and stores with their respective atomic acquire / release versions.

kiskra-nvidia (approved):

kiskra-nvidia:
@sjeaugey I agree. I'll edit the comment and the commit message as part of the review process.

kiskra-nvidia:
This was merged to the `dev` branch (41027c8). As such, it should be included in the second 2.32 release (it missed the window for the first 2.32 release, which is frozen and currently being tested).

jsolman:
I adjusted the commit message.

xiaofanl-nvidia:
This was already merged to dev branch so I'm closing this PR. Thanks for your contribution! 

Please open a new PR is anything else needs to be changed.

PR #2366, opened 2026-08-23
Skip NET shared buffers for single-rank communicators

### LINKED ISSUE #2363: [Issue]: Single-rank communicator initializes a large NET shared P2P buffer
### Impact

A single-rank NCCL communicator initializes a large NET shared P2P GPU buffer even though it has no remote peer. In a model serving process where communicators are initialized lazily, the first collective on such a group can fail at request time when GPU memory is already heavily utilized.

This caused an application crash in our workload:

```text
ncclUnhandledCudaError
Failed to CUDA calloc 536870912 bytes
```

### Environment

- NCCL 2.28.9
- PyTorch 2.11.0 + CUDA 13.0
- 8x NVIDIA B300, single node / all-NVLink topology
- Containerized deployment
- No `NCCL_BUFFSIZE`, `NCCL_P2P_NVL_CHUNKSIZE`, or `NCCL_CUMEM_ENABLE` override in the natural observation

### Minimal reproducer

Run with `NCCL_DEBUG=INFO NCCL_DEBUG_SUBSYS=INIT,GRAPH,COLL,ALLOC torchrun --standalone --nproc-per-node=8 repro.py`:

```python
import os

import torch
import torch.distributed as dist

dist.init_process_group("nccl")
rank = dist.get_rank()
world_size = dist.get_world_size()
local_rank = int(os.environ["LOCAL_RANK"])
torch.cuda.set_device(local_rank)

# All ranks create all groups in the same order. Each process then uses its
# own singleton NCCL group.
groups = [dist.new_group([r], backend="nccl") for r in range(world_size)]

x = torch.zeros(1, device=f"cuda:{local_rank}")
dist.all_reduce(x, group=groups[rank])
torch.cuda.synchronize()
```

On the affected topology, the first collective on the singleton group logs:

```text
ncclCommInitRankConfig ... rank 0 nranks 1 ... Init START
P2P Chunksize set to 524288
transport/net.cc:656 Cuda Alloc Size 536870912
64 coll channels, 64 collnet channels, 0 nvls channels,
64 p2p channels, 64 p2p channels per peer
ncclCommInitRankConfig ... rank 0 nranks 1 ... Init COMPLETE
AllReduce ... count 1 ... [nranks=1]
```

The allocation happens before the count-one all-reduce. With sufficient free memory it succeeds; with less than 512 MiB allocatable memory it fails during communicator initialization.

### Source-level size calculation

In NCCL 2.28.9, the allocation comes from `sharedNetBuffersInit()` in `src/transport/net.cc`:

```cpp
#define NCCL_SHARED_STEPS 16
state->size = nChannels * NCCL_SHARED_STEPS * proxyState->p2pChunkSize;
...
ncclCudaCalloc(&state->cudaBuff, state->size);
```

For the observed topology:

```text
64 channels * 16 shared steps * 524288 bytes = 536870912 bytes
```

The call is reached through the unconditional local NET proxy shared initialization in communicator setup:

```text
ncclProxyMsgSharedInit -> proxySharedInit -> sharedNetBuffersInit
```

Ring/tree connection setup already has `nRanks > 1` guards, but the NET proxy shared initialization is still performed for `nranks=1`.

### Expected behavior

A communicator whose fixed size is one cannot communicate with a remote rank; all of its collectives have local identity semantics. Could NCCL skip NET/P2P proxy shared-buffer initialization for `nranks <= 1`, or otherwise provide a low-memory singleton fast path?

The exact allocation size is topology-dependent. The issue is the unnecessary transport initialization for a communicator that can only execute single-rank operations, not specifically the 512 MiB value.

zupengwang:
## Description

Skip eager local NET proxy shared-buffer initialization for single-rank communicators when runtime connect is disabled.

A single-rank communicator has no remote peer that can consume this pool. Communicators with more than one rank keep the existing initialization path unchanged.

## Related Issues

Fixes #2363.

## Changes & Impact

- Guard the local NET shared-buffer initialization with comm->nRanks > 1.
- Preserve the proxy subsystem and all multi-rank initialization behavior.
- No public API or ABI changes.

## Performance Impact

Validated on a single PCIe host with RTX 3090 GPUs, CUDA 13.0, and an SM86-only NCCL build.

- One GPU, NCCL_RUNTIME_CONNECT=0: initialization memory decreased from 996 MiB to 484 MiB (-512 MiB).
- One GPU, default runtime connect: initialization memory remained 484 MiB.
- Two-GPU all-reduce sweep (1 MiB to 256 MiB, 20 warmups, 100 iterations, 5 cycles): average bus bandwidth was 3.816 GB/s before and 3.826 GB/s after (+0.27%).
- Three-GPU all-reduce sweep with the same methodology: average bus bandwidth was 4.940 GB/s before and 4.954 GB/s after (+0.30%).
- All nccl-tests runs reported zero out-of-bounds values.
- 100 separate single-rank initialization/all-reduce/destruction cycles completed under a 20-second per-process timeout.
- A 20-cycle ncclCommSplit test with split resource sharing enabled completed before and after the change. Each cycle split a two-rank parent into two singleton children, ran all-reduce and self send/recv, then destroyed child and parent communicators.
- NCCL and nccl-tests both built successfully; the modified NCCL library also passed a fresh clean build.

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -1681,11 +1681,13 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
       }
     }
 
-    // Connect to local net proxy
-    NCCLCHECKGOTO(ncclProxyConnect(comm, TRANSPORT_NET, 1, comm->rank, &proxyConn), ret, fail);
-    NCCLCHECKGOTO(ncclProxyCallBlocking(comm, &proxyConn, ncclProxyMsgSharedInit, &comm->p2pnChannels, sizeof(int),
-                                        NULL, 0),
-                  ret, fail);
+    // A single-rank communicator cannot have a remote peer, so it does not need NET shared buffers.
+    if (comm->nRanks > 1) {
+      NCCLCHECKGOTO(ncclProxyConnect(comm, TRANSPORT_NET, 1, comm->rank, &proxyConn), ret, fail);
+      NCCLCHECKGOTO(ncclProxyCallBlocking(comm, &proxyConn, ncclProxyMsgSharedInit, &comm->p2pnChannels, sizeof(int),
+                                          NULL, 0),
+                    ret, fail);
+    }
 
     // Then to remote ones when using PXN
     if (ncclPxnDisable(comm) == 0) {


yuke-nvidia:
Thanks for the PR and the detailed testing. We’re still investigating the underlying issue and evaluating a more comprehensive fix. This change may address the reported single-rank case, but we’d like to understand the root cause and broader implications before moving forward. We’ll follow up here once we have an update.

zupengwang:
Thanks for the update. Understood—I'll leave the PR as-is while you investigate. Happy to help with targeted testing or validation of the broader fix when useful.

PR #2369, opened 2026-08-24
Fix profiler KernelCh timestamp drift for communicators with broadcast/AllGatherV

mmhzheng:
Fix profiler KernelCh timestamp drift for broadcast (AllGatherV path)

## Problem
With the profiler enabled (`ncclProfileKernelCh`) and the default
`NCCL_ALLGATHERV_ENABLE=1`, any workload containing `ncclBroadcast`
produced badly corrupted KernelCh profiling data: GPU kernel timestamps
were mapped onto the wrong task, so exported events showed out-of-order
timestamps (end before begin) and bogus durations in the multi-second
range (observed drift of ~3s / ~6s and larger). Setting `NCCL_ALLGATHERV_ENABLE=0` 
made the problem disappear, pointing at the AllGatherV/broadcast path.

## Root cause
A host/device `workCounter` desync:

- When AllGatherV is enabled, a plain `ncclBroadcast` is appended as an
  `ncclTaskBcast` (`collTaskAppend`). A single-peer broadcast
  (`BcastPeers==1`) is then rewritten into a fresh `ncclTaskColl`
  (`ncclPrepareTasks`).
- Both tasks are allocated from a memset-zeroed memory pool, and neither
  set/copied `eActivationMask`, so the `ncclProfileKernelCh` bit was lost.
- Host side (`ncclProfilerPostPlanWork`) only bumps
  `comm->profiler.workCounter[]` for tasks whose `eActivationMask` carries
  `ncclProfileKernelCh`; the device kernel advances `channel.workCounter`
  unconditionally (`src/device/common.h`). Each affected op left the host
  one bump behind the device. The drift accumulated permanently and
  shifted the ring-buffer slot that even correctly-sampled ops (e.g.
  AllGather) read, so they picked up GPU timestamps from the wrong slot.

## Solution
Propagate the profiler activation state along the whole broadcast
scheduling chain, in `src/enqueue/enqueue.cc`:

1. In `collTaskAppend` (AllGatherV/bcast branch): set `eActivationMask`,
   `groupApiEventHandle` and `collApiEventHandle` from
   `ncclProfilerApiState` when the `ncclTaskBcast` is created.
2. In `ncclPrepareTasks` (`BcastPeers==1` bcast->coll conversion): copy the
   same three fields from the bcast task onto the converted coll task.

## Limitations
Only restores propagation of profiler metadata. No changes to the data
path, device kernels, or the profiler plugin ABI. With profiling off
`ncclProfilerApiState.eActivationMask` is already 0, so behavior is
unchanged.

Signed-off-by: zhenghao.cs <zhenghao.cs@bytedance.com>

### DIFF
--- a/src/enqueue/enqueue.cc
+++ b/src/enqueue/enqueue.cc
@@ -440,6 +440,17 @@ ncclResult_t ncclPrepareTasks(struct ncclComm* comm, bool* algoNeedConnect, bool
       t->datatype = bcastTask->datatype;
       // Carry the profiler tag onto the converted coll task so its coll event reports it.
       t->profilerTag = bcastTask->profilerTag;
+      // Carry the profiler activation mask and API event handles onto the converted coll task.
+      // The single-peer broadcast is rewritten into a fresh ncclTaskColl here, so without this
+      // copy its eActivationMask stays memset-zeroed and the ncclProfileKernelCh bit is dropped.
+      // hasProfilerOps would then be false and ncclProfilerPostPlanWork() would never bump the
+      // host workCounter for this task, while the device still advances channel.workCounter for
+      // its work. That drift makes every subsequent KernelCh event read the wrong device slot
+      // (observed as out-of-order timestamps and multi-second bogus durations under
+      // NCCL_ALLGATHERV_ENABLE=1).
+      t->eActivationMask = bcastTask->eActivationMask;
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
++ @armratner @jynv can you check and mirror if this is good? Thanks!

PR #2374, opened 2026-08-26
bootstrap: close accepted sockets on error and abort paths

EylonKrause:
## Summary

Closes accepted / listening sockets that were leaked on error and abort paths in
the bootstrap layer. All changes are in `src/bootstrap.cc`.

### 1. `bootstrapRoot`: accepted socket leaked on `socketRecv` failure

In the root's receive loop the code accepts a connection into a local `sock` and
then does `socketRecv(&sock, &info, ...)`. On failure the `goto out` left the
just-accepted socket open, leaking its fd for every rank whose `extInfo` recv
fails. Now the socket is closed before bailing:

```c
if ((res = socketRecv(&sock, &info, sizeof(info))) != ncclSuccess) {
  (void)ncclSocketClose(&sock);
  goto out;
}
```

### 2. `bootstrapInit`: accepted socket and root listen socket leaked on `socketRecv` failure

When receiving the ring "next peer" info from the root, the code accepts into
`sock` and receives into it with `NCCLCHECK`, which `return`s on error without
closing either `sock` (accepted) or `listenSockRoot` (the root listen socket).
Both are now closed on failure. Note it jumps to `exit`, not `fail`: at this
point `proxySocket` is not yet allocated, and the `fail` label does
`free(proxySocket)` on that uninitialized pointer.

### 3. `unexpectedFree`: pending unexpected-connection sockets leaked on abort/destroy

`unexpectedFree` walked the `unexpectedConnections` list freeing each element but
never closed the socket embedded in it, leaking one fd per pending unexpected
connection whenever a comm is aborted/destroyed with connections still queued.
Each element's `sock` is now closed before the element is freed, and the list
head is set to `NULL` afterwards.

`ncclSocketClose` is a safe no-op on an already-closed / invalid socket, so these
additions are safe on every path that reaches them.

## Verification

Verified by static analysis against master (`7b83616d`, v2.31.2-1). I don't have
multi-GPU/NVSwitch hardware or a full NCCL build on my dev box, so these paths
were not runtime-tested; the changes are self-contained cleanup-path edits.

### DIFF
--- a/src/bootstrap.cc
+++ b/src/bootstrap.cc
@@ -300,6 +300,7 @@ static void* bootstrapRoot(void* rargs) {
   struct extInfo info;
   union ringConnectInfo* rankInfo = NULL;
   union ncclSocketAddress* rankAddressesRoot = NULL; // for initial rank <-> root information exchange
+  struct ncclSocket sock;
   // get zeros for comparison
   char zeroHandle[NCCL_NET_HANDLE_MAXSIZE];
   union ncclSocketAddress zeroAddress;
@@ -312,8 +313,8 @@ static void* bootstrapRoot(void* rargs) {
   TRACE(NCCL_BOOTSTRAP, "BEGIN");
   BOOTSTRAP_PROF_OPEN(timers[BOOTSTRAP_INIT_ROOT_WAIT]);
   /* Receive addresses from all ranks */
+  NCCLCHECKGOTO(ncclSocketInit(&sock), res, out);  // so out: can always close it
   do {
-    struct ncclSocket sock;
     NCCLCHECKGOTO(ncclSocketInit(&sock), res, out);
     NCCLCHECKGOTO(ncclSocketAccept(&sock, listenSock), res, out);
     NCCLCHECKGOTO(socketRecv(&sock, &info, sizeof(info)), res, out);
@@ -393,6 +394,7 @@ static void* bootstrapRoot(void* rargs) {
         timers[BOOTSTRAP_INIT_ROOT_WAIT] / 1e9, timers[BOOTSTRAP_INIT_ROOT_RECV] / 1e9,
         timers[BOOTSTRAP_INIT_ROOT_SEND] / 1e9);
 out:
+  (void)ncclSocketClose(&sock);  // no-op unless we bailed out with an accepted connection open
   if (listenSock != NULL) {
     (void)ncclSocketClose(listenSock);
     free(listenSock);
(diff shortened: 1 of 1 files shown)


src/bootstrap.cc
```diff
@@ -313,7 +313,11 @@ static void* bootstrapRoot(void* rargs) {
     struct ncclSocket sock;
     NCCLCHECKGOTO(ncclSocketInit(&sock), res, out);
     NCCLCHECKGOTO(ncclSocketAccept(&sock, listenSock), res, out);
-    NCCLCHECKGOTO(socketRecv(&sock, &info, sizeof(info)), res, out);
+    if ((res = socketRecv(&sock, &info, sizeof(info))) != ncclSuccess) {
+      // Close the just-accepted connection before bailing out to avoid leaking its fd.
+      (void)ncclSocketClose(&sock);
+      goto out;
+    }
```
kiskra-nvidia: This is generally not how we prefer to do error handling in NCCL. Could you move this clean-up code to `out`? I recognize that this would require moving the `sock` variable outside the loop and also initializing it there (so that we don't call `ncclSocketClose` without a prior `ncclSocketInit`).

Alternatively, do something like this:

```
    NCCLCHECKIGNORE(socketRecv(&sock, &info, sizeof(info)), res);
    if (res != ncclSuccess) {
      (void)ncclSocketClose(&sock);
      goto out;
    }
```

The advantage of using `NCCLCHECKIGNORE` is that the line will be included in the error backtrace.

src/bootstrap.cc
```diff
@@ -790,7 +794,14 @@ ncclResult_t bootstrapInit(int nHandles, void* handles, struct ncclComm* comm, s
...
     NCCLCHECK(ncclSocketAccept(&sock, &listenSockRoot));
-    NCCLCHECK(socketRecv(&sock, &nextPeer, sizeof(nextPeer)));
+    if ((result = socketRecv(&sock, &nextPeer, sizeof(nextPeer))) != ncclSuccess) {
+      // Close both the accepted connection and the root listen socket before bailing
+      // out to avoid leaking their fds. proxySocket is not yet allocated here, so jump
+      // to exit (fail would free the uninitialized proxySocket).
+      (void)ncclSocketClose(&sock);
+      (void)ncclSocketClose(&listenSockRoot);
+      goto exit;
+    }
```
kiskra-nvidia: Again, it would be much more robust to move this cleanup to `fail`. For that to be effective though, most `NCCLCHECK` uses in this function should be replaced by `NCCLCHECKGOTO` (yeah, the existing error handling is far from perfect...). Or, as before, use `NCCLCHECKIGNORE`.

src/bootstrap.cc
```diff
@@ -1059,8 +1070,11 @@ static void unexpectedFree(struct bootstrapState* state) {
   while (elem) {
     prev = elem;
     elem = elem->next;
+    // Close the pending unexpected connection's socket before freeing to avoid leaking its fd.
+    (void)ncclSocketClose(&prev->sock);
```
kiskra-nvidia, line 1277: I believe this one we've already fixed. Can you check on the [staging](https://github.com/NVIDIA/nccl/tree/staging) branch?

EylonKrause:
Thanks — reworked along the lines you suggested, and rebased onto `dev` (retargeted the PR) since the mirror to dev would otherwise conflict:

- **bootstrapRoot:** `sock` is hoisted out of the receive loop and initialized before it, and `out:` now closes it — a no-op unless we bailed out with an accepted connection open. The per-iteration `ncclSocketInit` stays so each accept starts clean.
- **bootstrapInit:** the cleanup moved to `fail:`, which now closes the accepted socket and the root listen socket before freeing `proxySocket`. To make that safe on every path I initialize both sockets up front, NULL-initialize `proxySocket`, and converted the function's `NCCLCHECK`s to `NCCLCHECKGOTO(..., result, fail)`. The three locals those earlier jumps would have crossed (`offset`, `curr_root`, `nRankRoot`) are declared at the top, otherwise the gotos would not compile.
- **unexpectedFree:** you're right, `dev`/`staging` already close the socket there — dropped.

Caveat as before: I can't build or run NCCL here, so this was checked by inspection (plus an independent review pass for goto/initialization and ownership issues), not by a compile. Happy to adjust if the build or your tests flag anything.

PR #2376, opened 2026-08-27
transport: close fds on the error paths of proxy connect and UDS import

EylonKrause:
## Problem

Three proxy-side paths open an fd and close it on the success path only, so a failure of the intervening call leaks the fd:

- **`net.cc`** `sendProxyConnect` / `recvProxyConnect`: the DMA-BUF fd from `cuMemGetHandleForAddressRange` is closed *after* `regMrDmaBuf`, but `NCCLCHECK(regMrDmaBuf(...))` returns first when registration fails (a NIC MR/registration limit, a transient plugin error), skipping `close(dmabuf_fd)`. The register-path siblings were already fixed this way in #2267; the connect paths were not.
- **`shm.cc`** `ncclShmImportShareableBuffer` and **`p2p.cc`** `p2pMap`: the UDS fd from `ncclProxyClientGetFdBlocking` is closed *after* `cuMemImportFromShareableHandle`, but `CUCHECK(cuMemImportFromShareableHandle(...))` returns first on failure, skipping the close.

Each leaked fd is per proxy-connect / per import, so a flapping NIC or a driver error under load accumulates descriptors.

## Fix

Capture the call result, close the fd unconditionally, then propagate the error — so the fd is released on both success and failure. Same shape as #2267.

## Testing

Static change (NCCL isn't buildable on my dev box). The edits only move the existing `close` ahead of the error check; the success path is unchanged, and the failure path now closes the fd before returning the same error. Brace balance verified on all three files.

### DIFF
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -1014,10 +1014,11 @@ static ncclResult_t sendProxyConnect(struct ncclProxyConnection* connection, str
         CUCHECK(cuMemGetHandleForAddressRange((void*)&dmabuf_fd, (CUdeviceptr)resources->buffers[p], dmaBufSize,
                                               CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD,
                                               getHandleForAddressRangeFlags(resources->useGdr)));
-        NCCLCHECK(proxyState->ncclNet->regMrDmaBuf(resources->netSendComm, resources->buffers[p],
+        ncclResult_t dmaBufRegRes = proxyState->ncclNet->regMrDmaBuf(resources->netSendComm, resources->buffers[p],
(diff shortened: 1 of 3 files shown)


xiaofanl-nvidia:
++ @AddyLaddy to take a look.

AddyLaddy:
Thanks LGTM, I'll get this merged to our dev tree.

AddyLaddy:
@EylonKrause the mirror failed, can you try and rebase it against dev ?

EylonKrause:
Rebased the fix onto `dev` and retargeted the PR to it, so the diff is just this change. dev had moved the UDS import to `ncclIpcFd`/`ncclIpcFdClose`, and it still runs `CUCHECK` before the close, so the same leak was there — the rebased version keeps the close on both paths using the new API.

PR #2382, opened 2026-08-31
feat(efa-gda): Implement Get

akkart-aws:
## Description

Get was a stub.

## Changes & Impact

Get rides the data endpoint. The plugin binds that endpoint's local completion counter FI_WRITE | FI_READ, so a read's completion is visible to the SQ ring-reuse backpressure and to the blocking Flush.

A read and a write populate identical WR fields — the RDMA address pair holds the remote address and rkey, the SGE holds the local buffer — and only the opcode differs, deciding which way the bytes move. So the whole post path is reusable: parameterize it on the operation instead of duplicating it.

- getImpl dispatches on the resource-sharing mode, like putImpl.

- getImplMode resolves the peer's window as the source and this rank's window as the destination, addresses the peer's data endpoint (target slot 0, the endpoint that serves window memory), and chunks the fetch at the EFA per-operation limit. A zero-byte fetch posts one zero-length read, so flush and flushAsync/wait have a completion to terminate on.

- postRdmaWrite becomes postRdmaOp, templated on efaGdaRdmaOp op, and selects the read or write WR builder. Every other line of it, including warp coalescing, the doorbell rendezvous, req_id stamping and both backpressure gates, is unchanged and now serves Get too.

### DIFF
--- a/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
+++ b/src/include/nccl_device/gin/efa_gda/gin_efa_gda.h
@@ -9,9 +9,9 @@
  *
  * Implemented: Put (data + signal/counter endpoints, signal-only via
  *              scratch buffer), PutValue (value staged through the
- *              per-endpoint slot pool), Flush, GetSignalPtr,
+ *              per-endpoint slot pool), Get, Flush, GetSignalPtr,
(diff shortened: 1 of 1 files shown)


kgioioso:
Merged

PR #2386, opened 2026-09-03
graph: size the fused topology XML for all local ranks

### LINKED ISSUE #2160: NCCL_TOPO_XML_MAX_NODES=256 limit hit during intra-node XML fusion on 32-NIC hosts (AWS p5.48xlarge)
# Issue: hit `NCCL_TOPO_XML_MAX_NODES=256` limit during intra-node XML fusion on 32-NIC hosts (AWS p5.48xlarge)

## Summary

On AWS p5.48xlarge (8 x H100, 32 EFA NICs), `ncclTopoFuseXml` hits its 256-node XML buffer cap while fusing the 8 local-rank XMLs at initialization. The error fires as:

```
graph/xml.h: NCCL WARN Error : too many XML nodes (max 256)
```

Each rank autogenerates ~126 XML nodes from its `/sys/class/pci_bus/*` walk on this host shape; per-rank dedup during `ncclTopoFuseXmlRecursive` is insufficient to hold the merged output below 256 nodes on our reproducer.

## Reproducer

- Hardware: AWS p5.48xlarge (8 GPUs, 32 EFA NICs)
- NCCL: 2.30.4 (`nvidia-nccl-cu13==2.30.4` pip wheel)
- Plugin: stock aws-ofi-nccl v1.19.1, no topology patches, no `NCCL_TOPO_FILE` override
- Workload: DeepSeek DeepEP V2 `tests/elastic/test_ep.py`. We expect any workload that exercises the same intra-node XML fusion path on this host shape to reproduce, but we have not attempted a minimal `nccl-tests` repro.

The overflow fires during early init, before any collective.

## Observation vs. inferred mechanism

**Observed:** The overflow reproduces consistently on this host shape with the stock setup above. We additionally tried supplying an identical 122-node `NCCL_TOPO_FILE` to every rank; the overflow still fires. We also tried narrowing visible EFA devices via libfabric (`FI_EFA_DEVICE_LIST`); the overflow still fires, because NCCL walks `/sys/class/pci_bus/*` independently of what libfabric exposes.

**Inferred (from reading the 2.30.4 source, not a live-debugger trace):** The non-MNNVL branch of `ncclTopoGetSystem` reuses the initial 256-slot XML buffer for the fused output rather than allocating a larger buffer (the MNNVL branch does allocate a larger buffer). This path relies on `xmlFindNode` dedup in `ncclTopoFuseXmlRecursive` to keep the merged buffer within 256. `xmlFindNode` (static inline in `src/graph/xml.h`) requires `nAttrs` match plus every attribute key/value match, and per-rank generated attrs on the `gpu`/`pci` nodes may be preventing sufficient collapse of otherwise-equivalent subtrees. We did not step through a live debugger to confirm which specific attribute difference blocks which dedup call.

## Candidate fixes

**(a) Allocate a larger destination XML buffer in the non-MNNVL intra-node fuse path** (analogous to the MNNVL path). Architecturally symmetric.

**(b) Bump `NCCL_TOPO_XML_MAX_NODES`** (the one-line `#define` change). Simpler but increases the buffer footprint for every topology-load path.

**(c) Fix the dedup** in `xmlFindNode` / `ncclTopoFuseXmlRecursive` so rank-specific attributes do not block collapse of otherwise-identical subtrees. We are not familiar enough with the invariants your dedup code relies on to recommend a specific change here.

We tested option (b) locally (`#define` 256 -> 2048) across three DeepEP-based images on the reproducer hardware: 351 combined `test_ep.py` config runs produced zero `too many XML nodes` warnings. That validates option (b) works for our workload; it does not say anything about the structural correctness of other paths in NCCL that allocate XML buffers.

## Happy to send a minimal PR

If the maintainers prefer option (a), I am happy to attempt a PR, though I recognize dynamic allocation in the fuse path may require touching the underlying `ncclXml` struct if it embeds a fixed-size array. If you prefer option (b), it is a 1-line bump. Either way I'd rather not guess which shape you'd accept without a maintainer nudge.

## Context / incident writeup

Full incident writeup with raw logs, controlled experiments, and the source-reading chain:
https://github.com/antonai-work/deepep-v2-efa-base/blob/main/docs/INCIDENT-PR1226-AWS-OFI-NCCL-XML-OVERFLOW.md

Thanks for your work on NCCL.

kodlan:
## Description

Fixes the "too many XML nodes (max 256)" failure in intra-node topology fusion.

In ncclTopoGetSystem each rank builds its own topology XML in a 256 node buffer, the local ranks exchange them and the result is fused into one tree. The MNNVL path allocates a bigger buffer for the fused tree (nLocalRanks x 256 nodes) but the intra-node path just reused the 256 node per-rank buffer. The fused tree is the union of all local trees though. Nodes only dedup when every attribute matches and the gpu node carries a per-rank "rank" attribute, so each rank's gpu subtree gets appended, plus anything else that differs per rank. On big hosts this goes past 256 and xmlAddTree fails, so ncclCommInitRank returns ncclInternalError.

This just makes the enlargement unconditional, same thing the MNNVL path already did.

## Related Issues

Fixes #2160

## Changes & Impact

One place in src/graph/topo.cc. The fused xml is now always allocated with nLocalRanks x NCCL_TOPO_XML_MAX_NODES nodes. The MNNVL path is unchanged in behavior, the intra-node path gets the same buffer size.

Tested on one GPU (RTX 2070) with several ranks on the same device (NCCL_MULTI_RANK_GPU_ENABLE=1) and a topology file that gives the gpu 18 nvlink children, so each rank contributes 19 nodes that cannot dedup: 12 ranks fuse to 233 nodes and pass, 14 ranks (271 nodes) fail before this change and pass after it. The fused NCCL_TOPO_DUMP_FILE output for 2 and 8 ranks is byte identical before and after. all_reduce_perf with -c 1 passes. Built with WERROR=1 and checked with an ASAN build. I could not test on a real multi GPU host, MNNVL, or with the EFA/IB plugins.

Out of scope: the 256 node limit when loading a topology file (NCCL_TOPO_FILE) is unchanged, and the warning text is the same.

## Performance Impact

Init only. The fused xml is freed at the end of ncclTopoGetSystem, so it is a transient host allocation of about 3.6 MB per local rank (8 GPUs: ~29 MB per process for the duration of topology detection). No change to the data path.

### DIFF
--- a/src/graph/topo.cc
+++ b/src/graph/topo.cc
@@ -1958,15 +1958,11 @@ ncclResult_t ncclTopoGetSystem(struct ncclComm* comm, struct ncclTopoSystem** sy
   NCCLCHECKGOTO(bootstrapIntraNodeAllGather(comm->bootstrap, localRanks, localRank, nLocalRanks, mem,
                                             xmlMemSize(NCCL_TOPO_XML_MAX_NODES)),
                 ret, fail);
-  if (comm->MNNVL) {
-    // Ensure that we have enough room when fusing topos from multiple nodes.
-    free(xml);
-    xml = NULL;
-    NCCLCHECKGOTO(xmlAlloc(&xml, nLocalRanks * NCCL_TOPO_XML_MAX_NODES), ret, fail);
-  } else {
-    // In the intra-node case there's no need to enlarge the topo xml.
-    xml->maxIndex = 0;
-  }
+  // Ensure that we have enough room when fusing the topos: subtrees with per-rank attributes (e.g. the gpu
+  // nodes) do not dedup, so the fused topo can exceed NCCL_TOPO_XML_MAX_NODES even in the intra-node case.
+  free(xml);
+  xml = NULL;
+  NCCLCHECKGOTO(xmlAlloc(&xml, nLocalRanks * NCCL_TOPO_XML_MAX_NODES), ret, fail);
   for (int i = 0; i < nLocalRanks; i++) {
     struct ncclXml* peerXml = (struct ncclXml*)(mem + xmlMemSize(NCCL_TOPO_XML_MAX_NODES) * i);
     NCCLCHECKGOTO(ncclTopoConvertXml(peerXml, (uintptr_t)peerXml->nodes, 0), ret, fail);


kodlan:
> /mirror master

Can this PR be closed now? How does this mirroring work here?

marksantesson:
> Can this PR be closed now? How does this mirroring work here?

Thanks for asking, @kodlan ! But, no, please don't close it yet, although if you want to you can. "/mirror master" is the first step of us taking a PR and integrating it with the internal NCCL source control. Eventually the change should make it back out as a commit on github. But that usually takes O(day) and we keep the PR open until it has made it back out, or sometimes until there has been a new release with the commit.

When it makes it out, we will usually post a comment back to you requesting that you test that the commit solves your problem. Thanks for the contribution!

marksantesson:
@kodlan , the change has made it to github's dev branch. Please check out [this sha](https://github.com/NVIDIA/nccl/commit/4ae8e52892b1362bbeda3b38a83e2f342d12a74c) and verify that it works for you. I took your original change and adjusted it to allocate to the minimum size, so it should take a bit less memory. If it works for you, then you can close the PR.

xiaofanl-nvidia:
Leaving this open for a week for feedback before I close this out.

kodlan:
> @kodlan , the change has made it to github's dev branch. Please check out [this sha](https://github.com/NVIDIA/nccl/commit/4ae8e52892b1362bbeda3b38a83e2f342d12a74c) and verify that it works for you. I took your original change and adjusted it to allocate to the minimum size, so it should take a bit less memory. If it works for you, then you can close the PR.

Thanks, it works for me, but I guess it's also worth asking the author of the original issue.

xiaofanl-nvidia:
Closing since this has been merged.

PR #2387, opened 2026-09-03
os: skip truncated shared objects before dlopen

### LINKED ISSUE #2367: [Issue]: dlopen() of a truncated plugin .so causes SIGBUS crash instead of graceful fallback (NCCL_PROFILER_PLUGIN / NET / TUNER / ENV)

Application crash

### The two traces below are the same fault
UCX merely installed a signal handler in the multi-rank run — it is not
involved in the crash path. The gdb trace is the same fault caught directly,
with no UCX and no PyTorch present.

### UCX-caught crash signature (16-rank DDP job, NCCL 2.30.1):
[worker-0:xxxxx:0:xxxxx] Caught signal 7 (Bus error: nonexistent physical address)
==== backtrace ====
 ... libc.so.6(dlopen+0x6f)
 ... libnccl.so.2(pncclCommInitRankConfig+0x298)
 ... (torch/PyTorch frames)
=================================

### gdb, catching the fault directly (minimal repro, single GPU, no PyTorch, no UCX):
Thread 1 "all_reduce_perf" received signal SIGBUS, Bus error.
memset () at ../sysdeps/x86_64/multiarch/../multiarch/memset-vec-unaligned-erms.S:392
#0  memset ()
#1  0x0000155555523f0f in _dl_map_segments (header=0x1555405f0018, ...) at ./dl-map-segments.h:176
#2  _dl_map_object_from_fd (name="/path/to/trunc_2000.so", ...) at ./elf/dl-load.c:1258
#3  0x0000155555525529 in _dl_map_object (name="/path/to/trunc_2000.so", type=2, ...) at ./elf/dl-load.c:2268
#4  0x00001555555298dc in dl_open_worker_begin (...) at ./elf/dl-open.c:578

Fault address: 0x1555405f0008
info proc mappings (surrounding region):
  0x1555405ef000  0x1555405f1000  size 0x2000  offset 0x2000  rw-p  trunc_2000.so

### readelf confirms the file is structurally inconsistent:
$ readelf -l trunc_2000.so
readelf: Error: the dynamic segment offset + size exceeds the size of the file
  LOAD  0x0000000000002e68 0x0000000000003e68 ... FileSiz 0x1a0  RW
  DYNAMIC 0x0000000000002e78 ... FileSiz 0x150   <- offset+size = 12232, file is only 2000 bytes

Minimal repro (single GPU, single process, bare command — no container,
no multi-node, no plugin-specific code):

1. Build any trivial valid shared library — no NCCL symbols needed at all:
     echo 'int f(int x){return x+1;}' > t.c
     gcc -shared -fPIC -o t.so t.c

2. Confirm baseline: NCCL loads and gracefully skips a valid-but-irrelevant .so
   (no profiler symbols found -> plugin disabled, job runs fine):
     NCCL_PROFILER_PLUGIN=/path/to/t.so ./build/all_reduce_perf -g 1 -b 8 -e 8 -n 1

3. Truncate the SAME file to a length that leaves its own program headers
   describing segments beyond the new (shorter) EOF. `readelf -l t.so` will
   flag this itself ("... exceeds the size of the file") -- for a ~15KB
   library, 2000 bytes reproduces it reliably:
     truncate -s 2000 t.so

4. Re-run the identical command:
     NCCL_PROFILER_PLUGIN=/path/to/t.so ./build/all_reduce_perf -g 1 -b 8 -e 8 -n 1

Observed: process dies with SIGBUS inside dlopen() (see backtrace above),
          instead of the graceful "plugin not usable, continuing without it"
          behavior seen in step 2.

Scope — full malformed-state matrix (tested on both NCCL versions,
identical results on each):

| State                                            | File size | Result         |
|--------------------------------------------------|-----------|----------------|
| valid .so (control)                              | 15128     | graceful skip  |
| truncated to 100 bytes                           | 100       | graceful skip  |
| truncated to 2000 bytes                          | 2000      | SIGBUS crash   |
| truncated to half                                | 7564      | SIGBUS crash   |
| full size, random bytes at offset 8000-10000     | 15128     | graceful skip  |
| zero-length                                      | 0         | graceful skip  |
| non-ELF (text file renamed .so)                  | 51        | graceful skip  |
| valid ELF header, all program headers zeroed     | 15128     | graceful skip  |

Only files whose program headers describe segments extending past the
actual file length crash. Too-short files (can't contain the program-header
table) and same-size corruption both fail cleanly.

Other notes:
  - Reproduces on NCCL 2.28.9 and 2.30.1 identically -- longstanding
    behavior in the shared plugin-loader code, not a version regression.
  - Reproduces with a plugin file that never mentions NCCL/profiler symbols
    at all -- generic dlopen()/ELF-loading issue, not profiler-specific.
    By code inspection (see root cause) equally reachable via
    NCCL_NET_PLUGIN, NCCL_TUNER_PLUGIN, NCCL_ENV_PLUGIN.
  - Reproduces with a single GPU, single rank, single process.
  - Not intermittent: 100% reproducible.

### NCCL Version

2.30.1+cuda13.0 (host install, driver 580.159.04) and 2.28.9+cuda12.8 (built from source, same repro)

### Your platform details

The minimal repro above needs none of the following -- it reproduces with a
bare all_reduce_perf on one GPU. Listed for completeness only.

GPU: 8x NVIDIA H200 per node, 2 nodes, driver 580.159.04
Topology: GPUs via NVLink (NV18), NICs at NODE/PHB/PIX distance
Environment: originally observed under Slurm-on-Kubernetes (Soperator),
  container nvcr.io/nvidia/pytorch:25.01-py3 via pyxis/enroot, with host
  NCCL shadow-mounted to test 2.30.1; NCCL 2.28.9 built from source and
  selected via LD_LIBRARY_PATH. Neither the container nor the mount is
  required to reproduce.
Scalability: irrelevant -- crash occurs during dlopen() of the plugin path
  before any collective communication begins. Reproduces from 1 rank up to
  16 ranks (2 nodes x 8 GPUs) identically.

### Error Message & Behavior

NCCL already degrades gracefully for every other unusable plugin file --
missing, zero-length, non-ELF, or a valid ELF with no matching plugin
symbols. In all of those cases it logs that the plugin couldn't be
loaded/matched and continues without it. A truncated ELF file is the one
malformed state that does NOT degrade gracefully: it crashes the whole
process instead. This report is about closing that single gap in otherwise
consistent behavior.

First error: no NCCL WARN is ever emitted -- the process receives SIGBUS
directly from the dynamic loader before NCCL's own plugin-load logging
path executes. The only visible signal is the crash handler output:
  "Caught signal 7 (Bus error: nonexistent physical address)"
followed by a backtrace through dlopen() -> pncclCommInitRankConfig.

Actual: SIGBUS terminates the process. Under torch.distributed/torchrun,
this surfaces as ChildFailedError and kills the entire distributed job
(observed: 13/16 ranks died this way in a 2-node DDP job when the
profiler plugin binary was corrupted on disk).

Root cause (from src/plugin/plugin_open.cc, NCCL 2.28.9 source):
tryOpenLib() calls dlopen(name, RTLD_NOW | RTLD_LOCAL) directly, with no
stat()/access()/size/ELF-magic validation beforehand:

    static void* tryOpenLib(char* name, int* err, char* errStr) {
      ...
      void *handle = dlopen(name, RTLD_NOW | RTLD_LOCAL);   // line 37

This function is shared by all four plugin types (NET, TUNER, PROFILER,
ENV) via openPluginLib(), so the exposure is not specific to the profiler
path. RTLD_NOW forces eager relocation at dlopen() time, which is when
glibc's _dl_map_segments() memset()s the bss-tail of the last LOAD segment
and faults on the unbacked (truncated-away) page.

Suggested fix direction: before calling dlopen() on a user-supplied plugin
path, verify the file is self-consistent -- specifically that
max(p_offset + p_filesz) across program headers does not exceed the actual
file size (per the matrix above, this is exactly the condition that
separates the crashing cases from the gracefully-handled ones). On
mismatch, treat it the same as any other unloadable plugin: log and
continue without it.

kodlan:
## Description

A plugin .so that got truncated (copy interrupted, disk full, partial download) makes NCCL die with SIGBUS inside dlopen(). glibc checks that the ELF headers fit in the file but not that the PT_LOAD segments do, so it maps the missing part and faults while zero-filling the last page. There is no error to catch, the process is just gone, with no NCCL log line. Every other kind of broken plugin (missing, empty, not ELF, no symbols) is logged and skipped, this one was the exception.

This adds a best effort check in ncclOsDlopen (Linux only): when NCCL can tell which file dlopen is going to open, meaning the name is a path or a bare name that sits in one of the LD_LIBRARY_PATH directories, it reads the ELF program headers first and refuses the file if any PT_LOAD segment ends past the end of the file. The refusal goes through the normal dlerror path, so the plugin loader logs "Plugin: <name>: <path>: file is truncated (a segment ends beyond the end of the file)" and falls back like for any other unloadable plugin. Objects that are already loaded are not checked (dlopen does not touch the file for those), files the loader would skip (unreadable, wrong ELF class) are skipped the same way, and names that come from RPATH, hwcaps subdirs, ld.so.cache or the system directories are passed to dlopen unchanged.

## Related Issues

Fixes #2367

## Changes & Impact

src/os/linux.cc only: two static helpers plus one line at the top of each ncclOsDlopen overload, about 70 lines. Windows untouched. The check only runs on files it can find, and only rejects files that would either crash dlopen or load with silently truncated data, so a plugin that loads today still loads.

Tested on one GPU with a tiny .so truncated to 2000 bytes: before, NCCL_PROFILER_PLUGIN, NCCL_NET_PLUGIN, NCCL_TUNER_PLUGIN, NCCL_ENV_PLUGIN and a truncated libnccl-net.so found through LD_LIBRARY_PATH all killed the process with signal 7; after, all five are logged and skipped and the communicator comes up. A size sweep shows the check rejects every size between the end of the program headers and the end of the last segment, glibc's own "file too short" still handles smaller files, and intact files load. Also checked: empty LD_LIBRARY_PATH entry, ';' as separator, an unreadable copy earlier in the path, an already loaded copy (LD_PRELOAD with SONAME), directory or text or empty file in place of the plugin, missing file, zeroed program headers (all give the same dlopen messages as before), NVML still loads, the net and tuner example plugins load, WERROR=1 build, ASAN build, all_reduce_perf -c 1. Not tested: Windows, plugins resolved via RPATH or ld.so.cache (those are simply not pre-checked).

## Performance Impact

None on the data path. One open/fstat and two small reads per dlopen at init.

### DIFF
--- a/src/os/linux.cc
+++ b/src/os/linux.cc
@@ -12,6 +12,7 @@
 
 #include <cstdint>
 #include <dlfcn.h>
+#include <elf.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -48,7 +49,75 @@ static void saveDlError() {
   }
 }
 
+// dlopen() of a shared object truncated inside one of its PT_LOAD segments kills the process with SIGBUS (glibc
+// maps the missing part and faults while zero-filling it), so refuse such files before loading them.
+// Returns 1 if the file is truncated, 0 if it looks loadable, -1 if dlopen() would skip it and search on.
+static int dlFileCheck(const char* path) {
+  int fd = open(path, O_RDONLY | O_CLOEXEC);
+  if (fd < 0) return -1;
+  int res = 0;
+  struct stat st;
+  Elf64_Ehdr ehdr;
+  if (fstat(fd, &st) == 0 && S_ISREG(st.st_mode) && pread(fd, &ehdr, sizeof(ehdr), 0) == (ssize_t)sizeof(ehdr) &&
+      memcmp(ehdr.e_ident, ELFMAG, SELFMAG) == 0) {
+    if (ehdr.e_ident[EI_CLASS] != ELFCLASS64) {
+      res = -1;
+    } else if (ehdr.e_phentsize == sizeof(Elf64_Phdr)) {
+      for (int i = 0; i < ehdr.e_phnum && res == 0; i++) {
+        Elf64_Phdr phdr;
+        off_t off = (off_t)ehdr.e_phoff + (off_t)i * (off_t)sizeof(phdr);
+        // Headers past the end of the file are rejected by dlopen() itself.
+        if (pread(fd, &phdr, sizeof(phdr), off) != (ssize_t)sizeof(phdr)) break;
+        if (phdr.p_type == PT_LOAD &&
+            (phdr.p_offset > (uint64_t)st.st_size || phdr.p_filesz > (uint64_t)st.st_size - phdr.p_offset)) {
+          res = 1;
+        }
+      }
+    }
+  }
+  close(fd);
+  return res;
+}
+
+// Best effort: only names we can resolve ourselves are checked, i.e. paths and bare names found in LD_LIBRARY_PATH.
+// dlopen() also consults RPATH/RUNPATH, hwcaps subdirectories and ld.so.cache, which are not emulated here.
+static bool dlCheckFile(const char* filename) {
+  if (filename == NULL) return true;
+  // Already loaded: dlopen() will not touch the file.
+  void* loaded = dlopen(filename, RTLD_NOLOAD | RTLD_LAZY);
+  if (loaded) {
+    dlclose(loaded);
+    return true;
+  }
+  char path[PATH_MAX];
+  const char* found = NULL;
+  int res = 0;
+  if (strchr(filename, '/')) {
+    found = filename;
+    res = dlFileCheck(found);
+  } else {
+    const char* dirs = getenv("LD_LIBRARY_PATH");
+    while (dirs && found == NULL) {
+      const char* end = strpbrk(dirs, ":;");
+      int len = end ? (int)(end - dirs) : (int)strlen(dirs);
+      // An empty entry stands for the current directory.
+      if (snprintf(path, sizeof(path), "%.*s/%s", len ? len : 1, len ? dirs : ".", filename) < (int)sizeof(path)) {
+        res = dlFileCheck(path);
+        if (res >= 0) found = path;
+      }
+      dirs = end ? end + 1 : NULL;
+    }
+  }
+  if (res == 1) {
+    snprintf(ncclDlErrorBuf, sizeof(ncclDlErrorBuf),
+             "%.180s: file is truncated (a segment ends beyond the end of the file)", found);
+    return false;
+  }
+  return true;
+}
+
 ncclOsLibraryHandle ncclOsDlopen(const char* filename) {
+  if (!dlCheckFile(filename)) return NULL;
   ncclOsLibraryHandle handle = dlopen(filename, RTLD_NOW | RTLD_LOCAL);
   if (handle == NULL) {
     saveDlError();
@@ -69,6 +138,7 @@ const char* ncclOsDlerror() {
 }
 
 ncclOsLibraryHandle ncclOsDlopen(const char* path, int mode) {
+  if (!dlCheckFile(path)) return NULL;
   ncclOsLibraryHandle handle = dlopen(path, (mode == NCCL_OS_DL_NOW) ? RTLD_NOW : RTLD_LAZY);
   if (handle == NULL) saveDlError();
   return handle;


xiaofanl-nvidia:
++ @MoraruMaxim

MoraruMaxim:
Before moving forward, I think we first need to agree on where this fix should reside. This is currently being discussed in [issue #2367](https://github.com/NVIDIA/nccl/issues/2367).

Once we reach a conclusion there, we can determine how to proceed with this PR.

MoraruMaxim:
Following the discussion in [#2367](https://github.com/NVIDIA/nccl/issues/2367), we have decided not to address this at the NCCL level and are therefore closing this PR. Thank you for your contribution.

kodlan:
> Following the discussion in [#2367](https://github.com/NVIDIA/nccl/issues/2367), we have decided not to address this at the NCCL level and are therefore closing this PR. Thank you for your contribution.

ok, thanks!

PR #1973, opened 2026-01-07
Bug fix: profiler kernelch event type field

CavalryXD:
## Description

I run NCCL Profiler to check the kernelCh event in single node with 8 * A100 GPUs. And I just clone the master branch code from NCCL. And I got the unexpected results where most of kernelCh events have stopTs 0 and StopGpuClk 0. 
I use the recommended compilation, just make.
The launch script is follows:
```
export NCCL_PROFILER_PLUGIN=example
export NCCL_PROFILE_EVENT_MASK=127
export LD_LIBRARY_PATH=/workspace/nccl/ext-profiler/example:$LD_LIBRARY_PATH
export NCCL_DEBUG=INFO
export LD_LIBRARY_PATH=/usr/local/mpi/lib:$LD_LIBRARY_PATH
export LD_LIBRARY_PATH=/usr/local/ucx/lib:$LD_LIBRARY_PATH
export NCCL_PROFILE_DUMP_FILE=/workspace/profiler-test/
all_reduce_perf -b 128M -e 128M -g 8
```

Then the event result is(Show only a portion to save space.)
```
{"name": "Group API", "cat": "GROUP_API", "ph": "b", "id": 0, "pid": 75087, "tid": 1, "ts": 59338.332001, "args": {"groupApiId": 0, "groupDepth": 2}},
{"name": "AllReduce", "cat": "COLL_API", "ph": "b", "id": 0, "pid": 75087, "tid": 1, "ts": 0.000000, "args": {"count": 33554432, "datatype": "ncclFloat32", "root": 0, "GraphCaptured": 0, "Stream": "0x5653e4837400"}},
{"name": "AllReduce", "cat": "COLL", "ph": "b", "id": 0, "pid": 75087, "tid": 1, "ts": 337378.026001, "args": {"SeqNum": 0, "CommHash": 17093776592125785438, "Rank": 0, "Count": 33554432, "Datatype": "ncclFloat32", "Algorithm": "RING", "Protocol": "SIMPLE", "nChannels": 24}},
{"name": "KernelCh", "cat": "GPU", "ph": "b", "id": 0, "pid": 75087, "tid": 1, "ts": 339964.869995, "args": {"Channel": 0, "StartGpuClk": 1767766829233827840, "StopGpuClk": 1767766829235279872}},
{"name": "KernelCh", "cat": "GPU", "ph": "e", "id": 0, "pid": 75087, "tid": 1, "ts": 339967.748016},
[...]

PR #2391, opened 2026-09-05
profiler/example: fix zero timestamps in the trace

### LINKED ISSUE #1961: [Issue]: Start timestamp is 0.0 for certain reported events when using NCCL example profiler

Data corruption

### Share Your Debug Logs

I've been examining the example profiler and its output trace. I've identified a potential issue regarding the start timestamps of several event types.
The example trace in README.md shows that KernelLaunch and COLL_API events share an identical start timestamp of 0.0, which is highly improbable in a real-world scenario. When I'm profiling intranode AllReduce via example profiler, I can reproduce the phenomenon described before.
`// From ext-profiler/example/README.md
{"name": "KernelLaunch", ..., "ts": 0.000000, ...},
{"name": "AllReduce", ..., "ts": 0.000000, ...}`

Looking at the implementation of "exampleProfilerStartEvent" in "ext-profiler/example/plugin.cc", it seems that the start timestamp is not being updated for CollApi, P2pApi, and KernelLaunch events. This would explain why the final JSON output shows ts: 0.0.
Please let me know if this is the intended behavior for the example or if it is indeed a bug. I'd be happy to provide more details if needed.
Thanks for your time and effort!

_No response_

### NCCL Version

2.29.2

### Your platform details

_No response_

### Error Message & Behavior

_No response_

kodlan:
## Description

Two small fixes to the example profiler plugin, both visible as "ts": 0.000000 in the JSON trace it writes.

The CollApi, P2pApi and KernelLaunch events never had their startTs set in exampleProfilerStartEvent, so every one of their begin records prints ts 0.0 (the README sample trace shows it too). Every other event kind stamps startTs there, these three just missed it when they were added in 2.28.

struct kernelCh still declares type as uint8_t, but since 2.28 the plugin reads the event type of every handle as *(uint64_t*) (the new API event bits do not fit in a byte). For a KernelCh handle that read returns type + 256 * channelId, which only matches ncclProfileKernelCh for channel 0. For every other channel updateEvent and recordEventState fall through, so the end record has ts 0.0 and StopGpuClk 0, and the parent event's refCount is never dropped. This is the same one-line change as #1973, I folded it in because the issue thread covers both and a trace still looks broken with only one of them fixed. Happy to drop that hunk if you would rather take #1973.

## Related Issues

Fixes #1961
Overlaps with #1973 (kernelCh.type width)

## Changes & Impact

plugins/profiler/example/plugin.cc: three lines, event->startTs = gettime() - startTime for CollApi, P2pApi and KernelLaunch.
plugins/profiler/example/event.h: struct kernelCh type is uint64_t. This grows the struct by 8 bytes per channel slot in the collective and p2p pools, nothing serializes it.

One visible side effect of the event.h change: since the KernelCh stop now runs for every channel, the refCount cascade KernelCh -> Coll/P2p -> CollApi/P2pApi completes again, so the end records of those parent events move from enqueue time to the completion of the last channel, which is how the plugin already behaved for single channel traces and what the comment in exampleProfilerStopEvent describes. Traces diffed before and after differ only in numeric fields.

Tested on one GPU (RTX 2070) with 2 and 4 ranks on the same device (NCCL_MULTI_RANK_GPU_ENABLE=1), NCCL_PROFILER_PLUGIN=example, NCCL_PROFILE_EVENT_MASK=4095 and 4 channels. Before: all 20 API begin records per rank have ts 0.0 and 28 of 32 KernelCh end records (every one on channels 1..3) have ts 0.0 and StopGpuClk 0. After: none. Single channel traces are unchanged. The plugin builds clean with -Wall -Wextra (same pre-existing warnings as before). Not tested on multi GPU hosts, with PXN, or with CE collectives, none of which touch these lines.

Out of scope: the README sample trace still shows the old zeros and lists pool sizes of 256 while the defaults in plugin.cc are 8. The inspector plugin has its own event structs and is not affected.

## Performance Impact

None on NCCL itself. In the plugin one extra clock_gettime per profiled API call, same as every other event already does.

### DIFF
--- a/plugins/profiler/example/event.h
+++ b/plugins/profiler/example/event.h
@@ -72,7 +72,7 @@ struct kernelPhase {
 };
 
 struct kernelCh {
-  uint8_t type;
+  uint64_t type;                    // must be first and 64-bit: handle dispatch reads *(uint64_t*)
   uint8_t channelId;
   struct taskEventBase* parent;
   double startTs;
--- a/plugins/profiler/example/plugin.cc
+++ b/plugins/profiler/example/plugin.cc
@@ -553,6 +553,7 @@ __hidden ncclResult_t exampleProfilerStartEvent(void* context, void** eHandle, n
       return ncclSuccess;
     }
     event->type = ncclProfileCollApi;
+    event->startTs = gettime() - startTime;
     event->collApiId = collApiId;
     event->ctx = ctx;
     event->func = eDescr->collApi.func;
@@ -582,6 +583,7 @@ __hidden ncclResult_t exampleProfilerStartEvent(void* context, void** eHandle, n
       return ncclSuccess;
     }
     event->type = ncclProfileP2pApi;
+    event->startTs = gettime() - startTime;
     event->p2pApiId = p2pApiId;
     event->ctx = ctx;
     event->func = eDescr->p2pApi.func;
@@ -607,6 +609,7 @@ __hidden ncclResult_t exampleProfilerStartEvent(void* context, void** eHandle, n
       return ncclSuccess;
     }
     event->type = ncclProfileKernelLaunch;
+    event->startTs = gettime() - startTime;
     event->ctx = ctx;
     event->stream = (cudaStream_t) eDescr->kernelLaunch.stream;
     struct groupApi* parent = (struct groupApi *) eDescr->parentObj;


xiaofanl-nvidia:
++ @armratner @gcongiu please take a look and mirror.

armratner:
I am pushing a fix for this, I think you uncover a problem with a larger scope here.
I'll post the fix here once it's pushed.

kodlan:
> I am pushing a fix for this, I think you uncover a problem with a larger scope here. I'll post the fix here once it's pushed.

ok thanks

Issue #2379, opened 2026-08-29
Probabilistic multi-communicator init hang in the cuMem UDS fd exchange on single-node PCIe (no NVLink); NCCL_CUMEM_ENABLE=0 isolates it, 20/20 vs 5/9 A/B


tomylin890:
We spent three debugging rounds chasing a boot-time hang on a single-node 8×RTX 3090 box (no NVLink, P2P over PCIe at PXB level, NCCL 2.29.7+cuda13.2, cudaDriverVersion 13010) serving a large MoE model under vLLM with TP2×PP4 plus expert parallelism. That layout creates on the order of fifty communicators in quick succession at engine start, and roughly one boot in seven would never come up: the affected ranks pinned at 100% SM around 150W doing nothing — the classic spin-wait signature — until the framework's distributed watchdog gave up. Under rapid restart cycling the odds got dramatically worse, to roughly every second attempt. One operational footnote for anyone triaging something similar: with torch's distributed timeout lowered to 300s, the failure surfaces as the launcher dying quietly rather than printing "Watchdog caught collective operation timeout", so log-signature-based retry gates miss it.

We initially suspected stale IPC files left behind by earlier crashed runs; cleaning them on every boot attempt reduced the frequency but did not eliminate it, so the residual race is not explained by leftover state.

With `NCCL_DEBUG=INFO NCCL_DEBUG_SUBSYS=INIT,P2P` left on permanently we caught a live specimen. The count was 52 communicator Init STARTs against 50 COMPLETEs; the two stragglers were both 4-rank pipeline-parallel communicators, each missing exactly one rank. The stuck rank's trail ends mid P2P-channel build, in the middle of the shareable-buffer sequence — the last visible lines are a clean `UDS converted handle ... to fd ... on remote peer` / `Imported shareable buffer` pair, and then the sequence simply stops; the peer side waits forever on its half of the exchange. Notably the race does not always stall at the same depth: our first live specimen died during Init proper (the 52/50 count above), while the later control-arm trails show all 52 communicators completing Init and then wedging in the same UDS import sequence during post-init P2P channel setup. Both shapes end inside the identical fd-passing conversation, which is what points us at the exchange itself rather than any one call site. Our best guess at the window is the interleaving of several concurrent communicators' fd-passing conversations over those sockets during a many-PG init burst; we have not read the code deeply enough to name the exact line, which is part of why we are filing this.

[...]

PR #2396, opened 2026-09-09
Propagate UDS proxy service errors

ToLiveAndLove:
## Description

Propagate errors from the UDS proxy service to the communicator instead of
leaving clients blocked waiting for an fd response that will never arrive.

`ncclProxyServiceUDS` currently ignores the return value from
`proxyUDSRecvReq`. If exporting a cuMem handle or sending the fd response
fails, the service continues polling while the requesting thread remains in
`ncclIpcSocketRecvMsg` indefinitely.

This change stores the service error in `proxyState->asyncResult` and sets the
communicator abort flag. The waiting client then returns an NCCL error through
the existing abort check.

## Related Issues

Related to #2379. The reporter's spontaneous failure has not been reproduced
on the available four-GPU PCIe system, but the permanent-wait behavior after a
UDS response failure is deterministic.

## Changes & Impact

- Check the result of `proxyUDSRecvReq`.
- Publish the failure through the proxy async result.
- Abort the affected communicator to wake clients blocked in UDS receive.

The successful UDS request path is unchanged.

## Performance Impact

No expected steady-state performance impact. The new branch runs only after a
UDS request has already failed.

Four-GPU AllReduce from 8 B through 64 MiB:

- Baseline average bus bandwidth: 5.42546 GB/s
- Fixed average bus bandwidth: 5.46973 GB/s
- Baseline and fixed: `#wrong 0`

## Testing

Environment:

- Four GPUs, PCIe topology
- CUDA 13.3
- NCCL master `fd168324a3dc0c9080fd4881b6c7f4bb252a95a2`
- Strict build with `WERROR=1`, `sm_80`

Focused fault-injection test:

1. An `LD_PRELOAD` shim makes the first UDS fd response `sendmsg` fail with
   `EIO`.
2. Baseline initializes the communicators but completes no collective and
   times out after 30.056 seconds.
3. Fixed NCCL returns an error on all four ranks in 645 milliseconds instead
   of hanging.

Normal-path stress:

- 50 communicators initialized as one NCCL group.
- First AllReduce submitted on every communicator.
- 20 create/use/destroy cycles.
- 4000 communicator lifecycles and 4000 validated collectives completed.
- No errors; fd count returned to baseline after every cycle.

### DIFF
--- a/src/proxy.cc
+++ b/src/proxy.cc
@@ -2043,7 +2043,12 @@ void* ncclProxyServiceUDS(void* _args) {
 
     if (pollfds[0].revents) {
       // A request was seen on the UDS fd
-      proxyUDSRecvReq(proxyState, pollfds[0].fd);
+      ncclResult_t ret = proxyUDSRecvReq(proxyState, pollfds[0].fd);
+      if (ret != ncclSuccess) {
+        COMPILER_ATOMIC_STORE(&proxyState->asyncResult, ret, std::memory_order_release);
+        COMPILER_ATOMIC_STORE(proxyState->abortFlag, uint32_t(1), std::memory_order_release);
+        break;
+      }
     }
   }
 


sjeaugey:
@AddyLaddy to review

PR #2399, opened 2026-09-10
Efa gda putvalue payload inline and flushAsync

akkart-aws:
## Description

efa_gda: encode PutValue payload inline
Keep SQ posting shared while PutValue writes values directly into 128-byte RDMA-write WQEs through efa-dp-direct's published payload encoder.

Signed-off-by: Anshuman Goswami <anshumgo@amazon.com>

feat(efa-gda): Implement FlushAsync and Wait
FlushAsync and Wait were stubs. The backend tracked completion only
through the per-QP FI_WRITE NIC counter, which cannot attribute a
completion to a peer or to a specific write, so a terminating per-peer
wait was impossible.

Implement both on a CQ-derived, per-peer completion prefix that the
host publishes (the plugin's CQ progress pass, driven by NCCL's GIN
progress thread):

- Each WQE stamps req_id with a (peer, pseq) split, where pseq is the
write's position in that peer's own posting sequence. The host reads
peer/pseq from the echoed CQE and advances a contiguous
ordered_completed_count_per_peer.
- FlushAsync snapshots the context's per-peer submitted count into the
request as a terminating ticket: writes issued after the call are
excluded and cannot extend it. Wait spins until the peer's ordered
prefix reaches the ticket, then fences at the caller's memory order.

A contiguous prefix, not a bare count, is required because EFA SRD
completes out of order even to a single peer: only "everything below N
completed" proves a given earlier write finished.

Post path:

- Add per-peer backpressure (cap outstanding at peer_window) so the
host's per-peer state stays fixed size, and a per-context CQ-overflow
gate
(submitted_per_ctx - completed_per_ctx < cq_depth) since all QPs
share one CQ and an overflow drops completions.
- Route put's signal-ordering wait through the same per-peer prefix
rather than a whole-endpoint drain, so a signal waits only on its own
peer's traffic.

A failed completion does not advance that peer's prefix, so a wait
covering it does not complete, and queryLastError reports the failure.
The device is given no completion status, matching the GDAKI backend.
The deadline Wait overload maps an abort to success, matching Flush.

Handle changes (gin_efa_gda_dev.h): add submitted_count_per_peer[],
ordered_completed_count_per_peer[], peer_window,
submitted_count_per_ctx, completed_count_per_ctx and cq_depth to the
dev handle; drop the now-unused cq pointer and sq_lock.
ordered_completed_count_per_peer is published by the aws-ofi-nccl
plugin, so its type and length must match the plugin's definition
(nccl_ofi_gin_gdaki_dev.h).

Signed-off-by: Arun Karthik <akkart@amazon.com>


kgioioso:
The mirror failed because we had already merged the previous MR internally, so the cherry-pick had conflicts. @akkart-aws the previous MR should be on the `staging` branch shortly. Go ahead and rebase and then I'll re-mirror

bhasunit:
Kindly ignore/close this PR and use https://github.com/NVIDIA/nccl/pull/2405 to merge these commits.

kgioioso:
Closing in favor of 2405

PR #2183, opened 2026-05-18
Fix RoCE GID cache after port flap during P2P recovery

daxayrawal:
Problem:
After a RoCE port flap, the kernel may re-register GIDs at a different index, leaving NCCL's cached localGidIndex invalid for recovery QPs. Solution:
Add ncclIbRefreshGidInfo() and rebuild recovery QPs with a refreshed GID on Ethernet links during P2P resiliency recovery.

## Description

Problem

On RoCE (Ethernet link layer), NCCL caches localGid / localGidIndex at connect time and reuses them when moving QPs through reset → RTR → RTS during P2P port recovery. After a port flap, the kernel may re-register GIDs at a different GID table index. The cached sgid_index can then point at an empty or stale slot. Recovery QPs and data QPs modified back to RTR with the old GID fail to come up correctly, so the alive-message handshake and subsequent failback do not complete.

Native IB is unaffected: GID layout is stable across port events, so no refresh is performed for IBV_LINK_LAYER_INFINIBAND.

Solution

ncclIbRefreshGidInfo() (connect.cc, declared in common.h): For Ethernet ports only, re-query ibv_query_port, re-run ncclIbGetGidIndex(), and update gidInfo and ibDev->portAttr. On query failure, log at NCCL_NET and keep the existing cache (no hard failure).

ncclIbPortRecoveryQpsRebuild() (p2p_resiliency_recovery.cc): Before the alive-message phase, refresh GID info for the failed device, reset/re-init the port-recovery QP, transition to RTS with the updated localGid / localGidIndex, and re-post receive WRs flushed by reset (batch on receiver, single ACK WR on sender).

Recovery progress gating: In ncclIbPortRecoveryStateAliveMessages, retry rebuild throttled by IbResiliencyPortRecoveryAliveMsgBatchInterval until rebuild succeeds; reset alive-message state after a successful rebuild.

Data-path QP restore: In ncclIbPortRecoveryQpsRestore(), apply refreshed devBase->gidInfo to each data QP’s rtrAttr (and flush QP local/remote GID) before ncclIbQpRtr().

Testing

RoCE: induce port down/up (or cable pull) on a backend NIC during an active NCCL job with P2P resiliency enabled; verify recovery completes and collective traffic resumes.

Confirm
NET/IB: ... refreshed GID .../ Rebuilt recovery QP ... log lines when GID index changes after flap.

IB (non-Ethernet) path: no behavioral change expected (ncclIbRefreshGidInfo
returns immediately).

make -j src.build
[...]

PR #2400, opened 2026-09-10
Dispatch IB async events by context and port

### LINKED ISSUE #1819: [Issue]: should handle IBV_EVENT_GID_CHANGE in ncclIbAsyncThreadMain

Lower performance than expected

### Share Your Debug Logs

"GID table change" has been an unknown event type since v2.23.4-1.

We can reproduce the issue by reassigning the IP address. 
For more info, see RoCE GID Change of [Broadcom Ethernet Network Adapter User Guide](https://techdocs.broadcom.com/us/en/storage-and-ethernet-connectivity/ethernet-nic-controllers/bcm957xxx/adapters.html)

### NCCL Version

v2.23.4-1

### Your platform details

_No response_

### Error Message & Behavior

_No response_

ToLiveAndLove:
## Description

Fix port-scoped IB async event dispatch when multiple NCCL devices share an
`ibv_context`.

Each port currently starts an event reader, but `ibv_get_async_event()` reads
from the shared context. A port 2 event may be consumed by port 1's reader.
The handler then queries `dev->portNum` and refreshes port 1, leaving port 2's
GID or speed cache stale. New connections can copy that stale GID cache.

## Related Issues

Fixes #1819.

#2183 refreshes the GID during recovery. This patch fixes the earlier async
event dispatch to the per-device cache.

## Reproduction

The focused test compiles the production `common.cc` and mocks the verbs/GID
query boundaries:

1. Create port 1 and port 2 entries sharing one context.
2. Inject a port 2 `IBV_EVENT_GID_CHANGE`.
3. Deliver it to the reader represented by port 1.
4. Return a new GID for port 2 and inspect both caches and a new communicator
   snapshot.

Baseline:

```text
event port=2, reader port=1
query_port port=1
query_gid  port=1
port 2 cache/snapshot: index 2 (stale)
```

Fixed:

```text
event port=2, reader port=1
query_port port=2
query_gid  port=2
port 2 cache/snapshot: index 102
```

The same test covers speed changes, query failures, Data Direct aliases,
context isolation, and fatal-event ownership.

Results:

| Test | Baseline | Fixed |
|---|---:|---:|
| 27 focused cases | 10 failures | 0 failures |
| 100 runs in two reader modes | Same failures every run | 5,400 cases passed |
| UBSan | N/A | 27 cases passed |

## Changes & Impact

- Start one event reader per context after device enumeration and sorting.
- Dispatch GID and speed events by source context and
  `event.element.port_num`.
- Update all matching entries, including Data Direct aliases.
- Mark all entries on a context when it reports `IBV_EVENT_DEVICE_FATAL`.

This only affects the NET/IB async control path. There are no API,
configuration, collective, or transfer-path changes.

## Performance Impact

No data-path impact is expected. Default and forced-NET/IB AllReduce smoke
tests passed with zero validation errors.

## Testing

- Base: `fd168324a3dc0c9080fd4881b6c7f4bb252a95a2`
- Fix: `d3797ba68e7e3164bb995b2ba9ff5441421ec97d`
- Release builds: CUDA 13.3, `sm_80`, `WERROR=1`
- Fixed UBSan focused test: 27/27 passed
- Default and forced-NET/IB AllReduce: six runs per version, all correct

The reproducer injects events at wrapper boundaries; it does not reproduce an
IP reassignment on multi-port hardware. It also bypasses `ncclIbInitDevices()`,
so the post-sort reader startup is not covered by the focused test.

### DIFF
--- a/src/transport/net_ib/common.cc
+++ b/src/transport/net_ib/common.cc
@@ -165,9 +165,10 @@ static void ncclIbUpdateDeviceSpeed(struct ncclIbDev* dev) {
 std::thread ncclIbAsyncThread;
 void* ncclIbAsyncThreadMain(void* args) {
   struct ncclIbDev* dev = (struct ncclIbDev*)args;
+  struct ibv_context* context = dev->context;
   while (1) {
     struct ibv_async_event event;
-    if (ncclSuccess != wrap_ibv_get_async_event(dev->context, &event)) break;
+    if (ncclSuccess != wrap_ibv_get_async_event(context, &event)) break;
     char* str;
     struct ibv_cq* cq = event.element.cq;    // only valid if CQ error
     struct ibv_qp* qp = event.element.qp;    // only valid if QP error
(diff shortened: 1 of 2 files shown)


xiaofanl-nvidia:
++ @raminudelman @jynv to take a look

PR #2403, opened 2026-09-11
graph/os: close the topology file and sysfs directory on error paths

EylonKrause:
## Problem

Four error paths return with an open handle:

- `ncclTopoDumpXmlToFile`, `ncclTopoGetXmlFromFile`, `ncclTopoGetXmlGraphFromFile` (`src/graph/xml.cc`): each `fopen`s the topology/graph file and then `NCCLCHECK`s the XML dump/parse **before** `fclose`, so a failure — e.g. a malformed `NCCL_TOPO_FILE` / `NCCL_GRAPH_FILE` — returns with the `FILE*` leaked (and, for the dump, unflushed).
- `ncclOsGetBcmLinks` (`src/os/linux.cc`): `opendir`s the Broadcom switch link directory and `NCCLCHECK`s the peers `ncclRealloc` inside the `readdir` loop, returning without `closedir` if it fails.

## Fix

- xml.cc: capture the result, `fclose`, then `NCCLCHECK(res)` — the same shape as the fd fixes in #2267 / #2376.
- linux.cc: `ncclResult_t ret`, `NCCLCHECKGOTO(..., ret, exit)` and an `exit:` label just before `closedir(dir)` inside the same `if (dir)` block, returning `ret`.

No behaviour change on the success paths.

I don't have a CUDA build environment here, so this was checked by inspection plus an independent review pass (goto legality, double close, semantic change).

### DIFF
--- a/src/graph/xml.cc
+++ b/src/graph/xml.cc
@@ -298,8 +298,9 @@ ncclResult_t ncclTopoDumpXmlToFile(const char* xmlTopoFile, struct ncclXml* xml)
     INFO(NCCL_GRAPH | NCCL_ENV, "Unable to open %s, not dumping topology.", xmlTopoFile);
     return ncclSuccess;
   }
-  NCCLCHECK(ncclTopoDumpXmlRec(0, file, xml->nodes));
-  fclose(file);
+  ncclResult_t res = ncclTopoDumpXmlRec(0, file, xml->nodes);
+  fclose(file);  // close the file whether or not the dump succeeded
+  NCCLCHECK(res);
   return ncclSuccess;
 }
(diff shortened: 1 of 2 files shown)


marksantesson:
Hi @EylonKrause , I have brought this in and the change has made it to `dev`, a bit tweaked, as [this commit](https://github.com/NVIDIA/nccl/commit/b2abf6340a1321595a2928d12045359218dd8c84). If you like, please confirm that this achieves your goal and you can close the PR. Thank you for your contribution!

xiaofanl-nvidia:
This was merged. Closing.

PR #2404, opened 2026-09-11
strongstream: destroy the temporary stream on the error paths of ncclStreamAdvanceToEvent

EylonKrause:
## Problem

`ncclStreamAdvanceToEvent` (graph-capture path) creates a temporary non-blocking stream, uses it to pull the event's capture dependencies into `s`, and destroys it at the end. Every `CUDACHECK` between the create and the destroy — `cudaStreamWaitEvent`, `cudaStreamGetCaptureInfo_v2/v3`, `cudaStreamUpdateCaptureDependencies(_v2)` — returns directly on failure, so the temporary stream is leaked on any of those error paths.

## Fix

Convert those checks to `CUDACHECKGOTO(..., ret, fail)` with a `fail:` label that destroys the temporary stream and returns the error, matching the label style already used in this file (`do_unlock`). The locals the earlier jumps would cross (`status`, `nodes`, `count`, `res`) are declared before the first `goto`, and `res` is assigned rather than initialized inside the `CUDART_VERSION` branches. The success path is unchanged.

Same class as #2267 / #2376 / #2403 (release on the error path).

I don't have a CUDA build environment here, so this was checked by inspection plus an independent review pass (goto legality across the `#if` branches, double destroy, semantic change).

### DIFF
--- a/src/misc/strongstream.cc
+++ b/src/misc/strongstream.cc
@@ -393,43 +393,51 @@ ncclResult_t ncclStreamAdvanceToEvent(struct ncclCudaGraph g, cudaStream_t s, cu
   if (g.graphId == ULLONG_MAX) {
     CUDACHECK(cudaStreamWaitEvent(s, e, 0));
   } else {
+    ncclResult_t ret = ncclSuccess;
     cudaStream_t tmp;
-    CUDACHECK(cudaStreamCreateWithFlags(&tmp, cudaStreamNonBlocking));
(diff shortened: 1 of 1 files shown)


xiaofanl-nvidia:
++ @jbachan @kiskra-nvidia to take a look

kiskra-nvidia:
Thank you for your contribution! I'll take it from here. I may tweak it a bit to reduce the number of places we `return` from but the gist of it should be the same...

Issue #1996, opened 2026-01-23
[Issue]: Segmentation fault  at RAS in NCCL 2.27.7


mmhzheng:
### How is this issue impacting you?

NCCL crash training tasks.

### Share Your Debug Logs

We found that NCCL 2.27.7 triggers segmentation fault errors in RAS related code.
This problem can be bypassed using `export NCCL_RAS_ENABLE=0`
We did not actively trigger the ncclRas operation.

* case 1 using H100, NCCL version 2.27.7 + cuda12.4

```
[n125-011-039:288505:0:296459] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x48)
==== backtrace (tid: 296459) ====
 0  /usr/lib/libucs.so.0(ucs_handle_error+0x2d4) [0x7fedc925c274]
 1  /usr/lib/libucs.so.0(+0x2d44c) [0x7fedc925c44c]
 2  /usr/lib/libucs.so.0(+0x2d61a) [0x7fedc925c61a]
 3  /opt/tiger/nccl/lib/libnccl.so(+0x120057) [0x7ff1c18c1057]
 4  /opt/tiger/nccl/lib/libnccl.so(+0x12209d) [0x7ff1c18c309d]
 5  /opt/tiger/nccl/lib/libnccl.so(+0x125331) [0x7ff1c18c6331]
 6  /opt/tiger/nccl/lib/libnccl.so(+0x122e70) [0x7ff1c18c3e70]
 7  /usr/lib/x86_64-linux-gnu/libc.so.6(+0x89144) [0x7ff1c151e144]
 8  /usr/lib/x86_64-linux-gnu/libc.so.6(+0x1097dc) [0x7ff1c159e7dc]
```

<img width="1280" height="306" alt="Image" src="https://github.com/user-attachments/assets/f86528b5-4803-47cd-baef-8af7a32049fa" />

* case 2 using H20, NCCL version 2.27.7 + cuda12.4

```
==== backtrace（tid:136570）====
0 0x000000000003c050 _sigaction（）？？？：0
1 0x0000000000122927rasMsgHandlePeersUpdate（） /opt/tiger/compile_path/src/code.byted.org/data/nccl/src/ras/peers.cc:54
2 0x000000000012496d rasMsgHandle（） /opt/tiger/compile_path/src/code.byted.org/data/nccl/Src/ras/ras.cC:402
[...]

PR #2406, opened 2026-09-13
RAS: fix use-after-free in rasMsgHandlePeersUpdate

NorthSecond:
## Problem

`rasMsgHandlePeersUpdate()` merges a received `RAS_MSG_PEERSUPDATE` into the local peer/dead-peer tables. For every *newly* dead peer in the message, `rasDeadPeersUpdate()` calls `rasConnDisconnect()`, which terminates the connection and frees both the `rasConnection` and **all of its sockets** — including, possibly, the very socket the message is being received on.

If the message's `deadPeers` list contains the address of the peer it was received from, the handler frees its own connection/socket mid-processing and then keeps dereferencing `sock->conn`:

- reads at `src/ras/peers.cc` (`conn->lastSentPeersHash` / `lastRecvPeersHash` etc.),
- writes (`conn->lastSentDeadPeersHash = ...`),
- `rasConnEnqueueMsg()` into the freed connection's send queue.

This is a use-after-free on the RAS thread, which shares the address space of the training process, so a crash takes down the rank. Two properties make this worse than a rare race:

1. **No race is required.** The sender fully controls the `deadPeers` payload; a faulty (or malicious) peer can deterministically trigger the UAF on a neighbor by including its own address.
2. **It can be hit organically at scale.** When a peer is declared dead prematurely (e.g., a keepalive timeout false positive under load), it merges its own "death" and propagates the full `rasDeadPeers` array over its still-live links before exiting — delivering exactly the triggering message to its ring neighbors.

We originally root-caused this after a production training job crashed with a RAS thread segfault whose signature resembled the case-2 report in #1996; with this fix applied, the crash no longer reproduces.

## Root cause

```
rasMsgHandlePeersUpdate            src/ras/peers.cc   (sock->conn used after merge)
  -> rasDeadPeersUpdate            src/ras/peers.cc   (rasConnDisconnect on newly dead)
    -> rasConnDisconnect           src/ras/rasnet.cc
      -> rasConnTerminate          src/ras/rasnet.cc  (frees conn)
        -> rasSocketTerminate(..., finalize=true)     (frees every sock with sock->conn == conn)
```

Note that the event loop caller is already prepared for handlers terminating sockets (`rasSockEventLoop()` re-checks `rasPfds[].revents` after `rasMsgHandle()`), and `rasMsgHandleConnInitAck()` explicitly copies the address "because rasConnDisconnect will terminate the rasSocket" — the peers-update path was simply missing this protection.

## Evidence

AddressSanitizer, harness compiling the real `src/ras/peers.cc` + `src/ras/rasnet.cc` with stubbed socket I/O (no multi-node setup needed):

```
ERROR: AddressSanitizer: heap-use-after-free
READ of size 8 ... rasMsgHandlePeersUpdate      src/ras/peers.cc:609
freed by ... freeSockEntry                       src/ras/rasnet.cc:429
             rasSocketTerminate                  src/ras/rasnet.cc:578
             rasConnTerminate                    src/ras/rasnet.cc:345
             rasConnDisconnect                   src/ras/rasnet.cc:336
             rasDeadPeersUpdate                  src/ras/peers.cc:873
             rasMsgHandlePeersUpdate             src/ras/peers.cc:593
```

## Fix

In `rasMsgHandlePeersUpdate()`:

- save the connection and a copy of its address *before* the dead-peer merge;
- after `rasDeadPeersUpdate()`, if the sender is now in the dead set, drop the connection reference (otherwise re-validate it with `rasConnFind()`), so the freed connection/socket is never touched again;
- newly learned dead peers are still propagated to the remaining RAS links (`rasNetUpdatePeers(..., fromConn=nullptr)`), so dead-peer dissemination is not delayed.

The normal path (sender not in the dead set) is byte-for-byte behaviorally unchanged.

## Testing

- ASan reproducer, "self" mode (dead peer == sender): heap-use-after-free before the fix; clean pass after.
- ASan reproducer, "other" mode (dead peer is a third party): passes before and after — no regression on the normal path.
- File formatted with clang-format 22 per the repo's `.clang-format`; `maint/run-clang-format.sh --diff` reports no changes.


xiaofanl-nvidia:
++ @ayenpure to review and mirror

src/ras/peers.cc
```diff
@@ -593,6 +600,8 @@ ncclResult_t rasMsgHandlePeersUpdate(struct rasMsg* msg, struct rasSocket* sock)
       NCCLCHECKGOTO(rasDeadPeersUpdate((union ncclSocketAddress*)(((char*)msg) + deadPeersOffset),
                                        &msg->peersUpdate.nDeadPeers),
                     ret, fail);
+      // Fallback selection may have recreated the dead sender's connection during the merge.
+      conn = rasPeerIsDead(&connAddr) ? nullptr : rasConnFind(&connAddr);
```
kiskra-nvidia: Good catch with `rasPeerIsDead`! It freaks me out that this has actually been observed to trigger in real-world scenarios, but even if that were not the case, this is a gap that we should close...

Can you motivate the call to `rasConnFind` though? If the connection is not dead, then why could it possibly change? The `rasConnection` structures should be fairly persistent...

kiskra-nvidia (commented):
Looks pretty good overall; my only doubt is that extra `rasConnFind` call...

src/ras/peers.cc
NorthSecond: Removed it. In this path, `rasDeadPeersUpdate()` only disconnects connections for newly dead peers, while fallback selection does not free or replace the sender’s connection. Therefore, if `rasPeerIsDead(&connAddr)` is `false`, the cached conn remains valid; if it is `true`, the connection must no longer be used. The extra `rasConnFind()` was unnecessary, so I simplified the code to set conn to nullptr only when the sender is dead and updated the comments accordingly.

kiskra-nvidia (approved):

ayenpure (approved):
Looks good to me

xiaofanl-nvidia:
This was merged to dev branch. Closing.

PR #2407, opened 2026-09-13
Validate the p2p connect info exchange

### LINKED ISSUE #2057: [Issue]: Single-node 2-GPU ncclSend/ncclRecv fails with internal error when P2P and SHM are disabled: NET/Socket receives AF_UNIX address

Application crash

### Share Your Debug Logs

_No response_

## Environment

- NCCL version: `2.29.2+cuda12.8`
- NCCL git version: `master be4d833-dirty`
- CUDA driver version: `12080`
- OS: `Ubuntu`
- GPUs: `2 GPUs on a single node`
- Network backend: `Socket`
- Run mode: `2 ranks launched separately on the same node`

---

## 1. Summary

I hit an internal NCCL error in a single-node 2-GPU `ncclSend/ncclRecv` test when both P2P and SHM are disabled.

NCCL appears to fall back to `NET/Socket` as expected, but the `NET/Socket` path then receives an address with family `1` (`AF_UNIX`), and `ncclSocketInit` rejects it because it only supports `AF_INET` / `AF_INET6`:

```text
misc/socket.cc:537 NCCL WARN ncclSocketInit: connecting to address  with family 1 is neither AF_INET(2) nor AF_INET6(10)
```

This eventually results in:

```text
internal error - please report this issue to the NCCL developers
```

This looks like a local/shared-node address is being passed into the `NET/Socket` transport path in this configuration.

## 2. Source code reproducer

```cpp
#include <stdio.h>
#include <stdlib.h>
#include <nccl.h>
#include <cuda_runtime.h>

#define CHECK_CUDA(cmd) do { cudaError_t e = (cmd); if (e != cudaSuccess) { \
    fprintf(stderr,"CUDA error %s:%d: %s\n", __FILE__, __LINE__, cudaGetErrorString(e)); \
    exit(EXIT_FAILURE);} \
} while(0)

#define CHECK_NCCL(cmd) do { ncclResult_t r = (cmd); if (r != ncclSuccess) { \
    fprintf(stderr,"NCCL error %s:%d: %s\n", __FILE__, __LINE__, ncclGetErrorString(r));  \
    exit(EXIT_FAILURE);} \
} while(0)

int main(int argc, char* argv[]) {
    if (argc != 2) {
        fprintf(stderr, "Usage: %s <rank>\n", argv[0]);
        return EXIT_FAILURE;
    }
    int rank = atoi(argv[1]);
    int nranks = 2; // 2 gpu

    CHECK_CUDA(cudaSetDevice(rank));

    ncclUniqueId id;
    CHECK_NCCL(ncclGetUniqueId(&id));
    ncclComm_t comm;
    if (rank == 0) printf("Initializing NCCL across %d nodes...\n", nranks);
    CHECK_NCCL(ncclCommInitRank(&comm, nranks, id, rank));

    size_t total_size_bytes = 1ULL * 1024 * 1024 * 1024;
    size_t element_count = total_size_bytes / sizeof(float);
    float *d_sendbuff = nullptr;
    float *d_recvbuff = nullptr;
    CHECK_CUDA(cudaMalloc(&d_sendbuff, total_size_bytes));
    CHECK_CUDA(cudaMalloc(&d_recvbuff, total_size_bytes));

    cudaStream_t stream;
    CHECK_CUDA(cudaStreamCreate(&stream));
    printf("Starting send/recv...\n");

    // (Rank 0 <-> Rank 1)
    // CHECK_NCCL(ncclGroupStart()); // NOTE: without nccl group
    if (rank == 0) {
        CHECK_NCCL(ncclSend(d_sendbuff, element_count, ncclFloat, 1, comm, stream));
        CHECK_NCCL(ncclRecv(d_recvbuff, element_count, ncclFloat, 1, comm, stream));
    } else {
        CHECK_NCCL(ncclSend(d_sendbuff, element_count, ncclFloat, 0, comm, stream));
        CHECK_NCCL(ncclRecv(d_recvbuff, element_count, ncclFloat, 0, comm, stream));
    }
    // CHECK_NCCL(ncclGroupEnd());
    CHECK_CUDA(cudaStreamSynchronize(stream));
    printf("Done send/recv\n");

    CHECK_CUDA(cudaFree(d_sendbuff));
    CHECK_CUDA(cudaFree(d_recvbuff));
    CHECK_CUDA(cudaStreamDestroy(stream));
    CHECK_NCCL(ncclCommDestroy(comm));
    return 0;
}
```

## 3. How I run it

I launch 2 ranks on the same node, each with the same `NCCL_COMM_ID`.

Example command for rank 0:

```bash
NCCL_P2P_DISABLE=1 \
NCCL_SHM_DISABLE=1 \
NCCL_COMM_ID=172.17.16.64:23456 \
NCCL_DEBUG=INFO \
./a.out 0
```

Rank 1 is launched similarly with:

```bash
NCCL_P2P_DISABLE=1 \
NCCL_SHM_DISABLE=1 \
NCCL_COMM_ID=172.17.16.64:23456 \
NCCL_DEBUG=INFO \
./a.out 1
```

## 4. Full log

```text
ubuntu@VM-16-64-ubuntu:~$ NCCL_P2P_DISABLE=1 NCCL_SHM_DISABLE=1 NCCL_COMM_ID=172.17.16.64:23456 NCCL_DEBUG=INFO ./a.out 0
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO ENV/Plugin: Could not find: libnccl-env.so
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Bootstrap: Using eth0:172.17.16.64<0>
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL_COMM_ID set by environment to 172.17.16.64:23456
Initializing NCCL across 2 nodes...
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO cudaDriverVersion 12080
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL version 2.29.2+cuda12.8
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL git version master be4d833-dirty
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL_COMM_ID set by environment to 172.17.16.64:23456
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NET/Plugin: Could not find: libnccl-net.so
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL_COMM_ID set by environment to 172.17.16.64:23456
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NET/IB : No device found.
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NET/IB : Using [RO]; OOB eth0:172.17.16.64<0>
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Failed to initialize NET plugin IB
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL_COMM_ID set by environment to 172.17.16.64:23456
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NET/Socket : Using [0]eth0:172.17.16.64<0>
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Initialized NET plugin Socket
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Assigned NET plugin Socket to comm
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Using network Socket
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO [Rank 0] ncclCommInitRank comm 0x55eb22a04a50 rank 0 nranks 2 cudaDev 0 nvmlDev 0 busId 80 commId 0x2ef13c5ed0553a05 - Init START
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO RAS client listening socket at 127.0.0.1<28028>
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Bootstrap timings total 1.927735 (create 0.000034, send 0.000107, recv 1.927205, ring 0.000036, delay 0.000000)
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL_P2P_DISABLE set by environment to 1
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO NCCL_SHM_DISABLE set by environment to 1.
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO ncclTopoGetCpuAffinity: Affinity for GPU 0 is empty, ignoring. (GPU affinity =  ; CPU affinity = 0-19).
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO comm 0x55eb22a04a50 rank 0 nRanks 2 nNodes 2 localRanks 1 localRank 0 MNNVL 0
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Channel 00/02 : 0 1
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Channel 01/02 : 0 1
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Trees [0] 1/-1/-1->0->-1 [1] -1/-1/-1->0->1
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO P2P Chunksize set to 131072
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO PROFILER/Plugin: Could not find: libnccl-profiler.so
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Check P2P Type isAllDirectP2p 1 directMode 0 isAllCudaP2p 1
VM-16-64-ubuntu:28471:28499 [0] NCCL INFO [Proxy Service] Device 0 CPU core 19
VM-16-64-ubuntu:28471:28502 [0] NCCL INFO [Proxy Service UDS] Device 0 CPU core 0
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO TUNER/Plugin: Could not find: libnccl-tuner.so
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO threadThresholds 8/8/64 | 16/8/64 | 512 | 512
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO 2 coll channels, 2 collnet channels, 0 nvls channels, 2 p2p channels, 1 p2p channels per peer
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Symmetric memory is not supported. cuMemEnable 1, ginSupport 0, globalNicFused 1
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO CC Off, workFifoBytes 1048576
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO ncclCommInitRank comm 0x55eb22a04a50 rank 0 nranks 2 cudaDev 0 nvmlDev 0 busId 80 commId 0x2ef13c5ed0553a05 - Init COMPLETE
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Init timings - ncclCommInitRank: rank 0 nranks 2 total 2.13 (kernels 0.17, alloc 0.02, bootstrap 1.93, allgathers 0.00, topo 0.01, graphs 0.00, connections 0.00, rest 0.00)
Starting send/recv...
VM-16-64-ubuntu:28471:28503 [0] NCCL INFO [Proxy Progress] Device 0 CPU core 12
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO Channel 01/1 : 0[0] -> 1[1] [send] via NET/Socket/0/Shared

[2026-03-18 11:13:09] VM-16-64-ubuntu:28471:28499 [0] misc/socket.cc:537 NCCL WARN ncclSocketInit: connecting to address  with family 1 is neither AF_INET(2) nor AF_INET6(10)
VM-16-64-ubuntu:28471:28499 [0] NCCL INFO transport/net_socket.cc:399 -> 3
VM-16-64-ubuntu:28471:28499 [0] NCCL INFO transport/net.cc:860 -> 3
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO transport.cc:212 -> 3
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO group.cc:142 -> 3
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO group.cc:78 -> 3 [Async thread]
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO group.cc:575 -> 3
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO group.cc:781 -> 3
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO enqueue.cc:3014 -> 3
NCCL error a.cu:46: internal error - please report this issue to the NCCL developers
VM-16-64-ubuntu:28471:28471 [0] NCCL INFO ENV/Plugin: Closing env plugin ncclEnvDefault
```

## 5. Expected behavior

I would expect one of the following:

1. NCCL successfully uses `NET/Socket` in this configuration and passes a valid IPv4/IPv6 socket address to the socket transport, or
2. If single-node `P2P_DISABLE=1` + `SHM_DISABLE=1` is not a supported configuration for this path, NCCL should return a clear user-facing error instead of an internal error.

I realize this may not be a common production configuration, but since NCCL reaches the `NET/Socket` path and then fails with `internal error`, I think it would still be useful either to fix the fallback path or return a more explicit unsupported-configuration error.

### NCCL Version

2.29.2+cuda12.8

### Your platform details

_No response_

### Error Message & Behavior

_No response_

kodlan:
## Description

When two ranks talk to each other directly (any 2-rank communicator, or the rank at distance n/2 in a communicator with an even number of ranks), `ncclTransportP2pSetup` exchanges the send-side and recv-side connect infos in a single bootstrap message and assumes the other side's message mirrors its own. If the two programs don't match, for example both ranks do `ncclSend` and then `ncclRecv` without a group, the sizes can still line up and each rank ends up passing the peer's send-side info to its own send connector.

On the SHM transport that makes the kernel write past the peer's small `ncclSendMem` segment and both ranks die with `cudaErrorIllegalAddress`. On NET/Socket it's the error from the issue, `ncclSocketInit: connecting to address with family 1 ...` followed by `internal error - please report this issue`. If only one rank groups its calls the bootstrap layer reports `Message truncated` on one side and reads short on the other. The docs say this pattern should just block, so none of that should happen. The branch where the send and recv peer are different ranks has the same short-read problem.

This change fixes the exchange itself: every connect info message now carries a header with the sender's connect masks toward the receiving rank, the receiver always receives into the full message, checks that the peer's masks mirror its own and takes the layout from the header. A mismatch fails with `ncclInvalidUsage` and a warning that prints both sides' send/recv connection counts. On that error path the connectors that did not get connected are reset, so a correct grouped send/recv afterwards works on the same communicator.

## Related Issues

Fixes #2057

## Changes & Impact

- `src/transport.cc`: new `struct p2pConnectMsg` (masks header plus the connect infos), small helpers to fill, size and check it, and `ncclTransportP2pSetup` uses it for the same-peer message and for both messages of the different-peer branch. Messages are received into the full buffer. `p2pResetUnconnected` runs on the failure path and clears `hasSeen`/`p2pOnly` on the connectors that stayed unconnected.

Each peer slot now holds two message buffers during setup (64 KB instead of 32 KB), freed at the end of the call.

Tested on one node with two GPUs (RTX 5070 Ti + RTX 2070, no P2P, so 2-rank communicators go through SHM; NET/Socket via `NCCL_P2P_DISABLE=1 NCCL_SHM_DISABLE=1`) with a small two-process reproducer: both ranks ungrouped send, and one rank grouped with the other not, now return `ncclInvalidUsage` under SHM, `NCCL_SHM_LOCALITY=1` and NET/Socket instead of the illegal memory access, the hang or the internal error; grouped and correctly ordered send/recv still pass; a mismatched round followed by two grouped send/recv rounds on the same communicator passes with correct data. nccl-tests `all_reduce_perf` (default, `NCCL_ALGO=Ring`, `NCCL_ALGO=Tree`), `sendrecv_perf`, `alltoall_perf`, `broadcast_perf`, `reduce_perf` with `-g 2 -c 1` and on each GPU alone pass, and `mpirun` runs with 2, 4 and 6 ranks (`NCCL_MULTI_RANK_GPU_ENABLE=1`) pass for all_reduce (default, Ring, Tree), sendrecv, alltoall and broadcast, which exercises the different-peer branch. `make src.build WERROR=1` is clean.

Not tested: PAT (the tuner refuses `NCCL_ALGO=PAT` here with 2 and 4 ranks), NVLS tree, CollNet chain and direct, the P2P/IPC and NVLink transports, IB and multi-node. They all go through the same exchange, and for a well-formed program the check is a no-op.

Two things the check does not cover: it only runs while connections are being set up, so a mismatched pattern on a pair that is already connected is not detected and blocks in the kernel like before; and a mismatched pattern between ranks that are not each other's send and recv peer in the same round can still block on the bootstrap receive. Both are what the docs describe.

## Performance Impact

16 bytes of header per connect info message and the larger receive buffer; no extra messages. No change on the data path.


xiaofanl-nvidia:
++ @justus-nv to review and mirror

justus-nv:
@kodlan Thanks for tracking this down. The diagnosis is right. However, you are addressing the problem with a guard, rather than fixing the root cause. We also want to be cautious about adding overhead to the protocol with extra remote communication.

I'd like to fix the protocol itself rather than guard it with a separate handshake.

I recommend the following change. Prepend a header with the sender's own masks and always receive the message into the max-size buffer. Validate that the peer's masks mirror ours, and derive the layout from the header rather than from local counts.

```c++
struct p2pConnectMsg {
  uint64_t recvMask, sendMask;          // sender's own masks
  struct ncclConnect conn[2 * MAXCHANNELS];
};

// send: fill header, send sizeof(header) + (recvChannels + sendChannels) * sizeof(ncclConnect)
// recv: bootstrapRecv(..., msg, sizeof(*msg))   // max size; framing yields the true length
if (msg->recvMask != sendMask || msg->sendMask != recvMask) {
  WARN(...);
  ret = ncclInvalidUsage;
  goto fail;
}
int peerRecv = countOneBits(msg->recvMask);
sendData[p] = msg->conn;             // peer's recv-side entries feed my send connectors
recvData[p] = msg->conn + peerRecv;  // peer's send-side entries feed my recv connectors
```

Please apply the same header to the two messages in the different-peer branch as well. That path has the same short-read hazard.

*Other items*

- Use `countOneBits()` from `bitops.h` instead of `__builtin_popcountll`.
- Roll back connector state on the error path. When we return `ncclInvalidUsage`, the `hasSeen` and `p2pOnly` bits set during task append stay set, but group teardown only clears `connectSend` and `connectRecv`.
- Reword the warning. The actual constraint is that the connections each side opens in one preconnect round must mirror the peer's. "Must be in the same group" is only one of the possible operating modes.
- Ring, tree, PAT, NVLS tree, CollNet chain, CollNet direct all go through `ncclTransportP2pConnect`. Please consider testing these cases.

kodlan:
> @kodlan Thanks for tracking this down. The diagnosis is right. However, you are addressing the problem with a guard, rather than fixing the root cause. We also want to be cautious about adding overhead to the protocol with extra remote communication.
> 
> I'd like to fix the protocol itself rather than guard it with a separate handshake.
> 
> I recommend the following change. Prepend a header with the sender's own masks and always receive the message into the max-size buffer. Validate that the peer's masks mirror ours, and derive the layout from the header rather than from local counts.
> 
> ```c++
> struct p2pConnectMsg {
>   uint64_t recvMask, sendMask;          // sender's own masks
>   struct ncclConnect conn[2 * MAXCHANNELS];
> };
> 
> // send: fill header, send sizeof(header) + (recvChannels + sendChannels) * sizeof(ncclConnect)
> // recv: bootstrapRecv(..., msg, sizeof(*msg))   // max size; framing yields the true length
> if (msg->recvMask != sendMask || msg->sendMask != recvMask) {
>   WARN(...);
>   ret = ncclInvalidUsage;
>   goto fail;
> }
> int peerRecv = countOneBits(msg->recvMask);
> sendData[p] = msg->conn;             // peer's recv-side entries feed my send connectors
> recvData[p] = msg->conn + peerRecv;  // peer's send-side entries feed my recv connectors
> ```
> 
> Please apply the same header to the two messages in the different-peer branch as well. That path has the same short-read hazard.
> 
> _Other items_
> 
> * Use `countOneBits()` from `bitops.h` instead of `__builtin_popcountll`.
> * Roll back connector state on the error path. When we return `ncclInvalidUsage`, the `hasSeen` and `p2pOnly` bits set during task append stay set, but group teardown only clears `connectSend` and `connectRecv`.
> * Reword the warning. The actual constraint is that the connections each side opens in one preconnect round must mirror the peer's. "Must be in the same group" is only one of the possible operating modes.
> * Ring, tree, PAT, NVLS tree, CollNet chain, CollNet direct all go through `ncclTransportP2pConnect`. Please consider testing these cases.

Thanks! I've pushed a new verion.

src/transport.cc
```diff
@@ -122,10 +122,59 @@ ncclResult_t ncclTransportCheckP2pType(struct ncclComm* comm, bool* isAllDirectP
...
+        struct ncclConnector* conn = comm->channels[c].peers[peer]->send + connIndex;
+        if (!conn->connected) conn->hasSeen = conn->p2pOnly = 0;
+      }
+      if (comm->connectRecv[peer] & (1ULL << c)) {
+        struct ncclConnector* conn = comm->channels[c].peers[peer]->recv + connIndex;
+        if (!conn->connected) conn->hasSeen = conn->p2pOnly = 0;
+      }
+    }
+  }
+}
```
justus-nv, line 177: My guidance and your implementation missed some  additional cleanup steps.
 
```suggestion
static void p2pResetUnconnected(struct ncclComm* comm, int connIndex) {
  for (int peer = 0; peer < comm->nRanks; peer++) {
    for (int c = 0; c < MAXCHANNELS; c++) {
      struct ncclConnector* conns[2] = {NULL, NULL};
      if (comm->connectSend[peer] & (1ULL << c)) conns[0] = comm->channels[c].peers[peer]->send + connIndex;
      if (comm->connectRecv[peer] & (1ULL << c)) conns[1] = comm->channels[c].peers[peer]->recv + connIndex;
      for (int k = 0; k < 2; k++) {
        struct ncclConnector* conn = conns[k];
        if (conn == NULL || conn->connected) continue;
        if (conn->transportComm) {
          (void)conn->transportComm->free(comm, conn);
          conn->transportResources = NULL;
          conn->transportComm = NULL;
        }
        conn->hasSeen = conn->p2pOnly = 0;
      }
    }
  }
}

```

justus-nv (changes requested):
Thanks, this is the shape we were after.

Please update the PR description, it still describes your original code.

justus-nv:
@kodlan Can you address the DCO requirement. Then I will mirror the PR to our internal git repo.

kodlan:
> @kodlan Can you address the DCO requirement. Then I will mirror the PR to our internal git repo.

added Signed-off-by to the last commit

Issue #1496, opened 2024-10-28
result of sendrecv_perf is wrong


yanminjia:
I would like to improve the bus bandwidth of sendrecv_perf by resetting NCCL_CHUNK_SIZE. But it's tricky that the result of sendrecv_perf went wrong. It would be highly appreciated if any hint. Thanks.

![Image](https://github.com/user-attachments/assets/5371f19c-bf60-4e11-9e5c-9bc6a008ded5)

yanminjia:
When I set NCCL_P2P_NET_CHUNKSIZE to the same value as NCCL_CHUNK_SIZE, this problem was fixed. And the bus bandwidth reach 40+ GB/s.

PR #2410, opened 2026-09-14
enqueue: clamp NCCL_CHUNK_SIZE to the P2P step size

### LINKED ISSUE #2408: NCCL_CHUNK_SIZE / NCCL_P2P_NET_CHUNKSIZE >= 1MiB gives incorrect results and Xid 31 GPU MMU faults (H100, 8x400G RoCE)
### Summary

Setting `NCCL_CHUNK_SIZE` / `NCCL_P2P_NET_CHUNKSIZE` to **>= 1 MiB** makes `sendrecv_perf` (and `alltoall_perf`) return **incorrect results**, and both nodes log **GPU MMU faults (Xid 31)**. The throughput increase is large (~7x), but the data is wrong.

Values <= 512 KiB with matched `NCCL_CHUNK_SIZE`/`NCCL_P2P_NET_CHUNKSIZE` are correct, but give no speedup at all.

We hit this while investigating why a single GPU<->GPU connection only reaches ~13 GB/s on a 400Gb/s RoCE rail (details at the end — enlarging the chunk is the only knob that helps), so a correct way to enlarge the chunk would be very useful.

### Environment

| | |
|---|---|
| NCCL | `2.31.2-1+cuda13.3` (`libnccl.so.2.31.2`, md5 `241dc74211d08b2986ad36e478069974`) |
| nccl-tests | master `a0b82b2260cf5152b9f8c061bbf7eaf0ba096432` |
| GPUs | 2 nodes x 8x H100 80GB HBM3 SXM. GPUi <-> NICi all PIX, PCIe Gen5 x16 |
| NICs | 8x ConnectX-7 per node, 400 Gb/s, RoCEv2, MTU 9000, fw `28.43.8016` |
| Driver / CUDA | `595.71.05` / `13.2` |
| OS / kernel | Ubuntu 24.04 / `6.8.0-137-generic` |
| RDMA stack | NVIDIA OFED 24.10 (`mlx5_core` 24.10-5.1.6), `nvidia_peermem` loaded |
| Fabric | single leaf switch, PFC + ECN enabled |

### Reproduction (2 ranks, 1 rail)

```bash
mpirun -np 2 --hostfile <2 nodes, 1 slot each> --map-by ppr:1:node \
  -x NCCL_IB_HCA=mlx5_0 \
  -x NCCL_CHUNK_SIZE=1048576 -x NCCL_P2P_NET_CHUNKSIZE=1048576 \
  ./build/sendrecv_perf -b 1G -e 1G -n 20 -w 5 -g 1
```

### Observed

Default (no chunk env) — correct:

```
# Avg bus bandwidth    : 13.09
# Out of bounds values : 0 OK
```

With chunk = 1 MiB (fails on both processes):

```
.. node-a pid 1871349: Test failure common.cu:1454
.. node-b pid 2073154: Test failure common.cu:1454
prterun detected that one or more processes exited with non-zero status,
thus causing the job to be terminated. Exit code: 7
```

`dmesg` on both nodes immediately after that run:

```
NVRM: Xid (PCI:0000:18:00): 31, pid=1847919, name=sendrecv_perf, channel 0x0000000a, intr 00000000.
      MMU Fault: ENGINE GRAPHICS GPC1 GPCCLIENT_T1_7 faulted @ 0x7b79_6ba00000.
      Fault is of type FAULT_PDE ACCESS_TYPE_VIRT_READ            # node A
NVRM: Xid (PCI:0000:18:00): 31, pid=2048399, name=sendrecv_perf, channel 0x0000000a, intr 00000000.
      MMU Fault: ENGINE GRAPHICS GPC1 GPCCLIENT_T1_7 faulted @ 0x7646_91a04000.
      Fault is of type FAULT_PDE ACCESS_TYPE_VIRT_READ            # node B
```

The fault is a kernel read of a virtual address outside the registered buffer, i.e. this looks like a genuine size/payload-count bug, not a test artifact.

### Characterised behaviour (sendrecv_perf, 2 ranks, 1 rail, 1 GiB)

| configuration | algbw | result correct? |
|---|---|---|
| default | 13.08 GB/s | yes |
| `CHUNK_SIZE`=256K (P2P default) | 26.11 GB/s | no |
| `CHUNK_SIZE`=512K (P2P default) | 52.29 GB/s | no |
| `CHUNK_SIZE`=256K + `P2P_NET_CHUNKSIZE`=256K | 12.17 GB/s | yes (no speedup) |
| `CHUNK_SIZE`=512K + `P2P_NET_CHUNKSIZE`=512K | 12.24 GB/s | yes (no speedup) |
| `CHUNK_SIZE`=1M + `P2P_NET_CHUNKSIZE`=1M | 87.57 GB/s | **no** (`#wrong` ~4.7e8) |
| `CHUNK_SIZE`=4M + `P2P_NET_CHUNKSIZE`=4M, `QPS_PER_CONNECTION`=1, `SPLIT_DATA_ON_QPS`=0 | 93.12 GB/s | **no** |

So: (a) an unmatched `CHUNK_SIZE` vs `P2P_NET_CHUNKSIZE` breaks correctness at any size (same as #1496), and (b) even with the two matched, >= 1 MiB still breaks correctness. The mismatch also holds for `NCCL_IB_QPS_PER_CONNECTION` in {1, 8, 16} and `NCCL_IB_SPLIT_DATA_ON_QPS` in {0, 1}.

### Relevant source (master, `src/enqueue/enqueue.cc`)

```c
1009:    stepSize[dir] = comm->buffSizes[protocol[dir]] / NCCL_STEPS;   /* NCCL_STEPS == 8, src/include/device.h:26 */
1010:    if (protocol[dir] == NCCL_PROTO_SIMPLE) stepSize[dir] = comm->p2pChunkSize;
1011:    chunkSize[dir] = stepSize[dir];
1012:    if (paramChunkSize != 0) {
1013:      chunkSize[dir] = paramChunkSize;                 /* no upper-bound validation */
1014:    } else if (network[dir]) {
1015:      /* Tune chunk size for the network */
1016:      if (protocol[dir] == NCCL_PROTO_SIMPLE && bytes[dir] < stepSize[dir]) chunkSize[dir] /= 4;
1017:      else if (bytes[dir] < 8 * stepSize[dir]) chunkSize[dir] /= 2;
1018:    }
1020:    chunkDataSize[dir] = chunkSize[dir];
1021:    if (protocol[dir] == NCCL_PROTO_LL) chunkDataSize[dir] /= 2;
1022:    chunkDataSize_u32fp8[dir] = u32fp8Encode(chunkDataSize[dir]);
1023:    chunkDataSize[dir] = u32fp8Decode(chunkDataSize_u32fp8[dir]);
1024:    chunkSize[dir] = chunkDataSize[dir];
1025:    if (protocol[dir] == NCCL_PROTO_LL) chunkSize[dir] *= 2;
```

### Questions

1. Is `NCCL_CHUNK_SIZE >= 1 MiB` a supported value? If not, should `paramChunkSize` be validated/clamped (line 1013 currently applies it unconditionally)? If it is supported, what is the cause of the wrong payload size / the Xid 31 read past the buffer?
2. Is the fixed 8-step in-flight window per connection (`NCCL_STEPS == 8`, `buffSizes[protocol]/NCCL_STEPS`) intended, and is there any supported way to increase the per-WQE payload (or the window) so a single connection can use more of a 400 Gb/s rail?
3. Are the `chunkSize /= 4` (line 1016) and `/2` (line 1017) network-tuning steps the reason why a 1 GiB `sendrecv` with a 128 KiB step ends up with only ~16 KiB per WQE?

### Motivation (why we were tuning the chunk at all)

On the same rail the hardware is demonstrably fine:

* `ib_write_bw` (perftest, host memory): **392 Gb/s** unidirectional, **779 Gb/s** bidirectional; 384–391 Gb/s with 1–8 MiB messages; **307 Gb/s even with a single WR (1 MiB) in flight**
* `ib_write_bw --use_cuda` (GPUDirect RDMA, GPU memory): identical to host memory, 384–391 Gb/s

but NCCL:

* single GPU<->GPU connection (`sendrecv_perf`, 2 ranks, 1 rail): **13.1 GB/s** (send 6.99 + recv 6.99)
* `all_reduce` 16 ranks x 8 rails: **217 GB/s** busbw at 1 GiB, 231 GB/s at 8 GiB (2 nodes x 8x400G, would expect ~470 GB/s)

`NCCL_DEBUG=INFO NCCL_DEBUG_SUBSYS=NET` shows a per-connection in-flight window of 8 WQEs that does not change with `NCCL_IB_QPS_PER_CONNECTION` (1..16), `NCCL_MIN_NCHANNELS`/`NCCL_MAX_NCHANNELS` (1..32) or `NCCL_IB_SPLIT_DATA_ON_QPS`, and the built-in WQE latency monitor reports `inflight=8`, p50 post-to-poll ~110 us. With the effective per-WQE payload at `chunk / QPS_PER_CONNECTION` ~= 16 KiB, the single-connection rate lands at ~13 GB/s. Enlarging the chunk is the only thing that moved the needle (up to 87 GB/s), but it is unusable because of the corruption reported above.

kodlan:
## Description

Fixes wrong ncclSend/ncclRecv data and GPU memory faults when NCCL_CHUNK_SIZE is set larger than the P2P step size.

The P2P kernels move one chunk per step into a buffer slot that is stepSize bytes (comm->p2pChunkSize for the simple protocol, the LL buffer divided by NCCL_STEPS for LL). addP2pToPlan applies NCCL_CHUNK_SIZE to the chunk without checking it against that step, so a larger value makes the kernel write past its slot: it overwrites the neighbouring slots that are still in flight (wrong values) and, from the last slot, runs past the end of the buffer (illegal memory access, or an Xid 31 read fault on a GPUDirect registered network buffer). On a socket network the receiver posts a step sized receive and the sender ships the chunk, so the plugin logs "message truncated" and the kernel waits forever.

This clamps the chunk to the step size and prints a warning once (simple protocol only, the LL clamp never changes anything in practice), saying which variables actually raise the step (NCCL_BUFFSIZE together with NCCL_P2P_NET_CHUNKSIZE / NCCL_P2P_PCI_CHUNKSIZE / NCCL_P2P_NVL_CHUNKSIZE). That is the same thing computeBuffSizes already does silently for p2pChunkSize itself. While at it, values at or below zero are ignored and the chunk gets a 16 byte floor: a negative value used to encode to a huge chunk and NCCL_CHUNK_SIZE=1 crashed with a division by zero on the LL path.

## Related Issues

Fixes #2408. Same symptom as the older #1496.

## Changes & Impact

One block in src/enqueue/enqueue.cc (addP2pToPlan), 12 added lines, 2 changed. Only NCCL_CHUNK_SIZE users are affected. Values up to the step size behave as before. Values above it now get the step size plus one warning per process instead of corrupt data. That includes values that happened to work on PCIe/SHM systems because the physical slot there (buffer / NCCL_STEPS) is bigger than the step; those users will see a smaller chunk than before (about 12% less send/recv bandwidth for 256K on my PCIe host) and the warning tells them how to get a bigger one properly.

Tested on one host with two GPUs (RTX 5070 Ti and RTX 2070, no P2P, SHM transport, and NET/Socket with P2P and SHM disabled) with a small two rank send/recv program that checks every byte: without the patch a 1 MiB chunk over the 128 KiB step gives an illegal memory access on SHM, any chunk above the step hangs on NET/Socket after the "message truncated" warning, and two ranks on one GPU with a 4 MiB chunk get wrong bytes. With the patch all cases pass, and sendrecv_perf / alltoall_perf -g 2 -c 1 with NCCL_CHUNK_SIZE=1M, 256K and unset report 0 wrong values. Built with WERROR=1. I could not test IB/RoCE with GPUDirect (the reporter's setup), NVLink or multi node.

Out of scope: making bigger chunks actually work (bigger buffers / steps, what the reporter really wants for bandwidth) is a design change; and the docs still describe NCCL_CHUNK_SIZE as a bandwidth knob (happy to add a sentence about the clamp to performance_and_tuning.rst if you want it in this PR).

## Performance Impact

None on the data path. One extra comparison per P2P task at plan time. Users who set NCCL_CHUNK_SIZE above the step size get the step sized chunk, which is what the buffers can carry anyway.


sjeaugey:
The NCCL_CHUNK_SIZE environment variables is only there to help developers try different tuning when optimizing a new platform. It's not supposed to be set by users, and there was never any guarantee that you could use any value. NCCL_P2P_NET_CHUNKSIZE is documented, but in the "debugging" section, i.e. use at your own risk.

So I'm not sure I'd want to merge this PR, given it makes an already complex piece of code even harder to read and understand, for an unclear benefit.

kodlan:
> The NCCL_CHUNK_SIZE environment variables is only there to help developers try different tuning when optimizing a new platform. It's not supposed to be set by users, and there was never any guarantee that you could use any value. NCCL_P2P_NET_CHUNKSIZE is documented, but in the "debugging" section, i.e. use at your own risk.
> 
> So I'm not sure I'd want to merge this PR, given it makes an already complex piece of code even harder to read and understand, for an unclear benefit.

sure, thanks

PR #2412, opened 2026-09-15
Logging plugin proposal

zwood-meta:


xiaofanl-nvidia:
@zwood-meta is this still WIP? :) I also recommend opening a RFC to explain the change so that we can review the design before the work is all done. Thanks!

PR #2413, opened 2026-09-15
fix: preserve NIC sharing across topology trimming

polarstormx:
## Description

  With `NCCL_P2P_DISABLE=1` and `NCCL_SHM_DISABLE=1`, topology trimming can remove peer GPUs from the graph even though they still share its NICs.

  Recounting GPUs afterward underestimates NIC sharing, affecting AUTO NIC counts and potentially causing multiple GPUs to select the same NIC for GIN connection 0.

  ## Related Issues

  N/A

  ## Changes & Impact

  - Preserve per-NIC GPU sharing counts before topology trimming.
  - Reuse those counts for NET, GIN, and RMA selection.
  - Retain live-path counting during initial topology computation.
  - Add INIT logging for GIN connection-to-device mappings.

  No public API or environment-variable default changes. Existing selection policies are preserved.

  ## Performance Impact

  The fix addresses unbalanced NIC usage caused by losing sharing information during trimming.


xiaofanl-nvidia:
@polarstormx could you help resolve the conflicts for the above? 
At this point, please use `dev` branch instead since we will target the next release. Thanks!!

polarstormx:
> @polarstormx could you help resolve the conflicts for the above? At this point, please use `dev` branch instead since we will target the next release. Thanks!!

done

thomasgillis:
@polarstormx I have a few questions regarding this patch:

The code already supports this by setting `NCCL_NETDEVS_POLICY=MAX:1` which will force each GPU to use a single rail. Is this setting not sufficient for your use case? Pls note that the starting rail of each GPU is computed using the `nvmlDev` index, which will lead to a single (and different) rail per GPU to be selected.

We have designed the MAX:1 env to be usable in production. On most systems, the number of rails to be used by each GPU is known and not expected to change regardless of the GPU layout and communicator configuration.
Do you have a use-case for which this approach doesn't work?

polarstormx:
@thomasgillis Thanks! The upstream update fixes the duplicate NIC selection we were seeing, so my patch no longer seems necessary for our use case.

However, we do have a case where `MAX:1` runs into a problem when combined with `NCCL_CROSS_NIC=0`. We reproduced it on upstream NCCL 2.32.3 (`12df1a11`), built without modifications.

On two identical 8-GPU nodes, we create a communicator with two GPUs per node. Each selected GPU prefers a different local PCIe NIC, and the two GPUs are connected through NVLink.

With `NCCL_NETDEVS_POLICY=MAX:1` and `NCCL_CROSS_NIC=0`, Ring search fails:

```text
Could not find a path for pattern 4, falling back to simple order
```

The fallback Ring does not use PXN (`use ring PXN 0`) and uses different ingress and egress NICs on each node despite `CROSS_NIC=0`, potentially degrading performance or causing communication failures on rail-only networks. Changing only the policy to `AUTO` finds a same-rail Ring using PXN (`use ring PXN 1`).

PR #2423, opened 2026-09-19
Honor per-call AlltoAll CTA bounds in P2P plans

### LINKED ISSUE #2421: ncclAlltoAllConfig does not honor per-call maxCTAs
`ncclAlltoAllConfig` launches more CTAs than the per-call `maxCTAs` limit.

Reproduced with a custom NCCL 2.31.2 build on two GB300 GPUs on one node, using 8 MiB per peer and ordinary `cudaMalloc` buffers. The same communicators were reused for all cases, without communicator-level CTA overrides.

The relevant call is:

```cpp
ncclCollConfig_t config = NCCL_COLLCONFIG_INITIALIZER;
config.CTAPolicy = NCCL_CTA_POLICY_DEFAULT;
config.maxCTAs = 1;

ncclAlltoAllConfig(send, recv, 2097152, ncclFloat, comm, stream, &config);
```

Expected: the A2A kernel respects `maxCTAs`.

Observed using CUPTI concurrent-kernel activity:

| Per-call maxCTAs | AllReduce CTAs/kernel (Ring control) | AlltoAll CTAs/kernel |
| --- | ---: | ---: |
| unset | 32 | 32 |
| 1 | 1 | 32 |
| 8 | 8 | 32 |
| 32 | 32 | 32 |

A2A launches one `ncclDevKernel_SendRecv` kernel per GPU per call. These are individual kernel grid sizes, not totals across ranks or calls. Each group contains one collective per GPU communicator.

The counts were consistent across three measured calls on both GPUs. All output validation passed, and CUPTI reported no dropped records.

Source inspection of public master at `12df1a11` suggests the same issue: A2A is lowered into P2P tasks, but [`p2pTaskAppend`](https://github.com/NVIDIA/nccl/blob/12df1a11afad322be5a204a2db890161cbf8131d/src/enqueue/enqueue.cc#L2687-L2700) does not carry over the CTA bounds. Master was not tested at runtime.

0z5a:
Fixes #2421. Preserve the AlltoAll per-call maxCTAs bound when lowering to P2P work, including both enqueue paths. Resolve the existing environment/per-call/communicator precedence before lowering and use a consistent bounded mapping for connection setup, registration, device work and proxy scheduling.

Grouped AlltoAll calls use the tightest cap; ordinary Send/Recv fused into that group shares the cap. Later groups recover the default mapping. Power-of-two channel mapping means maxCTAs=3 can select two channels. Wider collective plans are separated from capped P2P work. Matching caps across ranks are required. This does not implement minCTAs or CTAPolicy.

Validation: clean CUDA 12.8 shared/static SM89 build on upstream 12df1a11afad322be5a204a2db890161cbf8131d, L20. The included MPI regression checks exact output, eager launch grids through CUPTI, captured graph kernel grids, and three real graph replays. Both enqueue paths pass: 2 GPUs with counts 0/1/257/2097152 and 4 GPUs with counts 257/2097152, totaling 576 rank/scenario results. Cases include caps unset/1/3/7, grouped unset/3/1, mixed Send/Recv and AllReduce, and restored defaults. Baseline exceeded caps 1 and 3. Multi-node behavior and the originally reported GB300 are not tested.

Two fresh-process APPA/PAAP quartets on idle L20 GPUs 4/5; CPU affinity 40–47, 20 timed iterations after warmup, exact validation outside timing. Each row is the geometric mean of four process means per arm.

| AlltoAll FP32 elements/peer | maxCTAs | Baseline µs | Patch µs | Speed change |
|---:|---:|---:|---:|---:|
| 1 | unset | 23.838 | 23.387 | +1.93% |
| 1 | 1 | 24.213 | 23.391 | +3.51% |
| 1 | 3 | 24.092 | 23.682 | +1.73% |
| 2097152 | unset | 909.798 | 943.074 | -3.53% |
| 2097152 | 1 | 907.386 | 897.637 | +1.09% |
| 2097152 | 3 | 910.555 | 902.809 | +0.86% |

Speed change = baseline/patch − 1. These are descriptive checks, not established speedups: the baseline ignores the cap, and the host is shared with unlocked clocks. The large-message default patch samples include one 1048.871 µs outlier versus approximately 909–911 µs for the other samples. No model E2E performance is claimed.

Reproduction and the standalone test are in contrib/alltoall_config_max_ctas/.

Additional regression: eight fresh 2-GPU MPI jobs (both enqueue paths × eager/graph × counts 257/2097152), 160 rank/scenario results pass exact correctness. The tenth case mixes AlltoAll maxCTAs=1 with AllReduce maxCTAs=7. For 8 MiB it observes separate 1-CTA and 4-CTA kernels in both paths and modes. The harness reports this case's grids without applying a single global cap to the wider AllReduce; the recorded grids were checked separately. This brings the final regression total to 736 rank/scenario results.


yfguo:
Hi @0z5a, thanks for the contribution.

We already have an internal fix for this issue that is currently under review, so we’d prefer not to duplicate the work here.

We appreciate your contribution and the effort you put into this PR.

0z5a:
> Hi @0z5a, thanks for the contribution.
> 
> 
> 
> We already have an internal fix for this issue that is currently under review, so we’d prefer not to duplicate the work here.
> 
> 
> 
> We appreciate your contribution and the effort you put into this PR.

Thanks for the clarification. Closing this PR to avoid duplicating the internal fix under review.

